PR #1, opened 2019-12-16
Fixes todo names

JosephSalisbury:


.gitignore
```diff
@@ -1,3 +1,4 @@
 *.swp
 TODO
+prometheus-meta-operator
```
TheoBrigitte: Only ignore file at repository root

```suggestion
/prometheus-meta-operator
```

PR #5, opened 2019-12-16
Adds namespace resource

JosephSalisbury:
Closes https://github.com/giantswarm/giantswarm/issues/8122


service/controller/controller.go
```diff
@@ -13,19 +13,19 @@ import (
 	"github.com/giantswarm/prometheus-meta-operator/pkg/project"
 )
 
-type TODOConfig struct {
+type ControllerConfig struct {
```
rossf7:
```suggestion
type ClusterConfig struct {
```

Since this watches CAPI CR's I'd name the controller cluster which aligns with SOTA.

https://github.com/giantswarm/aws-operator/blob/ac69ef376548cfeff50577003331e6264a9e7c0c/service/controller/cluster.go#L28

service/controller/controller.go
```diff
@@ -13,19 +13,19 @@ import (
...
 	K8sClient k8sclient.Interface
 	Logger    micrologger.Logger
 }
 
-type TODO struct {
+type Controller struct {
```
rossf7:
```suggestion
type Cluster struct {
```

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/stretchr/testify v1.4.0 // indirect
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9 // indirect
 	golang.org/x/sys v0.0.0-20190826190057-c7b8b68b1456 // indirect
+	k8s.io/api v0.16.4
```
rossf7: ❤️ for using modules. Fancy!

It's true the future is not evenly distributed.

JosephSalisbury: Thanks to @TheoBrigitte and @mig4, because this shit is light years beyond me :D

service/controller/controller.go
JosephSalisbury: I'd disagree with this. Sure, we watch CAPI CRs, but we're not controlling clusters - at best, it'd be prometheus controller or some such.

rossf7: True, that makes sense. 👍

kopiczko: @rossf7 is right. It's about what it watches and not what it controls. In the newer lingo that would probably be "handler" which would probably make more sense.

PR #7, opened 2019-12-17
move share logic to key package

TheoBrigitte:


service/controller/resource/namespace/resource.go
```diff
@@ -46,20 +44,14 @@ func (r *Resource) Name() string {
 }
 
 func toNamespace(v interface{}) (*corev1.Namespace, error) {
-	if v == nil {
-		return nil, nil
-	}
```
TheoBrigitte: This feels wrong, when `v` is nil then something is wrong so we should return an error.

PR #9, opened 2019-12-17
Add certificates resource

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/8121


service/key/key.go
```diff
@@ -19,3 +19,7 @@ func ToCluster(obj interface{}) (*v1alpha2.Cluster, error) {
 func Namespace(cluster *v1alpha2.Cluster) string {
 	return fmt.Sprintf("%s-prometheus", cluster.GetName())
 }
+
+func Secret() string {
+	return "cluster-certificates"
```
JosephSalisbury: Shouldn't this be `$CLUSTER_ID-prometheus`?

JosephSalisbury: Or is this the name of the target secret?

TheoBrigitte: That's the name of the secret we are going to create. Just made this name up. I can change this.

JosephSalisbury: yeah whatever, this is super ez to change

TheoBrigitte: I thought `cluster-certificates` is a meaningful name, tells you that this secret hold the certificates for the cluster. And it would be the same for every cluster, so easy to find.

JosephSalisbury: I love it, let's roll

JosephSalisbury: #nobikeshedding

service/controller/resource/certificates/create.go
```diff
@@ -0,0 +1,39 @@
...
+		return microerror.Mask(err)
+	}
+
+	sourceSecret, err := toSourceSecret(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	_, err = r.k8sClient.K8sClient().CoreV1().Secrets(targetSecret.GetNamespace()).Get(targetSecret.GetName(), metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
```
kopiczko: What if source secret is rolled?

service/controller/resource/certificates/create.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
```
kopiczko: Can we add a comment somewhere why all that stuff is done? It doesn't look obvious. From the issue:

> The Kubernetes Secrets we currently use for prometheus (e.g: `$CLUSTER_ID-prometheus`) are held in the `default` namespace.
> 
> We want to run the tenant cluster Prometheus servers in a per-tenant cluster namespace (`$CLUSTER_ID-prometheus`). `prometheus-operator` does not support referencing a secret in a different namespace.
> 
> So, we need to copy the secret `default/$CLUSTER_ID-prometheus` to the tenant cluster prometheus namespace - e.g: `$CLUSTER_ID-prometheus/$CLUSTER_ID-prometheus`

service/controller/resource/certificates/create.go
TheoBrigitte: Do you mean updated? That's no handled.

kopiczko: Isn't that a problem?

service/controller/resource/certificates/create.go
TheoBrigitte: added this here: https://github.com/giantswarm/prometheus-meta-operator/pull/16

service/controller/resource/certificates/create.go
TheoBrigitte: so far we are good with this. We can improve it later.

PR #10, opened 2019-12-17
delete namespace

TheoBrigitte:
Towards https://github.com/giantswarm/prometheus-meta-operator/pull/9
Closes https://github.com/giantswarm/giantswarm/issues/8146

### DIFF
--- a/service/controller/resource/namespace/delete.go
+++ b/service/controller/resource/namespace/delete.go
@@ -15,7 +15,7 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
 	}
 
 	err = r.k8sClient.K8sClient().CoreV1().Namespaces().Delete(namespace.GetName(), &metav1.DeleteOptions{})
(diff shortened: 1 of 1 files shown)


service/controller/resource/namespace/delete.go
```diff
@@ -13,7 +14,7 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	_, err = r.k8sClient.K8sClient().CoreV1().Namespaces().Create(namespace)
+	err = r.k8sClient.K8sClient().CoreV1().Namespaces().Delete(namespace.GetName(), &metav1.DeleteOptions{})
 	if apierrors.IsAlreadyExists(err) {
```
JosephSalisbury: This isn't right either, right?

TheoBrigitte: Do you mean the error should be `NotFound`? That's right, fixing this.

PR #19, opened 2019-12-18
Adds requests to prometheus

JosephSalisbury:
Closes https://github.com/giantswarm/giantswarm/issues/8154

We don't currently have limits on g8s-prometheus, so I don't feel like putting them in here, we've had some bad experiences. This is a good-enough default to get started with.

This ends up with:
```
    resources:
      requests:
        cpu: "1"
        memory: 1Gi
```
for the prometheus servers

### DIFF
--- a/service/controller/resource/prometheus/resource.go
+++ b/service/controller/resource/prometheus/resource.go
@@ -9,6 +9,7 @@ import (
 	"github.com/giantswarm/micrologger"
 	"github.com/giantswarm/prometheus-meta-operator/service/key"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/cluster-api/api/v1alpha2"
 )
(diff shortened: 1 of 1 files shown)


service/controller/resource/prometheus/resource.go
```diff
@@ -75,6 +75,13 @@ func toPrometheus(v interface{}) (*promv1.Prometheus, error) {
 					KeyFile:  fmt.Sprintf("/etc/prometheus/secrets/%s/key", key.Secret()),
 				},
 			},
+			Replicas: &replicas,
```
JosephSalisbury: I also alphabetised the spec because it was making me sad

service/controller/resource/prometheus/resource.go
```diff
@@ -75,6 +75,13 @@ func toPrometheus(v interface{}) (*promv1.Prometheus, error) {
...
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("1000m"),
+					corev1.ResourceMemory: resource.MustParse("1Gi"),
```
TheoBrigitte: We will be limited quite fast by this.
e.g. archon has 7g memory, so we can only monitor 7 clusters

JosephSalisbury: the entire cluster has 7g of memory?

JosephSalisbury: I remember the idea here. As we only support CAPI clusters, we were reckoning we'd be able to transition slowly. It's not that we'd suddenly double up on prometheus

kopiczko: What with customers who create fancy 3 node clusters?

JosephSalisbury: I feel like we might actually need to be smart about this. Any 'default' value is either gonna be too small for big clusters, or too big for small clusters.

JosephSalisbury: Closing here for now

PR #27, opened 2020-02-21
Use architect 0.6.0

mig4:
Adds additional checks for import rules in go code and stops injecting
version from git at build time.


TheoBrigitte:
This was done in #36 
* using architect-orb@0.8.18
* fixed goimports

PR #45, opened 2020-06-02
Reduces prometheus server replicas to one

JosephSalisbury:


JosephSalisbury:
Tested on `anteater`

JosephSalisbury:
I think I thought it was cool during the hackathon, but it's a bit much now

PR #47, opened 2020-06-03
Add SSO and HTTPS for tenant cluster prometheuses

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/8165 and https://github.com/giantswarm/giantswarm/issues/8164

This will stay as a draft as this is still untested and I need to find a way to generate certs per prometheus. I could use some help here.

This could use your review anyway.

Also, I am wondering if we should not namespace resources:
resources/
  namespace
  prometheus/
    ingress
    oauth2proxy
    frontend
.. so it would be easier to maintain

### DIFF
--- a/flag/service/prometheus/prometheus.go
+++ b/flag/service/prometheus/prometheus.go
@@ -2,4 +2,19 @@ package prometheus
 
 type Prometheus struct {
 	BaseDomain string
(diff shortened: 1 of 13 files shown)


TheoBrigitte:
Can we close this?

PR #55, opened 2020-06-15
alert: use generic resource

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/11392.


service/controller/resource/alert/rules/example.go
```diff
@@ -8,10 +8,10 @@ import (
 	"github.com/giantswarm/prometheus-meta-operator/service/key"
 )
 
-func APIServer(cluster metav1.Object) *promv1.PrometheusRule {
+func ExampleRule(cluster metav1.Object) *promv1.PrometheusRule {
```
omegas27, line 11: Why did you move to example-rule?

mig4: To highlight this is just an example, we'll want to create the actual rules from yaml either using `go generate` or by moving them into the helm chart. My thinking was that maintaining a long list of rules in go wouldn't be very readable/friendly.

omegas27: In that case, I would actually remove the code and add in a readme how to add a prometheus rule programmatically as this does not look like code we really want there anyway. @mig4  and @TheoBrigitte what do you think?

omegas27: TBH, I also think adding them using yaml is the best way as we can have them outside PMO and easier to apply

TheoBrigitte: I totally agree, we should get rid of this `alert` resource and move the alert rules into the helm chart.
But this was towards improving the reconciliation, so let's get this PR merged and make another PR where we experiment with the new idea and see if it actually works.

mig4: Cool, OK, I'm rolling with it for now then, we'll improve it next week ❤️

PR #56, opened 2020-06-16
certificates: use generic resource

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/11392.


service/controller/resource/certificates/resource.go
```diff
@@ -21,59 +23,82 @@ type Config struct {
...
 
-	return r, nil
-}
+	sc := secretCopier{clientFunc: clientFunc}
+
+	c := generic.Config{
+		ClientFunc:     clientFunc,
+		Logger:         config.Logger,
+		Name:           Name,
+		ToCR:           sc.ToCR,
```
TheoBrigitte, line 51: Nice!

mig4: I just had a thought that `ToCR` could also be a closure defined inside `New` so there wouldn't be a need for the struct, not sure if it's worth changing though.

TheoBrigitte: True, first I wondered what was this struct for. Maybe its simpler with the closure, I have no strong opinion on this.

mig4:
Tested on _godsmack_ BTW, all good.

PR #73, opened 2020-07-23
Bump memory limit for Prometheus containers to 1G

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/12251.

Memory requirements for Prometheus container depend on a couple of
factors which contribute to the fact that 200M was too low for some
installations causing Prometheus to get OOM killed as it hit the limit
shortly after startup. This in turn caused crash looping, which in turn
increased storage requirements for the WAL files.

### DIFF
--- a/service/controller/resource/prometheus/resource.go
+++ b/service/controller/resource/prometheus/resource.go
@@ -79,7 +79,7 @@ func toPrometheus(v interface{}) (metav1.Object, error) {
 					// cpu: 100m
 					corev1.ResourceCPU: *resource.NewMilliQuantity(100, resource.DecimalSI),
 					// memory: 100Mi
-					corev1.ResourceMemory: *resource.NewQuantity(200*1024*1024, resource.BinarySI),
+					corev1.ResourceMemory: *resource.NewQuantity(1*1024*1024*1024, resource.BinarySI),
 				},
 				Requests: corev1.ResourceList{
 					// cpu: 100m


mig4:
Tested on _victory_.

service/controller/resource/prometheus/resource.go
```diff
@@ -79,7 +79,7 @@ func toPrometheus(v interface{}) (metav1.Object, error) {
 					// cpu: 100m
 					corev1.ResourceCPU: *resource.NewMilliQuantity(100, resource.DecimalSI),
 					// memory: 100Mi
-					corev1.ResourceMemory: *resource.NewQuantity(200*1024*1024, resource.BinarySI),
```
omegas27, line 82: Have you tested 500Mi? 1 GB seems like a lot but maybe we should be using that. I think we need to change the requests as well, otherwise, prometheus might be rescheduled.

mig4: Yeah, I did, (see [comment](https://github.com/giantswarm/giantswarm/issues/12251#issuecomment-662985364)), it settles at ~540M at the moment, but this is highly dependant on how much we ingest and on usage, so given we're adding targets now I wanted to give it some breathing space.

Re requests, I thought about it, but in the end decided not to do it, because on most installations the memory usage was much lower, i.e. 100-200M. And if we bump _requests_ we're increasing the likelihood we'd run into scheduling issues. It seems there's no silver bullet here but the alternative is custom per-installation or per-TC values which is madness, I'd much rather spend time in looking into vertical autoscaling then. WDYT?

Re scheduling, I think we're all right. AFAIK only _requests_ are used for scheduling, _limits_ are only used to decide when to kill a container when it's getting too fat 😉

omegas27: Thanks for the detailed feedback  :)

PR #76, opened 2020-07-27
Move PMOs persistent storage to PVC

omegas27:
Towards: https://github.com/giantswarm/giantswarm/issues/11720

Tested on gauss

### DIFF
--- a/helm/prometheus-meta-operator/ci/default-values.yaml
+++ b/helm/prometheus-meta-operator/ci/default-values.yaml
@@ -4,6 +4,9 @@ Installation:
     Monitoring:
       Prometheus:
         Host: ""
+        Storage:
+          Enabled: true
+          CreatePVC: true
     Provider:
       Kind: ""
     Registry:
--- a/main.go
+++ b/main.go
@@ -110,6 +110,8 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CrtFile, "", "Certificate file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.KeyFile, "", "Key file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.BaseDomain, "", "Base domain to create Prometheus Ingress resources under.")
+	daemonCommand.PersistentFlags().Bool(f.Service.Prometheus.Storage.CreatePVC, false, "Should the operator create a PVC for storage.")
+	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Storage.Size, "20Gi", "Storage sze for prometheus.")
 	daemonCommand.PersistentFlags().String(f.Service.Provider.Kind, "", "Provider of the installation. One of aws, azure, kvm.")
 
 	err = newCommand.CobraCommand().Execute()
--- a/service/controller/resource/prometheus/resource.go
+++ b/service/controller/resource/prometheus/resource.go
@@ -64,10 +64,13 @@ func toPrometheus(v interface{}, createPVC bool, storageSize resource.Quantity) 
 
 	name := cluster.GetName()
 	var replicas int32 = 1
+	// Configured following https://github.com/prometheus-operator/prometheus-operator/issues/541#issuecomment-451884171
+	// as the volume could not mount otherwise
 	var uid int64 = 1000
-	var gid int64 = 65534
 	var fsGroup int64 = 2000
 	var runAsNonRoot bool = true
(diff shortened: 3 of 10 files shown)


JosephSalisbury:
What about installations where we don't have PVC support? I don't think we have it on a few on-prem installations

omegas27:
I was not aware of such exceptions. Do you know where I could test it?

omegas27:
This cannot be merged as we do not support storage onprem :(

JosephSalisbury:
We could have a flag in installations, and enable storage if that's true?

omegas27:
That is true yes. I suppose https://github.com/giantswarm/installations/blob/169d5852ae8bc9ca2e257ad9d101fa7eb4adc5bb/geckon/draughtsman-configmap-values.yaml#L23 could be used then?

JosephSalisbury:
Yeah, I reckon that's pretty safe

omegas27:
Tested on both ghost and geckon and this works :man_dancing: 
Maybe 20Gi is too small though

JosephSalisbury:
We'll probably have to learn about what a good storage size is over time, no worries :D

omegas27:
ping @giantswarm/team-atlas It would be nice to have a review :D

main.go
```diff
@@ -110,6 +110,8 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CrtFile, "", "Certificate file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.KeyFile, "", "Key file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.BaseDomain, "", "Base domain to create Prometheus Ingress resources under.")
+	daemonCommand.PersistentFlags().Bool(f.Service.Prometheus.Storage.IsPersistent, false, "Should the operator create a PVC for storage.")
+	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Storage.Size, "20Gi", "Storage sze for prometheus.")
```
mig4, line 114:
```suggestion
	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Storage.Size, "20Gi", "Storage size for Prometheus.")
```

helm/prometheus-meta-operator/ci/default-values.yaml
```diff
@@ -4,6 +4,9 @@ Installation:
     Monitoring:
       Prometheus:
         Host: ""
+        Storage:
+          Enabled: true
+          CreatePVC: true
```
mig4, line 9: What's the use-case for two separate settings?

AFAICS both have to be true for it to create the PVC, right? If so could we simplify this to just `CreatePVC` and likewise rename the `isPersistent` and `SupportsPersistentStorage` to `CreatePVC`?

I'd rather the CLI flag and struct attributes reflect what it's actually going to do, rather than use an indirection.

service/service.go
```diff
@@ -122,11 +124,13 @@ func New(config Config) (*Service, error) {
 	var legacyController *legacy.Controller
 	{
 		c := legacy.ControllerConfig{
-			BaseDomain:       config.Viper.GetString(config.Flag.Service.Prometheus.BaseDomain),
-			K8sClient:        k8sClient,
-			Logger:           config.Logger,
-			PrometheusClient: prometheusClient,
-			Provider:         config.Viper.GetString(config.Flag.Service.Provider.Kind),
+			BaseDomain:                config.Viper.GetString(config.Flag.Service.Prometheus.BaseDomain),
+			SupportsPersistentStorage: config.Viper.GetBool(config.Flag.Service.Prometheus.Storage.IsPersistent),
```
mig4, line 128: See note above, but my preference would be to keep this consistent everywhere and reflect what it's actually going to do, so I'd vote for renaming this to `CreatePVC`.

helm/prometheus-meta-operator/ci/default-values.yaml
omegas27: This is how it is configured under installations so I am not sure we can change it however we want but I can rename it to create PVC instead

service/controller/resource/prometheus/resource.go
```diff
@@ -58,6 +64,35 @@ func toPrometheus(v interface{}) (metav1.Object, error) {
 
 	name := cluster.GetName()
 	var replicas int32 = 1
+	// Configured following https://github.com/prometheus-operator/prometheus-operator/issues/541#issuecomment-451884171
+	// as the volume could not mount otherwise
+	var uid int64 = 1000
+	var fsGroup int64 = 2000
+	var runAsNonRoot bool = true
+	// Prometheus default image runs using the nobody user (65534)
```
TheoBrigitte, line 72: 👍

PR #79, opened 2020-08-08
Add a generic `ServiceMonitor`

mig4:
that selects services based on `giantswarm.io/monitoring` label.
Implements label based selection described in the spec
https://intranet.giantswarm.io/docs/architecture-specs-adrs/specs/configuration-of-targets-in-tc-prometheus/

Towards https://github.com/giantswarm/giantswarm/issues/11720

Towards https://github.com/giantswarm/giantswarm/issues/11724

### DIFF
--- a/service/controller/resource/servicemonitor/resource.go
+++ b/service/controller/resource/servicemonitor/resource.go
@@ -55,6 +55,7 @@ func toServiceMonitors(obj interface{}) ([]*promv1.ServiceMonitor, error) {
 	return []*promv1.ServiceMonitor{
 		service.APIServer(cluster),
 		service.NginxIngressController(cluster),
+		service.SimpleService(cluster),
 	}, nil
 }
 
--- a/service/controller/resource/servicemonitor/service/simple-service.go
+++ b/service/controller/resource/servicemonitor/service/simple-service.go
@@ -1,0 +1,119 @@
+package service
+
+import (
+	"fmt"
+	"strings"
+
+	promv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/prometheus-meta-operator/service/key"
+)
+
+// SimpleService returns a ServiceMonitor that selects services labelled with
+// the "giantswarm.io/monitoring" label.
+func SimpleService(cluster metav1.Object) *promv1.ServiceMonitor {
+	return &promv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "simple-service",
+			Namespace: key.Namespace(cluster),
+			Labels: map[string]string{
+				key.ClusterIDKey(): key.ClusterID(cluster),
+			},
+		},
+		Spec: promv1.ServiceMonitorSpec{
+			Selector: metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"giantswarm.io/monitoring": "true",
+				},
+			},
+			NamespaceSelector: promv1.NamespaceSelector{
+				Any: true,
+			},
+			Endpoints: []promv1.Endpoint{
+				{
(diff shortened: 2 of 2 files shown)


mig4:
This is untested yet (well, not all scenarios are), but I've added you as reviewers so y'all are aware of what I was working on while I'm away, just in case it's useful.

service/controller/resource/servicemonitor/resource.go
```diff
@@ -55,6 +55,7 @@ func toServiceMonitors(obj interface{}) ([]*promv1.ServiceMonitor, error) {
 	return []*promv1.ServiceMonitor{
 		service.APIServer(cluster),
 		service.NginxIngressController(cluster),
+		service.SimpleService(cluster),
```
omegas27, line 58: Why not call it TheOneAndOnly :D

service/controller/resource/servicemonitor/service/simple-service.go
```diff
@@ -0,0 +1,119 @@
...
+					Port:   "https",
+					Scheme: "https",
+					RelabelConfigs: []*promv1.RelabelConfig{
+						{
+							SourceLabels: []string{"__meta_kubernetes_namespace"},
+							TargetLabel:  "namespace",
+						},
+						{
+							SourceLabels: []string{"__meta_kubernetes_pod_name"},
+							TargetLabel:  "pod", // TODO or "pod_name" ??
```
omegas27, line 44: Pod is what we use

service/controller/resource/servicemonitor/service/simple-service.go
```diff
@@ -0,0 +1,119 @@
...
+							),
+						},
+						// Default metrics path
+						{
+							Replacement: "/metrics",
+							TargetLabel: "__tmp_metrics_path",
+						},
+						// Extract default port from __address__ before rewriting it.
+						{
+							Regex:        "(.+):(\\d+)",
```
omegas27, line 66: I thought ports could the name of the port like "metrics" as well

service/controller/resource/servicemonitor/service/simple-service.go
mig4: Looking at current targets I'd say it's not consistent, some use `pod` (e.g. the `workload` job) some use `pod_name` (pretty much everything else). I'd like to actually standardize on `pod_name` as then we can handle it using the generic relabelling rule by just stripping the `__meta_kubernetes_` prefix.

service/controller/resource/servicemonitor/service/simple-service.go
mig4: Good question. When they're specified by user, definitely, here the default `__address__` comes from Prometheus service discovery, so I'm not sure, but yeah good point, I better check it out.

service/controller/resource/servicemonitor/service/simple-service.go
omegas27: As the prometheus operator is relabelling to pod (https://github.com/prometheus-operator/prometheus-operator/blob/5c3520960e179458c0c98dae94d4cae65f71aa7f/pkg/prometheus/promcfg.go#L549), I think we need to fix the pod_name occurrences.
I just found out that pod_name is deprecated: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#removed-metrics

JosephSalisbury: Let's not stress too much about labels, we have a separate story in a bit to get them all straight

mig4: OK, I'm with you @omegas27, I've read the guidelines, it makes sense to follow them to make queries nicer.

Sure let's flesh out the details later, I'd just love to be consistent so that we can handle most of the labels with a single `labelmap` instead of having to add a ton of individual relabellings.

TheoBrigitte:
Can we close this ?

PR #81, opened 2020-08-18
add etcd target

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/11724

# cluster api
image

# legacy cluster
image


service/controller/resource/scrapeconfigs/resource.go
```diff
@@ -83,14 +89,33 @@ func toSecret(v interface{}, provider string) (*corev1.Secret, error) {
...
 
-func renderTemplate(clusterID string, provider string) (string, error) {
-	var templateData = TemplateData{
+func getTemplateData(cluster metav1.Object, provider string, clients k8sclient.Interface) (*TemplateData, error) {
+	c, ok := cluster.(*v1alpha2.Cluster)
+	if !ok {
+		return nil, microerror.Mask(fmt.Errorf("invalid v1alpha2.Cluster %T", cluster))
+	}
+
+	infra, err := clients.G8sClient().InfrastructureV1alpha2().AWSClusters(c.Spec.InfrastructureRef.Namespace).Get(c.Spec.InfrastructureRef.Name, metav1.GetOptions{})
```
omegas27, line 98: Does that not only make it usable on AWS Control planes?

service/controller/resource/scrapeconfigs/resource.go
```diff
@@ -83,14 +89,33 @@ func toSecret(v interface{}, provider string) (*corev1.Secret, error) {
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	clusterID := key.ClusterID(cluster)
+	domain := infra.Spec.Cluster.DNS.Domain
```
omegas27, line 104: Can we not use the base domain for this?

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -180,3 +180,37 @@
   - target_label: provider
     replacement: {{ .Provider }}
 {{ end }}
+# Add etcd confgiration
+- job_name: {{ .ClusterID }}-etcd
```
omegas27, line 184: Can you align the job name with the others?

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -180,3 +180,37 @@
   - target_label: provider
     replacement: {{ .Provider }}
 {{ end }}
+# Add etcd confgiration
```
omegas27, line 183:
```suggestion
# Add etcd configuration
```

service/controller/resource/scrapeconfigs/resource.go
TheoBrigitte: For this I've followed how the etcd url is built in was-operator: https://github.com/giantswarm/aws-operator/blob/e1111eef8fe7a32f9eeb2f23e130a308be2cd1c5/service/controller/key/cluster.go#L102

Where information are taken from the infra CR.

service/controller/resource/scrapeconfigs/resource.go
TheoBrigitte: Right, this make it only usable with latest cluster using cluster api. But I've added legacy cluster already, see https://github.com/giantswarm/prometheus-meta-operator/pull/81/files#diff-251764767bca0bc62afcb4201ca2d51cR95-R107

omegas27: Nice. A question from me, can the etcd host name change once for a tenant cluster or can we inject it in the config?

service/controller/resource/scrapeconfigs/resource.go
```diff
@@ -83,14 +90,36 @@ func toSecret(v interface{}, provider string) (*corev1.Secret, error) {
...
+	switch v := cluster.(type) {
+	case *v1alpha2.Cluster:
+		infra, err := clients.G8sClient().InfrastructureV1alpha2().AWSClusters(v.Spec.InfrastructureRef.Namespace).Get(v.Spec.InfrastructureRef.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		etcd = fmt.Sprintf("etcd.%s.k8s.%s:2379", key.ClusterID(cluster), infra.Spec.Cluster.DNS.Domain)
+	case *v1alpha1.AWSConfig:
+		etcd = fmt.Sprintf("%s:%d", v.Spec.Cluster.Etcd.Domain, v.Spec.Cluster.Etcd.Port)
```
JosephSalisbury, line 104: Do we need Azure and KVM configs too?

TheoBrigitte: eh, yeah ... thanks

TheoBrigitte: done

service/controller/resource/scrapeconfigs/resource.go
TheoBrigitte: I don't think the etched host name ever changes. Where in the config do you want to inject that?

omegas27: Well, maybe not the config but I was thinking we could extract the get infra stuff in a function outside getTemplateData and pass the infra to getTemplateData instead to avoid bloating the function. I am not sure where we could put it though

omegas27 (approved):

service/controller/resource/scrapeconfigs/resource.go
TheoBrigitte: I honestly also don't know where to put this logic, I think it causes no harm here.

PR #82, opened 2020-08-18
add kubelet target

TheoBrigitte:
Towards: giantswarm/giantswarm#11724

image

### DIFF
--- a/files/templates/additional-scrape-configs.template.yaml
+++ b/files/templates/additional-scrape-configs.template.yaml
@@ -196,7 +196,7 @@
     ca_file: /etc/prometheus/secrets/{{ .SecretName }}/ca
     cert_file: /etc/prometheus/secrets/{{ .SecretName }}/crt
     key_file: /etc/prometheus/secrets/{{ .SecretName }}/key
-    insecure_skip_verify: false
+    insecure_skip_verify: true
   relabel_configs:
   # Add cluster_id label.
   - target_label: cluster_id
--- a/files/templates/additional-scrape-configs.template.yaml
+++ b/files/templates/additional-scrape-configs.template.yaml
@@ -210,4 +210,3 @@
   # Add provider label.
   - target_label: provider
     replacement: {{ .Provider }}
-{{ end }}
--- a/files/templates/additional-scrape-configs.template.yaml
+++ b/files/templates/additional-scrape-configs.template.yaml
@@ -198,6 +198,8 @@
     key_file: /etc/prometheus/secrets/{{ .SecretName }}/key
     insecure_skip_verify: true
   relabel_configs:
+  - target_label: app
+    replacement: kubelet
   # Add cluster_id label.
   - target_label: cluster_id
     replacement: {{ .ClusterID }}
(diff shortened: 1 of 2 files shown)


files/templates/additional-scrape-configs.template.yaml
```diff
@@ -180,3 +180,30 @@
...
+    tls_config:
+      ca_file: /etc/prometheus/secrets/{{ .SecretName }}/ca
+      cert_file: /etc/prometheus/secrets/{{ .SecretName }}/crt
+      key_file: /etc/prometheus/secrets/{{ .SecretName }}/key
+      insecure_skip_verify: false
+  tls_config:
+    ca_file: /etc/prometheus/secrets/{{ .SecretName }}/ca
+    cert_file: /etc/prometheus/secrets/{{ .SecretName }}/crt
+    key_file: /etc/prometheus/secrets/{{ .SecretName }}/key
+    insecure_skip_verify: true
```
omegas27, line 199: Is there a specific reason why we need this to true?

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -180,3 +180,30 @@
   - target_label: provider
     replacement: {{ .Provider }}
 {{ end }}
+# Add kubelet configuration
+- job_name: {{ .ClusterID }}-kubelet
```
omegas27, line 184: Can you align the job name here?

files/templates/additional-scrape-configs.template.yaml
TheoBrigitte: Otherwise certificate verification fails because certificate only has DNS record and we connect via IP here.

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -180,3 +180,30 @@
...
+      ca_file: /etc/prometheus/secrets/{{ .SecretName }}/ca
+      cert_file: /etc/prometheus/secrets/{{ .SecretName }}/crt
+      key_file: /etc/prometheus/secrets/{{ .SecretName }}/key
+      insecure_skip_verify: false
+  tls_config:
+    ca_file: /etc/prometheus/secrets/{{ .SecretName }}/ca
+    cert_file: /etc/prometheus/secrets/{{ .SecretName }}/crt
+    key_file: /etc/prometheus/secrets/{{ .SecretName }}/key
+    insecure_skip_verify: true
+  relabel_configs:
```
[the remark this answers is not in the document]
TheoBrigitte: done

files/templates/additional-scrape-configs.template.yaml
TheoBrigitte: done PTALA

files/templates/additional-scrape-configs.template.yaml
omegas27: thanks

omegas27: Could we actually set it to false if we used the api server as a proxy like https://github.com/giantswarm/prometheus-meta-operator/blob/7533e3f59910d4c6a2d65467495783e0a2f961d1/files/templates/additional-scrape-configs.template.yaml#L3 ?

TheoBrigitte: We could but if api goes down we loose monitoring of kubelets I guess. I'd like to keep it like this.

omegas27 (approved):

PR #91, opened 2020-08-21
add app-operator target

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/11719

### DIFF
--- a/files/templates/additional-scrape-configs.template.yaml
+++ b/files/templates/additional-scrape-configs.template.yaml
@@ -362,3 +362,41 @@
     replacement: tenant_cluster
   - target_label: provider
     replacement: {{ .Provider }}
+# app-operator
+- job_name: {{ .ClusterID }}-prometheus/app-operator-{{ .ClusterID }}/0
+  honor_timestamps: false
+  metrics_path: /metrics
+  scheme: https
+  kubernetes_sd_configs:
+  - api_server: https://{{ .APIServerURL }}
+    role: service
+    tls_config:
+      ca_file: /etc/prometheus/secrets/{{ .SecretName }}/ca
+      cert_file: /etc/prometheus/secrets/{{ .SecretName }}/crt
+      key_file: /etc/prometheus/secrets/{{ .SecretName }}/key
+      insecure_skip_verify: false
+    namespaces:
+      names:
+      - giantswarm
+  tls_config:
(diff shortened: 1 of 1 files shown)


files/templates/additional-scrape-configs.template.yaml
```diff
@@ -362,3 +362,41 @@
     replacement: tenant_cluster
   - target_label: provider
     replacement: {{ .Provider }}
+# app-operator
```
omegas27, line 365: Should we not only add those if we are on the host cluster? Also, this reminds me we do not scrape the chart operator on the tenant clusters

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -362,3 +362,41 @@
...
+- job_name: {{ .ClusterID }}-prometheus/app-operator-{{ .ClusterID }}/0
+  honor_timestamps: false
+  metrics_path: /metrics
+  scheme: https
+  kubernetes_sd_configs:
```
omegas27, line 370: As those are in cluster, we probably do not need this

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -362,3 +362,41 @@
...
+    regex: http
+    action: keep
+  - source_labels: [__meta_kubernetes_pod_name]
+    target_label: pod_name
+  - target_label: app
+    replacement: kube-proxy
+  - target_label: cluster_id
+    replacement: {{ .ClusterID }}
+  - target_label: cluster_type
+    replacement: tenant_cluster
```
omegas27, line 400: This should be control_plane or management cluster

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -362,3 +362,41 @@
...
+  - source_labels: [__meta_kubernetes_service_label_app]
+    regex: app-operator
+    action: keep
+  - source_labels: [__meta_kubernetes_service_port_name]
+    regex: http
+    action: keep
+  - source_labels: [__meta_kubernetes_pod_name]
+    target_label: pod_name
+  - target_label: app
+    replacement: kube-proxy
```
omegas27, line 396:
```suggestion
    replacement: app-operator
```

PR #92, opened 2020-08-21
control plane controller

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/11719

The new control-plane controller reconcile the kubernetes api service. This results in the cluster being named `kubernetes`, because there's no way to get the installation name from within the cluster AFAIK.

Targets are added but not working yet, let handle that next.

image

### DIFF
--- a/files/templates/additional-scrape-configs.template.yaml
+++ b/files/templates/additional-scrape-configs.template.yaml
@@ -33,7 +33,7 @@
     replacement: {{ .ClusterID }}
   # Add cluster_type label.
   - target_label: cluster_type
-    replacement: tenant_cluster
+    replacement: {{ .ClusterType }}
   # Add provider label.
   - target_label: provider
     replacement: {{ .Provider }}
@@ -85,7 +85,7 @@
     replacement: {{ .ClusterID }}
   # Add cluster_type label.
   - target_label: cluster_type
-    replacement: tenant_cluster
+    replacement: {{ .ClusterType }}
   # Add provider label.
   - target_label: provider
     replacement: {{ .Provider }}
@@ -145,7 +145,7 @@
     replacement: {{ .ClusterID }}
   # Add cluster_type label.
   - target_label: cluster_type
-    replacement: tenant_cluster
+    replacement: {{ .ClusterType }}
   # Add provider label.
   - target_label: provider
     replacement: {{ .Provider }}
@@ -191,7 +191,7 @@
     replacement: {{ .ClusterID }}
   # Add cluster_type label.
   - target_label: cluster_type
-    replacement: tenant_cluster
+    replacement: {{ .ClusterType }}
   # Add provider label.
   - target_label: provider
     replacement: {{ .Provider }}
@@ -218,7 +218,7 @@
     replacement: {{ .ClusterID }}
   # Add cluster_type label.
   - target_label: cluster_type
-    replacement: tenant_cluster
+    replacement: {{ .ClusterType }}
   # Add provider label.
   - target_label: provider
     replacement: {{ .Provider }}
@@ -259,7 +259,7 @@
     replacement: {{ .ClusterID }}
   # Add cluster_type label.
   - target_label: cluster_type
-    replacement: tenant_cluster
+    replacement: {{ .ClusterType }}
   # Add provider label.
   - target_label: provider
     replacement: {{ .Provider }}
@@ -303,7 +303,7 @@
   - target_label: cluster_id
     replacement: {{ .ClusterID }}
   - target_label: cluster_type
-    replacement: tenant_cluster
+    replacement: {{ .ClusterType }}
   - target_label: provider
     replacement: {{ .Provider }}
 # kube-scheduler
@@ -346,7 +346,7 @@
   - target_label: cluster_id
     replacement: {{ .ClusterID }}
   - target_label: cluster_type
-    replacement: tenant_cluster
+    replacement: {{ .ClusterType }}
   - target_label: provider
     replacement: {{ .Provider }}
 # kube-proxy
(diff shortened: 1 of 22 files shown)


service/controller/control-plane/controller.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		selector = set.AsSelector()
+	}
+
+	var operatorkitController *controller.Controller
+	{
+		c := controller.Config{
+			K8sClient: config.K8sClient,
+			Logger:    config.Logger,
+			Name:      "control-plane-controller",
```
omegas27, line 59:
```suggestion
			Name:      project.Name() + "-control-plane-controller",
```

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -52,8 +56,9 @@
   honor_timestamps: false
   scheme: https
   kubernetes_sd_configs:
-  - api_server: https://{{ .APIServerURL }}
-    role: node
+  - role: node
+{{ if not .IsInCluster }}
```
omegas27, line 60: You might also need to change the cluster_type label so I am wondering if we should not add templatize is as a variable

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -64,12 +69,15 @@
     cert_file: /etc/prometheus/secrets/{{ .SecretName }}/crt
     key_file: /etc/prometheus/secrets/{{ .SecretName }}/key
     insecure_skip_verify: false
+{{ end }}
   relabel_configs:
+{{ if not .IsInCluster }}
   - target_label: __address__
     replacement: {{ .APIServerURL }}:443
   - source_labels: [__meta_kubernetes_node_name]
```
omegas27, line 77: Is there any reason why we are not relabelling the metrics path even when we are inside the cluster?

service/controller/resource/scrapeconfigs/resource.go
```diff
@@ -108,18 +109,22 @@ func getTemplateData(cluster metav1.Object, provider string, clients k8sclient.I
 		etcd = fmt.Sprintf("%s:%d", v.Spec.Cluster.Etcd.Domain, v.Spec.Cluster.Etcd.Port)
 	case *v1alpha1.KVMConfig:
 		etcd = fmt.Sprintf("%s:%d", v.Spec.Cluster.Etcd.Domain, v.Spec.Cluster.Etcd.Port)
+	case *corev1.Service:
+		// TODO: find a way to compute etcd url.
```
omegas27, line 113: It seems we can use relabelling for the management cluster https://github.com/giantswarm/g8s-prometheus/blob/e927fe8e3f17354b6ee884badd39f15cf0d9840e/helm/g8s-prometheus/templates/prometheus/prometheus-configmap.yaml#L231

omegas27:
It seems like you can get the installation name from .Values.Installation.V1.Monitoring.Prometheus.ClusterLabel https://github.com/giantswarm/installations/blob/cc4f7bacb85f118eca82ef598e2e6f2d36c20124/default-draughtsman-configmap-values.yaml#L204

files/templates/additional-scrape-configs.template.yaml
TheoBrigitte: We might actually need this or at least the port to get the targets correctly scraped.

TheoBrigitte:
Actually forgot that I needed some additional rbac permissions for this to work

```yaml
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRole
metadata:
  name: prometheus
rules:
- apiGroups: [""]
  resources:
  - nodes
  - nodes/metrics
  - services
  - endpoints
  - pods
  verbs: ["get", "list", "watch"]
- apiGroups: [""]
  resources:
  - configmaps
  verbs: ["get"]
- nonResourceURLs: ["/metrics"]
  verbs: ["get"]
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: prometheus
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: prometheus
subjects:
- kind: ServiceAccount
  name: default
  namespace: kubernetes-prometheus
```

files/templates/additional-scrape-configs.template.yaml
TheoBrigitte: Good catch, done.

files/templates/additional-scrape-configs.template.yaml
TheoBrigitte: Just tried this but we need to get the certificates for api server for this to work
image

service/controller/resource/scrapeconfigs/resource.go
TheoBrigitte: let's do this in next PRs, when we tackle etcd target.

omegas27: Sure :)

files/templates/additional-scrape-configs.template.yaml
omegas27: Do we not have them somewhere?

service/controller/resource/rbac/resource.go
```diff
@@ -0,0 +1,128 @@
...
+
+	clusterRole := &v1beta1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Rules: []v1beta1.PolicyRule{
+			{
+				APIGroups: []string{},
+				Resources: []string{
+					"nodes",
```
omegas27, line 58: I think we node node/proxy here

TheoBrigitte: What is this?

omegas27: This is so we can call the api/v1/nodes/node_name/proxy subresource IIUC

TheoBrigitte: added

omegas27 (approved):

PR #93, opened 2020-08-24
Rename controller names

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/12550

I followed the practice defined https://github.com/giantswarm/aws-operator/blob/02d8199912361284bffde9d6d034998baa6fe5ba/service/controller/control_plane.go#L115

### DIFF
--- a/service/controller/clusterapi/controller.go
+++ b/service/controller/clusterapi/controller.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/cluster-api/api/v1alpha2"
 
+	"github.com/giantswarm/prometheus-meta-operator/pkg/project"
 	controllerresource "github.com/giantswarm/prometheus-meta-operator/service/controller/resource"
 )
 
@@ -45,7 +46,7 @@ func NewController(config ControllerConfig) (*Controller, error) {
 		c := controller.Config{
 			K8sClient: config.K8sClient,
 			Logger:    config.Logger,
-			Name:      "clusterapi-controller",
+			Name:      project.Name() + "-cluster-api-controller",
 			NewRuntimeObjectFunc: func() runtime.Object {
 				return new(v1alpha2.Cluster)
 			},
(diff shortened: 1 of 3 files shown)


omegas27:
Tested on godsmack

service/controller/clusterapi/controller.go
```diff
@@ -45,7 +46,7 @@ func NewController(config ControllerConfig) (*Controller, error) {
 		c := controller.Config{
 			K8sClient: config.K8sClient,
 			Logger:    config.Logger,
-			Name:      "clusterapi-controller",
+			Name:      project.Name() + "-cluster-api-controller",
```
JosephSalisbury, line 49: Do we want to specify if this is for CPs or TCs?

omegas27: I was thinking we would actually have a control-plane controller as we do not need to differentiate between cluster api and legacy there

omegas27: But we can specify this is for tenant clusters if we think this is necessary

JosephSalisbury: I think this is fine, we'll know which one is which, ez

JosephSalisbury (approved):
lgmt if @TheoBrigitte agrees

JosephSalisbury:
lgmt? looks great mo tee?

TheoBrigitte (approved):
Fine with me, let's just find an efficient way to remove those finalisers

omegas27:
@JosephSalisbury can we write a batch script that patches finalizers to removes only some of them on all awsconfig, azureconfigs, kvmconfigs and clusters CR? 
I would image we could extract the list of finalizers of a CR, remove the ones we know and then reapply it?

TheoBrigitte:
I also thought about using patches but, there's no way to remove an element from an array by value. Doing as you suggest has a risk to miss some finalisers added/removed in between.

omegas27:
Then manually editing it is the only way? 
I found out that we can edit all the awsconfig in directly using `kubectl edit awsconfigs` and not do them one by one though

TheoBrigitte:
Then do a quick find & delete and we're done. Let's do that.

PR #96, opened 2020-08-25
add scraping of api services

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/11722

### DIFF
--- a/files/templates/additional-scrape-configs.template.yaml
+++ b/files/templates/additional-scrape-configs.template.yaml
@@ -505,6 +505,51 @@
   - target_label: provider
     replacement: {{ .Provider }}
 {{ if eq .ClusterType "control_plane" }}
+# app-operator
+- job_name: {{ .ClusterID }}-prometheus/app-operator-{{ .ClusterID }}/0
+  honor_timestamps: false
+  metrics_path: /metrics
+  scheme: http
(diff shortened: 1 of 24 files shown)


files/templates/additional-scrape-configs.template.yaml
```diff
@@ -505,6 +505,51 @@
...
+  - source_labels: [__address__, __meta_kubernetes_service_annotation_prometheus_io_port]
+    regex: (.+)(?::\d+);(\d+)
+    target_label: __address__
+    replacement: $1:$2
+    action: replace
+  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scheme]
+    regex: (https?)
+    target_label: __scheme__
+    replacement: $1
+    action: replace
```
TheoBrigitte, line 540: Do we really need to relabel those ? I feel this should just work and prometheus should do it itself.

TheoBrigitte (approved):
Looks correct, just wondering why we have to relabel prometheus.io annotation ourselves.

files/templates/additional-scrape-configs.template.yaml
omegas27: Technically, prometheus does it if we are using the default configuration, which we are not

PR #97, opened 2020-08-26
unit test namespace resource

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/12981

This is a first draft for unit test implementation. The goal is to have 100% coverage of our resources and be more confident with refactoring and other development tasks.

It basically verify that for a given cluster CR the get the correct namespace.

test input

```yaml
apiVersion: cluster.x-k8s.io/v1alpha2
kind: Cluster
metadata:
  name: bob
```

expected output
```yaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: null
  name: bob-prometheus
spec: {}
status: {}
```

```
$ go test -v ./service/controller/resource/namespace
=== RUN   TestNamespace
=== RUN   TestNamespace/case-0-cluster-api.golden
--- PASS: TestNamespace (0.01s)
    --- PASS: TestNamespace/case-0-cluster-api.golden (0.01s)
PASS
ok      github.com/giantswarm/prometheus-meta-operator/service/controller/resource/namespace    0.046s
```

Adding a new test case boils down to add 2 yawls files in the right places `pkg/unittest/input` for the input and `service/controller/resource/<resource name>/test` for the expected output. No go code is needed :)

Implementing a whole new test for a specific resource is done by adding those 20 lines, change `myResource` to whatever function is generating the resource and put the yamls output into `./test` directory.

```go
func TestMyResource(t *testing.T) {
	outputDir, err := filepath.Abs("./test")
	if err != nil {
		t.Fatal(err)
	}

	c := unittest.Config{
		OutputDir: outputDir,
		T:         t,
		TestFunc:  myResource,
	}
	runner, err := unittest.NewRunner(c)
	if err != nil {
		t.Fatal(err)
	}

	err = runner.Run()
	if err != nil {
		t.Fatal(err)
	}
}
```

Thoughts?


omegas27 (approved):
LGTM

PR #98, opened 2020-08-26
Add monitoring targets

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/11721


mig4:
Question: could these not be done using a `ServiceMonitor`? Seems like they would be perfect for that.

omegas27:
We actually decided to put everything in scrape configs for now to scrape everything and move everything away to one service generic service monitor once labels are added

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -28,7 +28,11 @@
     replacement: {{ .APIServerURL }}:443
   - source_labels: [__meta_kubernetes_node_name]
     target_label: __metrics_path__
+{{ if eq .ClusterType "control_plane" }}
+    replacement: /api/v1/nodes/${1}:9393/proxy/metrics
```
omegas27: why 9393 when the port is 9323? https://github.com/giantswarm/giantnetes-terraform/blob/5982dee3a754a43c5ca25749c86f614a90f18a62/templates/vault.yaml.tmpl#L59

omegas27: You changed it already, but the port on the tenant clusters should be 9323 and should only work for clusters running 12.0.0+.
Docker is advising to use 9323 https://docs.docker.com/config/daemon/prometheus/#configure-docker so I updated that value in k8scloudconfig

JosephSalisbury (approved):

TheoBrigitte:
* cluster v11.5.2 - 0 Unhealthy targets
* cluster v9.3.6 - 2 Unhealthy targets (docker and nginx, expected and tracked down)
* control plane - 3 Unhealthy targets (calico, kube-proxy, docker, expected and tracked down)

PR #101, opened 2020-08-28
Select Cluster type version based on CRD's storage version

tuommaki:
Towards: https://github.com/giantswarm/giantswarm/issues/13005

v1alpha2.Cluster and v1alpha3.Cluster are incompatible and when given storage
version is manipulated with non-matching type, it will always result in data
loss. Select used version during runtime based on CRD's storage version
configuration.

### DIFF
--- a/service/controller/resource/certificates/resource.go
+++ b/service/controller/resource/certificates/resource.go
@@ -66,10 +66,9 @@ func (sc *secretCopier) ToCR(v interface{}) (metav1.Object, error) {
 		return nil, microerror.Mask(err)
 	}
 
-	var sourceData map[string][]byte
 	sourceSecret, err := sc.getSource(context.TODO(), v)
-	if err == nil {
-		sourceData = sourceSecret.Data
+	if err != nil {
+		return nil, microerror.Mask(err)
 	}
 
 	secret := &corev1.Secret{
@@ -77,7 +76,7 @@ func (sc *secretCopier) ToCR(v interface{}) (metav1.Object, error) {
 			Name:      key.Secret(),
 			Namespace: key.Namespace(cluster),
 		},
-		Data: sourceData,
+		Data: sourceSecret.Data,
 	}
 
 	return secret, nil
--- a/service/controller/resource/certificates/resource.go
+++ b/service/controller/resource/certificates/resource.go
@@ -66,9 +66,10 @@ func (sc *secretCopier) ToCR(v interface{}) (metav1.Object, error) {
 		return nil, microerror.Mask(err)
 	}
 
+	var sourceData map[string][]byte
 	sourceSecret, err := sc.getSource(context.TODO(), v)
-	if err != nil {
-		return nil, microerror.Mask(err)
+	if err == nil {
+		sourceData = sourceSecret.Data
 	}
 
 	secret := &corev1.Secret{
@@ -76,7 +77,7 @@ func (sc *secretCopier) ToCR(v interface{}) (metav1.Object, error) {
 			Name:      key.Secret(),
 			Namespace: key.Namespace(cluster),
 		},
-		Data: sourceSecret.Data,
+		Data: sourceData,
 	}
 
 	return secret, nil
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add monitoring label
+- etcd target for control planes
+- vault target
+- gatekeeper target
+
 ## [1.1.0] - 2020-08-27
 
 ### Added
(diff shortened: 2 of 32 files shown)


tuommaki:
At this point this doesn't break the existing Cluster CRs, but the functionality is still not fully right because `scrapeconfigs` don't have support for `v1alpha3`.

service/controller/resource/certificates/resource.go
```diff
@@ -66,17 +66,18 @@ func (sc *secretCopier) ToCR(v interface{}) (metav1.Object, error) {
 		return nil, microerror.Mask(err)
 	}
 
+	var sourceData map[string][]byte
 	sourceSecret, err := sc.getSource(context.TODO(), v)
-	if err != nil {
-		return nil, microerror.Mask(err)
+	if err == nil {
```
mig4, line 71: The drawback of this is that in situation when a cluster is being **created** but the secret doesn't exist, we won't see an error here and we'll just create an empty secret which will lead to a failure further down the line.

Don't know what creates those source secret objects and how likely is it that they wouldn't be there when creating a cluster.

Looking forward to any comments on what y'all think about this.

omegas27: IIUC, those are created by the cert operator when the cluster operator creates cert config CRD. I would rather we handle the error when it happens and not create a dummy secret. Could we not wait till the secret is populated before creating the prometheus CR instead?

TheoBrigitte:
Can we just move from v1alpha2 to v1alpha3 ? Instead of having a logic to detect what's the version in storage.

omegas27:
Are we sure every cluster is using v1alpha3 though? Also, this might be needed once v1alpha4 or v1beta1 comes out

TheoBrigitte:
Ok apparently we only have v1alpha3 on Azure installations. So my idea won't work.

```
$ k --context giantswarm-asgard get crd clusters.cluster.x-k8s.io -oyaml|yq -r '.spec.versions[].name'
v1alpha2
$ k --context giantswarm-ghost get crd clusters.cluster.x-k8s.io -oyaml|yq -r '.spec.versions[].name'
v1alpha2
v1alpha3
$ k --context giantswarm-amagon get crd clusters.cluster.x-k8s.io -oyaml|yq -r '.spec.versions[].name'
v1alpha2
```

mig4:
Yeah problem is AWS is still on `v1alpha2`, Azure is on `v1alpha3` and in the future it's probably going to be even more complicated.

service/controller/resource/certificates/resource.go
omegas27: That would be cool :)

omegas27: Can we move that outside the PR so we can approve it and fix the CRD issue?

mig4: Yip, so https://github.com/giantswarm/prometheus-meta-operator/pull/115 is for dealing with the deletions and finalizer and refactoring, I'll update this to just deal with the `v1alpha3` resources.

omegas27 (approved):

service/controller/resource/certificates/resource.go
mig4: Done, reverted here.

mig4:
Tested on _godsmack_, the diff of `Cluster` CR before and after scaling up PMO:

``` diff
❯ diff -u $INSTALLATION.cluster.$TC.s0-nopmo.yaml $INSTALLATION.cluster.$TC.s1-withpmo.yaml
--- godsmack.cluster.9z4fo.s0-nopmo.yaml        2020-09-03 07:05:40.861288830 +0000
+++ godsmack.cluster.9z4fo.s1-withpmo.yaml      2020-09-03 07:07:20.110248811 +0000
@@ -4,6 +4,8 @@
   annotations:
     cluster.giantswarm.io/description: mig4-prom-test-1
   creationTimestamp: "2020-09-03T07:03:42Z"
+  finalizers:
+  - operatorkit.giantswarm.io/prometheus-meta-operator-cluster-api-controller
   generation: 1
   labels:
     azure-operator.giantswarm.io/version: 5.0.0-nodepools
@@ -13,7 +15,7 @@
     release.giantswarm.io/version: 42.0.0
   name: 9z4fo
   namespace: default
-  resourceVersion: "197427144"
+  resourceVersion: "197427786"
   selfLink: /apis/cluster.x-k8s.io/v1alpha3/namespaces/default/clusters/9z4fo
   uid: 010a462c-6d97-474b-9ee3-21fea4c2d83e
 spec:
```

TheoBrigitte:
Does the rest of the resources execute correctly when reconciling `v1alpha3.Cluster` object? I'm mainly thinking about cases like this: https://github.com/giantswarm/prometheus-meta-operator/blob/bb1758c83acff134f0711eac67aff321a8d596b6/service/key/key.go#L56-L57

mig4:
I wasn't aware of that, pushed a follow-up: https://github.com/giantswarm/prometheus-meta-operator/pull/116

PR #112, opened 2020-09-02
Update list of targets

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/12925

### DIFF
--- a/Documentation/targets.md
+++ b/Documentation/targets.md
@@ -102,7 +102,7 @@
 
 ## Unscrapeable
 
-* happa: not exposing any metrics because
+* happa: not exposing any metrics
 
 * metrics-server: not scrapping because according to official documentation it should not be used [as a source of monitoring solution metrics](https://github.com/kubernetes-sigs/metrics-server#kubernetes-metrics-server).
 
--- a/Documentation/targets.md
+++ b/Documentation/targets.md
@@ -1,30 +1,113 @@
 # Targets
 
+- alertmanager
+- api
+- app-exporter
+- app-operator
 - cadvisor
 - calico
 - cert-exporter
 - chart-operator
 - cluster-autoscaler
+- cluster-operator
+- cluster-service
+- companyd
+- credentiald
+- crsync
 - docker
+- draughtsman
 - etcd
+- etcd-backup-operator
+- fluentbit
+- g8s-grafana
+- g8s-prometheus
+- gatekeeper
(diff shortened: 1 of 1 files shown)


Documentation/targets.md
```diff
@@ -1,30 +1,113 @@
...
+
+### KVM specific
+
+- calico-kube-controllers
+- heapster
+- monitoring-influxdb
+
+## Unscrapeable
+
+* happa: not exposing any metrics because
```
TheoBrigitte, line 105: because what?

Documentation/targets.md
```diff
@@ -1,30 +1,113 @@
...
+
+- kvm-operator
+
+## Missing
+
+- bridge-operator
+- cert-operator
+- prometheus-operator-app
+- CP prometheus
+- TC prometheus
```
TheoBrigitte, line 65: What's the difference?

Documentation/targets.md
TheoBrigitte, line 105: Because what?

omegas27: There should not be any because, its just not exposing metrics :D

TheoBrigitte (changes requested):

Documentation/targets.md
omegas27: That's mostly so we do not forget and because CP prometheus should be checked via heartbeats while TC prometheus should be scraped by the CP prometheus

Documentation/targets.md
TheoBrigitte: ok, let's remove this `because` then, I think it makes the sentence weird.

Documentation/targets.md
TheoBrigitte: I don't think we're planning on scrapping TC prometheus with CP prometheus. But I'm now thinking we need a way to know when prometheuses are down.

Documentation/targets.md
omegas27: Done

Documentation/targets.md
omegas27: I thought it was what we intended 
image

TheoBrigitte (approved):

Documentation/targets.md
JosephSalisbury: We need to scrape the TC prometheus servers from the CP prometheus server.

PR #113, opened 2020-09-02
add managed-app/crsync target

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/11723

Tested ...

on `gauss`

image

on `gorilla`

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add monitoring label
 - gatekeeper target
+- managed-app target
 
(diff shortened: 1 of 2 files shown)


omegas27 (commented):
This might overlap with some of our existing targets

TheoBrigitte:
> This might overlap with some of our existing targets

Yes there's a potential overlap, but that's the current setup we have. And targets are already overlapping

image

image

How about we exclude target with `giantswarm.io/monitoring` label from other targets?

omegas27:
It would make sense that we drop them if we have the label

TheoBrigitte:
> them

Ok, I'll do that in next PR then.

omegas27 (approved):

PR #116, opened 2020-09-03
Fix API URL computation for Cluster objects

mig4:
Towards: https://github.com/giantswarm/giantswarm/issues/13005


TheoBrigitte (approved):
Nice 👍

mig4:
Tested on _godsmack_, the approach of testing against generic `Object` interface works so we don't need to enumerate all possible types.

PR #128, opened 2020-09-15
human readable test output for scrape configs

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/12981

`scrapeconfigs` resource produce a Secret where the data are base64 encoded. This makes the test result undreadable and meaningless for humans.
This PR improves the situation by testing the data produces before it is encoded. Therefore we can store the plain scrapeconfig as test result, which should help us understand better what changes are made over time.


omegas27 (approved):
This makes a lot of sense

PR #129, opened 2020-09-17
Add missing labels to metrics

omegas27:


service/controller/resource/prometheus/resource.go
```diff
@@ -118,6 +119,15 @@ func toPrometheus(v interface{}, createPVC bool, storageSize resource.Quantity)
 				key.ClusterIDKey(): key.ClusterID(cluster),
 				"cluster_type":     key.ClusterType(cluster),
 			},
+			PodMetadata: &promv1.EmbeddedObjectMetadata{
+				Labels: map[string]string{
+					"giantswarm.io/monitoring":     "true",
+					"app.kubernetes.io/name":       "prometheus",
+					"app.kubernetes.io/managed-by": project.Name(),
+					"app.kubernetes.io/version":    project.Version(),
```
mig4, line 127: This isn't going to be version of _"prometheus"_. I'd skip this label here as it could be confusing. There's also a version field in spec that specifies Prometheus version.

service/controller/resource/scrapeconfigs/test/aws/case-0-cluster-api.golden
```diff
@@ -26,6 +26,9 @@
     replacement: /api/v1/nodes/${1}:9323/proxy/metrics
   - target_label: app
     replacement: docker
+  - source_labels: [__address__]
+    regex: '.*'
+    target_label: instance
```
mig4, line 31: Isn't this the default?

service/controller/resource/scrapeconfigs/test/aws/case-0-cluster-api.golden
```diff
@@ -35,6 +38,9 @@
   # Add provider label.
   - target_label: provider
     replacement: aws
+  # Add installation label.
+  - target_label: installation
+    replacement:
```
mig4, line 43: Should this be empty?

service/controller/resource/scrapeconfigs/test/aws/case-0-cluster-api.golden
```diff
@@ -725,10 +799,18 @@
   relabel_configs:
   - source_labels: [__meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring]
     regex: (true)
-    action: keep
-  - source_labels: [__meta_kubernetes_service_annotation_giantswarm_io_monitoring]
+    target_label: __tmp_keep_me
+    replacement: true
+  - source_labels: [__meta_kubernetes_service_labelpresent_giantswarm_io_monitoring]
+    regex: (true)
```
mig4, line 805: I'd vote for only using the label `giantswarm.io/monitoring`, that's what we documented in the spec.

mig4 (changes requested):
Couple of questions from me 😉

service/controller/resource/scrapeconfigs/test/aws/case-0-cluster-api.golden
omegas27: Nope it is not :D

service/controller/resource/scrapeconfigs/test/aws/case-0-cluster-api.golden
omegas27: No, the cm is probably empty

service/controller/resource/scrapeconfigs/test/aws/case-0-cluster-api.golden
omegas27: We cannot as we need to support legacy apps that used this label

mig4: What are those? Just asking as annotation is not compatible with selectors so this would block being able to use the _prometheus-operator_ CRs to manage things.

omegas27: I adjusted some apps but they are present in existing clusters still. Older releases will have them

omegas27: The annotations were selected by halo and batman before atlas even existed

service/controller/resource/scrapeconfigs/test/aws/case-0-cluster-api.golden
omegas27: @TheoBrigitte I am not sure how I should set this one up. Should it be generated? Do you have a clue?

TheoBrigitte (approved):

PR #127, opened 2020-09-15
Add scraping of prometheuses

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/13223 and https://github.com/giantswarm/giantswarm/issues/13222

PR #131, opened 2020-09-17
Test other or

omegas27:


files/templates/additional-scrape-configs.template.yaml
```diff
@@ -826,12 +855,12 @@
     insecure_skip_verify: true
 {{ end }}
   relabel_configs:
-  - source_labels: [__meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring]
+  - source_labels: [__meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring;__meta_kubernetes_service_labelpresent_giantswarm_io_monitoring]
```
mig4, line 858:
```suggestion
  - source_labels: [__meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring, __meta_kubernetes_service_labelpresent_giantswarm_io_monitoring]
```

mig4 (changes requested):

files/templates/additional-scrape-configs.template.yaml
omegas27: I tried that before. Let me change it back so you can try to deploy it

omegas27: Done

mig4: Cool, thanks.

So this is on managed app target, so to test it I need to create a TC, then deploy some managed app and check if it's picked up, right? Which managed app did you use for testing BTW?

omegas27: It worked with externaldns, a default app

omegas27: You can always deploy your app with the label and annotation

mig4: OK, but what app? Bear with me here, I've never deployed an app and have no clue how you even do that yet.

Anyway, I'll try with externaldns sure.

omegas27: I was thinking you could deploy any kubernetes deployment, crearte a service that points to the metrics path ahd add the label and annotations htere but otherwise, you can do it via happ or by creating an app CR

omegas27: Nginx, the latest 1.9.2 version can also be installed

mig4: I tried with nginx-IC and I think I got it, but there's already a separate target specifically for nginx-IC and it's being picked up there.

OK, I'll try something else.

mig4:
OK this works with unbounded expression: `.*(true).*`.

omegas27:
Nice, I am wondering if it would work with (true;false)|(true;true)|(false;true) but that is cool. Could you maybe document this under relabelling patterns somewhere so we do not forget?

mig4:
Do we have docs for relabelling patterns somewhere?

omegas27:
Sadly, no, but now we will :D

omegas27:
Can you set the change in the other PR and close this one?

mig4:
It is actually documented in [Prometheus docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#regex), although I know that it a bit dry and without examples.

mig4:
Yeah I'm looking into it now, so the other PR is #127 right? Did you branch off from it? Like can I merge this into the other one or do I need to just move the changes there?

omegas27:
Yes, I branched out of this one so adding it here makes sense. I will rebase the other branch

What does anchor mean in that context in the docs?

mig4:
Oh, anchored means that it must match from the beginning to end of string. So in other words it's as if the regex was implicitly starting with `^` and ending with `$`, so if you want to match a substring anywhere in the string you need to wrap it with `.*`.

mig4:
I wonder if we can change the base branch here from master to `add-monitoring-of-prometheuses` branch and merge it?

mig4:
Yeah, I tested that locally and it works OK. If you're OK with that I can take care. I didn't want to just do it to not get in your way in case you were working on it so just let me know.

omegas27:
I think the original branch is add monitoring of prometheus already so you can just merge it there :)

omegas27:
I backported the or fix to https://github.com/giantswarm/prometheus-meta-operator/pull/127

PR #133, opened 2020-09-21
Fix some missing labels on metrics

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/10904

### DIFF
--- a/files/templates/additional-scrape-configs.template.yaml
+++ b/files/templates/additional-scrape-configs.template.yaml
@@ -140,7 +140,7 @@
     insecure_skip_verify: true
 {{ end }}
   relabel_configs:
-  - source_labels: [__meta_kubernetes_pod_ip]
+  - source_labels: [__address__]
     replacement: ${1}:9091
     target_label: instance
   - target_label: __address__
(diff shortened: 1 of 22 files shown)


service/controller/resource/alert/test/case-0-cluster-api.golden
```diff
@@ -0,0 +1,25 @@
+metadata:
+  creationTimestamp: null
+  labels:
+    cluster_id: bob
+  name: example-rules
```
omegas27, line 5: Why is this test generating the old alert?

files/templates/additional-scrape-configs.template.yaml
```diff
@@ -259,7 +257,7 @@
   - target_label: app
     replacement: etcd
   - source_labels: ['__address__']
-    target_label:  'instance'
+    target_label: 'instance'
```
TheoBrigitte, line 260: why is there quotes here ?

TheoBrigitte (approved):
Looks fine, but I'd like to understand why we sometime use `__address__` and some other time something else?

files/templates/additional-scrape-configs.template.yaml
omegas27: We have some for role: nodes so we need to change the port and managed apps as well
Also, the kubernetes component address is only the IP as kubernetes knows the port to use so the port is not scraped. I changed it to address still instead of podIP

omegas27:
I fixed the alert it is not triggering on any provider.

PR #135, opened 2020-09-23
Add ingress resource

mig4:
This provides access to the Prometheus pods via a corresponding HTTP
path on ingress proxy so as to avoid the need to access it via port
forwarding.

Closes giantswarm/giantswarm#13366

### DIFF
--- a/service/controller/resource/ingress/client.go
+++ b/service/controller/resource/ingress/client.go
@@ -1,0 +1,25 @@
+package ingress
+
+import (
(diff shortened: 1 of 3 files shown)


mig4:
Hey @giantswarm/team-atlas this is an initial boilerplate for the ingress resource. Just in case someone gets to it before me in the morning, let's roll from here, unless there are objections. Builds fail because stuff still needs to be implemented.

mig4:
Superseded by #136

PR #138, opened 2020-09-24
TLS certificates

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/11524

Instead of re-implementing a certificate resource which would do the exact same thing as the one already present, I made it re-usable by passing it 3 new function which generate the source secret name & namespace as well as the target secret name.

Tested on `gauss`

```bash
$ k get -n=4i0wf-prometheus secret monitoring-prometheus
NAME                    TYPE                DATA   AGE
monitoring-prometheus   kubernetes.io/tls   3      60s
```

NOTE: I used `monitoring-prometheus` during my test but the final name is `prometheus-certificates`.


service/controller/resource/resource.go
```diff
@@ -44,14 +45,35 @@ func New(config Config) ([]resource.Interface, error) {
...
-	var certificatesResource resource.Interface
+	var apiCertificatesResource resource.Interface
 	{
 		c := certificates.Config{
-			K8sClient: config.K8sClient,
-			Logger:    config.Logger,
+			Name:                "api-certificates",
+			K8sClient:           config.K8sClient,
+			Logger:              config.Logger,
+			SourceNameFunc:      key.Namespace,
```
TheoBrigitte, line 54: This sounds funny, but this function just returns `<cluster id>-prometheus` which is exactly what we want here. Maybe renaming it to `key.PrometheusSuffix` would make sense.

omegas27: I agree we should rename it but maybe not prometheus suffix as this name can be used for every cluster. Maybe PrometheusName or PrometheusInstanceName would make sense?

omegas27:
Just for info, is there any specific reason why you preferred to copied the one in the default namespace instead of creating the secret in the namespace? We could not rely on it being created which would remove a dependency we do not really need 
https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus/templates/prometheus/certs-secret.yaml


Also, is there a reason to create the secret when the installation supports letsencrypt?

TheoBrigitte:
> Just for info, is there any specific reason why you preferred to copied the one in the default namespace instead of creating the secret in the namespace? We could not rely on it being created which would remove a dependency we do not really need
> https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus/templates/prometheus/certs-secret.yaml
> 
> Also, is there a reason to create the secret when the installation supports letsencrypt?

So we do not manage the secret ourselves but let ingress and cert-manager do the job. Yeah let's try that.

omegas27 (approved):

PR #141, opened 2020-09-24
set external url

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/11524

### DIFF
--- a/main.go
+++ b/main.go
@@ -110,6 +110,7 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CrtFile, "", "Certificate file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.KeyFile, "", "Key file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.BaseDomain, "", "Base domain to create Prometheus Ingress resources under.")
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -110,6 +110,7 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CrtFile, "", "Certificate file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.KeyFile, "", "Key file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.BaseDomain, "", "Base domain to create Prometheus Ingress resources under.")
+	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Address, "", "Address to access Prometheus UI.")
```
omegas27, line 113: What is the difference between address and host?

mig4: Or `address` and `baseDomain` (which was the original option we had for setting external URL with the original _Ingress_ resource)?

omegas27 (approved):

main.go
omegas27: Actually, we have both in draughtman CM, address is the same as the host but with https in front :D

TheoBrigitte: address has the protocol

```yaml
          Prometheus:
            Address: https://prometheus.g8s.gauss.eu-west-1.aws.gigantic.io
            Host: prometheus.g8s.gauss.eu-west-1.aws.gigantic.io
```

PR #143, opened 2020-09-25
Add prometheus tls secret on non lets encrypt

omegas27:


helm/prometheus-meta-operator/templates/ingress-oauth.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: {{ include "resource.default.name" . }}
-  namespace: "monitoring"
+  namespace: {{ include "resource.default.namespace" . }}
```
TheoBrigitte, line 5: I was not so sure about this, because it does not matter where pmo is installed. This ingress always has to be in `monitoring` namespace.

TheoBrigitte (approved):

helm/prometheus-meta-operator/templates/ingress-oauth.yaml
omegas27: Do you want me to change it back?

omegas27: I changed it back.

PR #150, opened 2020-09-29
Add promxy

omegas27:


JosephSalisbury:
hmm. do we wanna approach this similarly to how we do prometheus-operator? i reckon having an app that we deploy separately to control planes is a better approach (we can provide it to customers, it breaks up concerns etc.)

omegas27:
Well, sure, but this will be a pain to test as I need the app to be present on control planes, and edit the config map of this specific app every time a new cluster is added or removed. There is a tight coupling between the two unless we provide a controller around promxy

JosephSalisbury:
i'm not sure if i see much difference between PMO cruding some theoretical promxy-operator CRs, or cruding the promxy configmap, but agree it's easier to just get it going here, and we can always split it out later, sure

omegas27:
This is merely so I can experiment when I have some free time as our focus is on storage this week, splitting out should be easy yes, the app technically already exists so I could experiment with using an App CR directly though :D Let me do that instead

omegas27:
https://github.com/giantswarm/promxy-app

JosephSalisbury:
if we already have an app, adding it to the control plane catalogs is pretty straightforward

omegas27:
Tested on gorgoth

omegas27:
Once 0.0.2 of promxy app is up, we "only" need to create a ConfigMap for the app and add the server groups

helm/prometheus-meta-operator/templates/_resource.tpl
```diff
@@ -26,3 +22,4 @@ room for such suffix.
 {{- define "resource.default.namespace" -}}
 monitoring
 {{- end -}}
+
```
TheoBrigitte, line 25: What is this new line for ?

helm/prometheus-meta-operator/templates/rbac.yaml
```diff
@@ -128,7 +128,15 @@ rules:
       - configmaps
     verbs:
       - get
-      - get
+      - create
```
TheoBrigitte, line 131: I think we should not create the promxy configmap. It should be a guarantee that promxy configmap is present.

That's the only interface we should agree on: having a configmap in a given namespace with a given name.

service/controller/resource/promxy/app/create.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("checking if %s app exists", objectMeta.Name))
+	app, err := r.k8sClient.G8sClient().ApplicationV1alpha1().Apps(objectMeta.Namespace).Get(ctx, objectMeta.Name, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	app.Spec.Config.ConfigMap.Name = key.PromxyConfigMapName()
+	app.Spec.Config.ConfigMap.Namespace = key.PromxyConfigMapNamespace()
```
TheoBrigitte, line 26: Feels wrong to change this here, I'd rather have this being changed in the promxy app itself, or being a default in installation but not here. Again this go in the direction of my previous comment.

> I think we should not create the promxy configmap. It should be a guarantee that promxy configmap is present.
>
> That's the only interface we should agree on: having a configmap in a given namespace with a given name.

service/controller/resource/promxy/configmap/delete.go
```diff
@@ -0,0 +1,9 @@
+package configmap
+
+import (
+	"context"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	return nil
```
mig4, line 8: We don't want this deleted? Maybe something like only delete if empty? Should be good enough to avoid leftovers.

service/controller/resource/promxy/configmap/resource.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	data["values"] = content
+	return &corev1.ConfigMap{
+		ObjectMeta: objectMeta,
+		Data:       data,
+	}, nil
+}
```
TheoBrigitte, line 70: Again we would not need to have all this code to create promxy config map, if it would be provided by promxy app itself. This feels like a lot of code just to create a configmap.

service/controller/resource/promxy/servergroup/create.go
```diff
@@ -0,0 +1,61 @@
...
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	promxyConfiguration, err := promxy.Deserialize(configmap.Data["values"])
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	url, err := url.Parse("https://kubernetes.default:443")
```
mig4, line 38: Pardon my ignorance. Does this work? What does it mean?

service/controller/resource/promxy/servergroup.go
```diff
@@ -0,0 +1,172 @@
...
+
+	config_util "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/discovery/kubernetes"
+	"github.com/prometheus/prometheus/pkg/relabel"
+)
+
+// Config is the configuration for a ServerGroup that promxy will talk to.
+// This is where the vast majority of options exist.
+type ServerGroup struct {
```
mig4, line 15: Question, would it not be better to import this from [github.com/jacksontj/promxy/pkg/servergroup](https://github.com/jacksontj/promxy/blob/master/pkg/servergroup/config.go)?

TheoBrigitte, line 15: ServerGroup implementation should be replace with https://godoc.org/github.com/jacksontj/promxy/pkg/servergroup#Config

Or please explain what changed here, otherwise it's hard to track and see the difference.

service/controller/resource/promxy/servergroup/create.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
```
TheoBrigitte, line 41: This can be moved inside `r.toServerGroup`.

helm/prometheus-meta-operator/templates/_resource.tpl
```diff
@@ -11,10 +11,6 @@ room for such suffix.
 {{- .Release.Name | replace "." "-" | trunc 47 | trimSuffix "-" -}}
 {{- end -}}
 
-{{- define "resource.networkPolicy.name" -}}
```
mig4, line 14: If it's not related I'd leave it out.

service/controller/resource/promxy/types.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"gopkg.in/yaml.v2"
+)
+
+type Promxy struct {
+	Promxy PromxyConfig `yaml:"promxy"`
+}
+
+// PromxyConfig is the configuration for Promxy itself
+type PromxyConfig struct {
```
TheoBrigitte, line 12: This can be replaced with https://godoc.org/github.com/jacksontj/promxy/pkg/config#Config

TheoBrigitte (changes requested):
I'd advocate in favor of moving out proxmy configmap creationg and promxy app config from here. It would make it easier to interface pmo and promx by not having to change those.

service/controller/resource/promxy/servergroup/create.go
omegas27: Yes, basically, Kube creates a kubernetes service in the defautl namespace to access the api server. I am sure we can do better here but I wanted to make it work on friday. I had to add the api server url as https://github.com/prometheus/prometheus/blob/8dbfa14607a5cb138df43638f06c1a938c73df84/discovery/kubernetes/kubernetes.go#L167 requires the API server to be set

service/controller/resource/promxy/servergroup/create.go
omegas27: Sure, but technically, I think we can actually inject it from the kubernetes client when we start the operator. I just did not figure out how yet

service/controller/resource/promxy/types.go
omegas27: Sadly,no, as promxy is using an old prometheus dependency on 1.8 so the version we get by adding this to the project makes us need to acually import prometheus 1.8 instead of 2.20. Morever, the prometheus version is a fork made jacksontj/prometheus

omegas27: This is the part I really do not like

service/controller/resource/promxy/servergroup.go
```diff
@@ -0,0 +1,172 @@
...
+	// If we saw this same config in promxy (pointing at prometheus hosts instead of some exporter), we'd expect a similar behavior:
+	//   (1) only targets with the prod consul label would be included in the servergroup
+	//   (2) it would add a label to all returned series of this servergroup of datacenter with the value set to whatever the value of __meat_consul_dc was.
+	//
+	// So in reality its "the same", the difference is in prometheus these apply to the labels/targets of a scrape job,
+	// in promxy they apply to the prometheus hosts in the servergroup - but the behavior is the same.
+	RelabelConfigs []*relabel.Config `yaml:"relabel_configs,omitempty"`
+	// Hosts is a set of discovery.Config options that allow promxy to discover
+	// all hosts in the server_group
+	KubernetesSDConfigs []*kubernetes.SDConfig `yaml:"kubernetes_sd_configs,omitempty"`
```
omegas27, line 74: In the original jacksontj/promxy code, the used type is *discovery.Config[] which cannot be deserialized as Config is an interface and this needs to be inlinded. Maybe you have an idea on how to configure it back to *discovery.Config[] yaml:",inline"

helm/prometheus-meta-operator/templates/_resource.tpl
omegas27: Let me revert the change in this file

helm/prometheus-meta-operator/templates/rbac.yaml
omegas27: I agree, but can we do that with our current tooling as promxy app is installed using the app-chart mechanism?

service/controller/resource/promxy/app/create.go
omegas27: I agree, I wanted to get this out first and I do not know how to configure architect to create a CM to do it.

service/controller/resource/promxy/configmap/delete.go
omegas27: Well, i am not sure we want to delete it anyway. As per discussions with @TheoBrigitte the lifecycle of this cm should be outside of this app

service/controller/resource/promxy/servergroup.go
omegas27: Sure, what changed is that promxy is using a fork of prometheus 1.8 as a dependency while we rely on 2.20 and a lot of things changed since then so I had to do a few changes to the ServerGroup struct like: // Hosts is a set of discovery.Config options that allow promxy to discover
	// all hosts in the server_group
	KubernetesSDConfigs []*kubernetes.SDConfig `yaml:"kubernetes_sd_configs,omitempty"`

service/controller/resource/promxy/servergroup/create.go
mig4: Cool, TIL 👆

omegas27: I changed it to use the host in the k8s client but now you know :D

mig4: If we can rely on kube doing this, I'd say this is actually a pretty nice and consistent way to access the API server. But sure, up to you.

service/controller/resource/promxy/servergroup.go
mig4: It seems to be `"github.com/prometheus/prometheus/discovery/config".ServiceDiscoveryConfig` here https://github.com/jacksontj/promxy/blob/master/pkg/servergroup/config.go#L86, or am I looking at a wrong place?

omegas27: So, yes, in prometheus 1.8 code which translates to https://github.com/prometheus/prometheus/blob/924ce978b734e8a738ef28e2207997a34710cfa6/discovery/discovery.go#L58 in prom 2.20 code

omegas27:
I like this new code way more than the old one if you want to take a look now :D

JosephSalisbury (approved):
IIWFYLGTM

service/controller/resource/promxy/create.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/giantswarm/prometheus-meta-operator/service/key"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	configMap, err := r.getConfigMap(ctx, obj)
+	if err != nil {
+		return microerror.Mask(err)
+	} else if configMap == nil {
+		return nil // Missing config map, we return immediately
```
mig4, line 16: So the expectation is that we'll just re-try on next reconciliation right? Maybe just log a message?

TheoBrigitte: Either there's an error or configmap is not nil. But there should not be a case where there's no error and no configmap.

service/controller/resource/promxy/create.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/prometheus-meta-operator/service/key"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	configMap, err := r.getConfigMap(ctx, obj)
```
TheoBrigitte, line 12: Need some log message along the line of: checking if promxy needs to be updated

service/controller/resource/promxy/delete.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/giantswarm/prometheus-meta-operator/service/key"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	configMap, err := r.getConfigMap(ctx, obj)
+	if err != nil {
+		return microerror.Mask(err)
+	} else if configMap == nil {
+		return nil // Missing config map, we return immediately
```
TheoBrigitte, line 16: Same remark

service/controller/resource/promxy/delete.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/prometheus-meta-operator/service/key"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	configMap, err := r.getConfigMap(ctx, obj)
```
TheoBrigitte, line 12: Needs logging too

service/controller/resource/promxy/resource.go
```diff
@@ -0,0 +1,152 @@
...
+	"time"
+
+	"github.com/giantswarm/k8sclient/v4/pkg/k8sclient"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/discovery/kubernetes"
+	"github.com/prometheus/prometheus/pkg/relabel"
+	"gopkg.in/yaml.v2"
```
TheoBrigitte, line 16: Please instead use https://github.com/ghodss/yaml

It's a drop in replacement and handles better empty fields, we use this everywhere

service/controller/resource/promxy/create.go
mig4: Do you mean this should never happen?

service/controller/resource/promxy/resource.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	apiServerHost := r.k8sClient.RESTConfig().Host
+	apiServerURL, err := url.Parse(apiServerHost)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return &ServerGroup{
+		AntiAffinity:   time.Second * 10,
```
TheoBrigitte, line 79: What is this for? Maybe a comment would help.

service/controller/resource/promxy/resource.go
```diff
@@ -0,0 +1,152 @@
...
+					Names: []string{
+						key.Namespace(cluster),
+					},
+				},
+			},
+		},
+	}, nil
+}
+
+func (r *Resource) getConfigMap(ctx context.Context, obj interface{}) (*v1.ConfigMap, error) {
```
TheoBrigitte, line 121: I do not see the purpose of this function.

service/controller/resource/promxy/create.go
omegas27: Well, I was checking in the getConfigMap method if the configMap is not found and logging there to avoid duplication but I can move that in the main resource code: 	
```
if apierrors.IsNotFound(err) {
	    r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("config map %s does not exists", key.PromxyConfigMapName()))
		return nil, nil
	} else if err != nil {
	}
```

service/controller/resource/promxy/resource.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, nil
+	} else if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return configMap, nil
+}
+
+func (r *Resource) readFromConfig(configMap *v1.ConfigMap) (*Promxy, error) {
+	content := configMap.Data[key.PromxyConfigFileName()]
```
TheoBrigitte, line 136: It would be safer to check if the key exists before accessing it to avoid panic.

service/controller/resource/promxy/types.go
```diff
@@ -0,0 +1,208 @@
...
+		if val.PathPrefix == group.PathPrefix {
+			index = key
+		}
+	}
+	p.ServerGroups = append(p.ServerGroups[:index], p.ServerGroups[index+1:]...)
+}
+
+// ServerGroup is the configuration for a ServerGroup that promxy will talk to.
+// This is where the vast majority of options exist.
+type ServerGroup struct {
```
mig4, line 53: Would you be able to at least add a comment with some reasoning as to why we're using a struct copied over from promxy, instead of importing it? Just so that if we come back and want to re-visit this in a while we have some context of what to watch out for. I recon this will get out of sync sooner or later.

service/controller/resource/promxy/resource.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func (r *Resource) updateConfig(ctx context.Context, configMap *v1.ConfigMap, config *Promxy) error {
+	bytes, err := yaml.Marshal(config)
+	if err != nil {
+		return err
+	}
+
+	configMap.Data[key.PromxyConfigFileName()] = string(bytes)
+	_, err = r.k8sClient.K8sClient().CoreV1().ConfigMaps(key.PromxyConfigMapNamespace()).Update(ctx, configMap, metav1.UpdateOptions{})
```
TheoBrigitte, line 149: I'd like to see the usual error handling here, otherwise we loose valuable debugging information.
```suggestion
	_, err = r.k8sClient.K8sClient().CoreV1().ConfigMaps(key.PromxyConfigMapNamespace()).Update(ctx, configMap, metav1.UpdateOptions{})
	if err != nil {
		return nil, microerror.Mask(err)
	}
```

service/controller/resource/promxy/types.go
TheoBrigitte: Can we at least please add a comment to tell where this comes from.
Maybe using replace in go.mod would do the trick of using the right dependency for us. It's really not good to have to copy all this code in here.

TheoBrigitte (changes requested):

service/controller/resource/promxy/resource.go
JosephSalisbury: from docs: '# anti-affinity for merging values in timeseries between hosts in the server_group' - so, I'd actually dump this, we only have one server per servergroup, so we don't need to worry about anti-affinitiy

service/controller/resource/promxy/create.go
TheoBrigitte: Yes please move this here, I see no reason to hide such simple logic.

service/controller/resource/promxy/resource.go
```diff
@@ -0,0 +1,152 @@
...
+
+	return microerror.Mask(err)
```
TheoBrigitte, line 151: I'd like to see the usual error handling here, otherwise we loose valuable debugging information.

```suggestion
	_, err = r.k8sClient.K8sClient().CoreV1().ConfigMaps(key.PromxyConfigMapNamespace()).Update(ctx, configMap, metav1.UpdateOptions{})
        if err != nil {
		return nil, microerror.Mask(err)
	}

	return nil
```

service/controller/resource/promxy/types.go
TheoBrigitte: Where is the original promxy then ?

service/controller/resource/promxy/resource.go
omegas27: Mask is doing the exact same thing though: 
if err == nil {
    return nil
}

return mask(err)

service/controller/resource/promxy/types.go
omegas27: I added it over the Promxy struct

service/controller/resource/promxy/create.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/prometheus-meta-operator/service/key"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	r.logger.LogCtx(ctx, "level", "debug", "message", "checking if promxy config map already exists")
```
TheoBrigitte, line 14: maybe ?

```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", "checking if promxy configmap needs to be updated")
```

service/controller/resource/promxy/create.go
```diff
@@ -0,0 +1,53 @@
...
+	configMap, err := r.k8sClient.K8sClient().CoreV1().ConfigMaps(key.PromxyConfigMapNamespace()).Get(ctx, key.PromxyConfigMapName(), metav1.GetOptions{})
+
+	if apierrors.IsNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "promxy needs to be updated")
+		return nil
```
TheoBrigitte, line 19: This is not where the decision is being made to update or not the configmap. And why do we return nil ?
I think we should return any error which occurs here: when we do not find the configmap there's a bigger issue and this resource cannot do its job.

service/controller/resource/promxy/create.go
```diff
@@ -0,0 +1,53 @@
...
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "reading promxy config")
```
TheoBrigitte, line 24:
```suggestion
```

service/controller/resource/promxy/create.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	serverGroup, err := r.toServerGroup(cluster)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "checking if server group must be added")
+	if !config.Promxy.contains(serverGroup) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "adding server group")
```
TheoBrigitte, line 42:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "promxy configmap needs to be updated")
		r.logger.LogCtx(ctx, "level", "debug", "message", "adding server group")
```

service/controller/resource/promxy/create.go
```diff
@@ -0,0 +1,53 @@
...
+		config.Promxy.add(serverGroup)
+
+		err = r.updateConfig(ctx, configMap, config)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "added server group")
+	}
```
TheoBrigitte, line 51:
```suggestion
	} else {
        	r.logger.LogCtx(ctx, "level", "debug", "message", "promxy do not need to be updated")
        }
```

service/controller/resource/promxy/create.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	serverGroup, err := r.toServerGroup(cluster)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "checking if server group must be added")
```
TheoBrigitte, line 40:
```suggestion
```

service/controller/resource/promxy/delete.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/prometheus-meta-operator/service/key"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	r.logger.LogCtx(ctx, "level", "debug", "message", "checking if promxy config map already exists")
```
TheoBrigitte, line 14:
```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", "checking if promxy configmap needs to be updated")
```

service/controller/resource/promxy/delete.go
```diff
@@ -0,0 +1,52 @@
...
+	configMap, err := r.k8sClient.K8sClient().CoreV1().ConfigMaps(key.PromxyConfigMapNamespace()).Get(ctx, key.PromxyConfigMapName(), metav1.GetOptions{})
+
+	if apierrors.IsNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "promxy needs to be updated")
+		return nil
```
TheoBrigitte, line 19: Same remark as in EnsureCreated

service/controller/resource/promxy/delete.go
```diff
@@ -0,0 +1,52 @@
...
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "reading promxy config")
```
TheoBrigitte, line 24:
```suggestion
```

service/controller/resource/promxy/delete.go
```diff
@@ -0,0 +1,52 @@
...
+		return microerror.Mask(err)
+	}
+
+	serverGroup, err := r.toServerGroup(cluster)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if config.Promxy.contains(serverGroup) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "removing server group")
```
TheoBrigitte, line 41:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "promxy configmap needs to be updated")
		r.logger.LogCtx(ctx, "level", "debug", "message", "removing server group")
```

service/controller/resource/promxy/types.go
TheoBrigitte: Beautiful thanks :heart:

TheoBrigitte (approved):
Looks about right, minor stuff about log messages.

service/controller/resource/promxy/create.go
omegas27: Should I still log that promxy needs to be updated?

PR #151, opened 2020-09-30
Configure retention and increase volume size

omegas27:


omegas27:
Tested on gorgoth:
```
      --storage.tsdb.retention.size=45GB
      --config.file=/etc/prometheus/config_out/prometheus.env.yaml
      --storage.tsdb.path=/prometheus
      --storage.tsdb.retention.time=2w
      --web.enable-lifecycle
      --storage.tsdb.no-lockfile
      --web.external-url=https://prometheus.g8s.gorgoth.gridscale.kvm.gigantic.io/gorgoth
      --web.route-prefix=/gorgoth
      --storage.tsdb.wal-compression
```

helm/prometheus-meta-operator/templates/configmap.yaml
```diff
@@ -24,9 +24,12 @@ data:
       prometheus:
         address: {{ .Values.Installation.V1.Monitoring.Prometheus.Address }}
         baseDomain: {{ .Values.Installation.V1.Monitoring.Prometheus.Host }}
+        retention:
+          duration: 2w
+          size: 90GB
         storage:
           createPVC: {{ and .Values.Installation.V1.Monitoring.Prometheus.Storage.Enabled .Values.Installation.V1.Monitoring.Prometheus.Storage.CreatePVC }}
-          size: 20Gi
```
TheoBrigitte, line 29: Resizing pvc via statefulset is not supported yet in kubernetes: https://github.com/kubernetes/kubernetes/issues/68737

TheoBrigitte (changes requested):

helm/prometheus-meta-operator/templates/configmap.yaml
omegas27: I am not sure what you want me to change here. We still need to increase it to 100Gi

TheoBrigitte: I'm just saying this does not work for existing cluster. The volume size won't be changed. So we need a workaround.

omegas27: So we would need a volume deleter if the volume size is 20Gi right?

TheoBrigitte: Yeah right, official doc says you have to manually delete the pvc created by a stateful set when cleaning it (also valable for resizing): https://kubernetes.io/docs/tasks/run-application/delete-stateful-set/#complete-deletion-of-a-statefulset

helm/prometheus-meta-operator/templates/rbac.yaml
```diff
@@ -49,6 +49,21 @@ rules:
     verbs:
       - create
       - delete
+  - apiGroups:
```
omegas27, line 52: Can you add a todo here to say we need to remove it once the resize is done and not leave some leftover rights?

main.go
```diff
@@ -112,7 +112,9 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.BaseDomain, "", "Base domain to create Prometheus Ingress resources under.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Address, "", "Address to access Prometheus UI.")
 	daemonCommand.PersistentFlags().Bool(f.Service.Prometheus.Storage.CreatePVC, false, "Should the operator create a PVC for storage.")
-	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Storage.Size, "20Gi", "Storage size for prometheus.")
+	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Storage.Size, "50Gi", "Storage size for prometheus.")
```
omegas27, line 115: Should this not be 100Gi and the retention size would be 90 as well?

service/controller/resource/volume-resize-hack/create.go
```diff
@@ -0,0 +1,88 @@
+package volumeresizehack
```
omegas27, line 1: I like the direction of this

omegas27:
I just commented a few things but this looks really nice

helm/prometheus-meta-operator/templates/rbac.yaml
TheoBrigitte: done

main.go
TheoBrigitte: You're probably right, but I always got 100Gi during my tests.

TheoBrigitte: That's because we set this value in the helm chart.

helm/prometheus-meta-operator/templates/rbac.yaml
omegas27: :1st_place_medal:

main.go
TheoBrigitte: done

TheoBrigitte:
This was successfully tested. The workflow might seems strange and goes as follow:

1. We detect PVC exist and does not have the same size as defined in the StS
2. Delete the PVC
3. Scale down StS; so pod is deleted and volume is freed
4. Wait for PVC to be deleted
5. Scale down StS again; we need this because prometheus-operator scale back up the StS and the statefuleset-controller does not kick to re-create the PVC from StS volume claim template. Also trying to pause prometheus-operator reconciliation via the pause field results in the StS being deleted when trying to scale it down.
6. prometheus-operator scale back up the StS once more, statefulset-controller kicks in and re-create the PVC.

omegas27:
I cannot approve but this looks legit to me

PR #153, opened 2020-10-05
upgrade prometheus-operator import

TheoBrigitte:
prometheus-operator has moved from [`github.com/coreos/prometheus-operator`](https://github.com/coreos/prometheus-operator) to [`github.com/prometheus-operator/prometheus-operator`](https://github.com/prometheus-operator/prometheus-operator)

In order to be able to upgrade this we need to replace import path and update the version.

This currently fails due to a tagging issue upstream which has be raised here: https://github.com/prometheus-operator/prometheus-operator/issues/3530

Once the issue is fixed update this PR with:

```bash
$ go get -u github.com/prometheus-operator/prometheus-operator
$ go mod tidy
$ # check go build -v
$ commit the result
```

#### What was done ?

* find/replace `github.com/coreos/prometheus-operator` with `github.com/prometheus-operator/prometheus-operator`
* upgrade `prometheus-operator` from `v0.41.1` to `v0.43.0`
* fix `InsecureSkipVerify` field which moved under `SafeTLSConfig`
* upgrade `sigs.k8s.io/controller-runtime` from `v0.6.1` to `v0.6.3`

This also means we can now configure web page titles ;)


omegas27 (approved):
I think that means we can specify the web page title now :D

TheoBrigitte:
> I think that means we can specify the web page title now :D

Yes we can

PR #158, opened 2020-10-12
Unit test Server Group serialization

omegas27:


TheoBrigitte (approved):
Looks great to me, and I like the alignement with upstream `sigs.k8s.io/yaml`.

PR #160, opened 2020-10-12
add alertmanager config

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/13093

This PR configure each prometheus to talk to our central `alertmanager` in the `monitoring` namespace.
It does so by:

1. Adding the config for the alertmanager endpoint in a secret called `alertmanager-config` in the `<cluster>-prometheus` namespace which contains

```yaml
- static_configs:
  - targets:
    - alertmanager.monitoring:9093
  scheme: http
  timeout: 10s
  api_version: v1
```
as in https://github.com/giantswarm/g8s-prometheus/blob/ba3bcade39849f9dbacecf1d45530805b8ec761d/helm/g8s-prometheus/templates/prometheus/prometheus-configmap.yaml#L36-L40

2. Configure prometheus via `Prometheus.Spec.AdditionalAlertManagerConfigs` field, see https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#prometheusspec
I used this because `Alerting` field only allow configuring an alertmanager endpoint which is insde  targeted prometheus, and additional scrape config didn't allowed me to add a new alertmanager endpoint.

#### Tested successfully

* new prometheus uses `alertmanager.monitoring` url to access alertmanager endpoint in the monitoring namespace.

image

* alertmanager receives alert from new prometheuses

image

### DIFF
--- a/service/controller/resource/alertmanagerconfig/test/case-0-cluster-api.golden
+++ b/service/controller/resource/alertmanagerconfig/test/case-0-cluster-api.golden
@@ -1,0 +1,13 @@
+metadata:
+  creationTimestamp: null
+  name: alertmanager-config
+  namespace: bob-prometheus
+stringData:
+  alertmanager-additional.yaml: |
+    - static_configs:
+      - targets:
+        - alertmanager.monitoring:9093
+      scheme: http
+      timeout: 10s
+      api_version: v1
+type: Opaque
--- a/service/controller/resource/alertmanagerconfig/test/case-0-cluster-api.golden
+++ b/service/controller/resource/alertmanagerconfig/test/case-0-cluster-api.golden
@@ -1,13 +1,6 @@
-metadata:
-  creationTimestamp: null
-  name: alertmanager-config
-  namespace: bob-prometheus
-stringData:
-  alertmanager-additional.yaml: |
-    - static_configs:
-      - targets:
-        - alertmanager.monitoring:9093
-      scheme: http
-      timeout: 10s
-      api_version: v1
-type: Opaque
+- static_configs:
+  - targets:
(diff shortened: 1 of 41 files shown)


TheoBrigitte:
Tested successfully

image

image

service/controller/resource/alertmanagerconfig/test/case-0-cluster-api.golden
```diff
@@ -0,0 +1,13 @@
+metadata:
```
omegas27, line 1: Is that not missing a kind?

service/controller/resource/alertmanagerconfig/test/case-0-cluster-api.golden
```diff
@@ -0,0 +1,13 @@
...
+  creationTimestamp: null
+  name: alertmanager-config
+  namespace: bob-prometheus
+stringData:
+  alertmanager-additional.yaml: |
```
omegas27, line 6: Also, should the content not be encoded here?

omegas27:
Nice job

service/controller/resource/alertmanagerconfig/test/case-0-cluster-api.golden
TheoBrigitte: Yeah that's right

service/controller/resource/alertmanagerconfig/test/case-0-cluster-api.golden
TheoBrigitte: true, but this is not relevant anymore.

TheoBrigitte:
Tested successfully for CP prometheus

image

PR #162, opened 2020-10-13
Remove Server Group if it needs to be updated

omegas27:


TheoBrigitte:
That's the reason why the promxy configmap kept growing, right ? Does this fixes https://github.com/giantswarm/giantswarm/issues/13675 ?

TheoBrigitte (approved):

omegas27:
This is the only thing that would make sense to me for the cm issue to happen. This would cause PMO to OOM as the configmap in memory becomes huge. I guess the api server ip changes so the structs are different.

omegas27:
Let's clean up the godsmack cm and check if that solves the issue

PR #163, opened 2020-10-13
Refactor common labels

omegas27:


service/controller/resource/servicemonitor/service/apiserver.go
```diff
@@ -37,14 +37,6 @@ func APIServer(cluster metav1.Object, provider string, installation string) *pro
...
 							TargetLabel:  "app",
 						},
-						{
-							SourceLabels: []string{"__meta_kubernetes_namespace"},
-							TargetLabel:  "namespace",
-						},
-						{
-							SourceLabels: []string{"__meta_kubernetes_pod_name"},
-							TargetLabel:  "pod_name",
-						},
```
TheoBrigitte, line 47: Why are those gone ?

files/templates/_common.yaml
```diff
@@ -0,0 +1,25 @@
+{{- define "_node_role" -}}
```
TheoBrigitte, line 1: Please move this to a `_node_role.yaml` to keep things oraganized.

TheoBrigitte (approved):
Good to see more label alignement

service/controller/resource/servicemonitor/service/apiserver.go
omegas27: Prometheus operator is adding them for ServiceMonitors

omegas27: https://github.com/prometheus-operator/prometheus-operator/blob/55baf034c431ed2c78d950b187f7d8b34dd06860/pkg/prometheus/promcfg.go#L567

PR #164, opened 2020-10-14
clean up promxy configmap

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/13486

This will actually clean up the promxy configmap so we don't have to do it manually.

### DIFF
--- a/service/controller/resource/promxy/types.go
+++ b/service/controller/resource/promxy/types.go
@@ -54,13 +54,15 @@ func (p *PromxyConfig) add(group *ServerGroup) {
 }
 
 func (p *PromxyConfig) remove(group *ServerGroup) {
-	var index int
-	for key, val := range p.ServerGroups {
-		if val.PathPrefix == group.PathPrefix {
-			index = key
+	var s []*ServerGroup
+
+	for _, val := range p.ServerGroups {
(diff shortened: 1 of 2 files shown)


service/controller/resource/promxy/types.go
```diff
@@ -54,13 +54,15 @@ func (p *PromxyConfig) add(group *ServerGroup) {
 }
 
 func (p *PromxyConfig) remove(group *ServerGroup) {
-	var index int
-	for key, val := range p.ServerGroups {
-		if val.PathPrefix == group.PathPrefix {
-			index = key
+	var s []*ServerGroup
```
omegas27, line 57: This makes it easier to read for sure :D For my curiosity, was there any issue with the existing code?

omegas27 (approved):

service/controller/resource/promxy/types.go
TheoBrigitte: Existing code was correct, but I thought this would clean up the configmap automatically. After testing I realized this won't work because resources might be gone. So we have to clean up manually.

omegas27: Thanks :)

PR #168, opened 2020-10-14
Add label to not monitor a cluster

omegas27:


omegas27:
Tested on godsmack

TheoBrigitte (approved):
Looks too easy to be true, but seems good.

Also review would have been easier with the refactoring done in another PR.

omegas27:
You are right. I did that for my first implementation and forgot to move it to another PR

PR #173, opened 2020-10-15
add heartbeat alert

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/13302

Add heartbeat alert in Prometheus.

New `installation` and `cluster` label will be used for routing in alertmanager.


service/controller/resource/alert/rules/heartbeat.go
```diff
@@ -0,0 +1,30 @@
...
+func Heartbeat(obj metav1.Object, installation string) promv1.RuleGroup {
+	return promv1.RuleGroup{
+		Name: "heartbeat",
+		Rules: []promv1.Rule{
+			promv1.Rule{
+				Alert: "Heartbeat",
+				Expr:  intstr.FromString(`time() - max(container_start_time_seconds{cluster_type="host", namespace="monitoring", container="prometheus"}) > 20 * 60`),
+				Labels: map[string]string{
+					"installation": installation,
+					"cluster":      key.ClusterID(obj),
```
omegas27, line 20: So we have one alert per prometheus?

TheoBrigitte: Yes, each prometheus fire its own heartbeat.

* 1 heartbeat alert per prometheus
* alertmanager receives multiple heartbeat and route them to different opsgenie endpoints
* opsgenie has multiple heartbeat endpoint configured

omegas27: Ok, but they are all throwing a heartbeat if g8s-prometheus is down and not themselves as we are using cluster_type host and not the correct cluster id. Is that right?

TheoBrigitte: We're not using `cluster_type`

current heartbeat looks like:

```yaml
groups:
- name: heartbeat
  rules:
  - alert: Heartbeat
    expr: time() - max(container_start_time_seconds{cluster_type="host", namespace="monitoring", container="prometheus"}) > 20 * 60
    labels:
      severity: heartbeat
      team: atlas
```

and routing in alertmanager is done as follow; anything which matches `severity: heartbeat` goes to `receiver: heartbeat`
```yaml
      - receiver: heartbeat
        match:
          severity: heartbeat
        continue: false
        group_interval: 1s
        group_wait: 1s
        repeat_interval: 15s
```
which then ping opsgenie
```yaml
    - name: heartbeat
      webhook_configs:
      - url: https://api.opsgenie.com/v2/heartbeats/{{ .Values.Installation.V1.Monitoring.Prometheus.Heartbeat.Name }}/ping
        http_config:
          basic_auth:
            password: {{ .Values.Installation.V1.Secret.Prometheus.Heartbeat.OpsgenieKey }}
        send_resolved: false
```

But this makes me think that re-using `severity` is not the most clever thing. We can change this to `type: heartbeat`. So we do not have to touch the old setup.

omegas27: Is the promql query not checking the health of g8s-prometheus tough `container_start_time_seconds{cluster_type="host", namespace="monitoring", container="prometheus"}` ? Also, why do we not use the opsgenie_receiver type? https://prometheus.io/docs/alerting/latest/configuration/#opsgenie_config

TheoBrigitte: Yeah right, this query will fail, we need to replace this.

The `opsgenie_receiver` only does send alert, and cannot handle heartbeats. They are 2 different concept in opsgenie API.

omegas27 (approved):

PR #177, opened 2020-10-16
add key values for heartbeats

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/13302

PR #180, opened 2020-10-16
add heartbeatrouting resource

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/13302
Needs: https://github.com/giantswarm/prometheus-meta-operator/pull/177 https://github.com/giantswarm/prometheus-meta-operator/pull/178 https://github.com/giantswarm/prometheus-meta-operator/pull/179

This PR adds `heartbeatrouting` resource in charge of configuring alertmanager to correctly route new heartbeats to their respective endpoints in opsgenie.

### DIFF
--- a/service/controller/resource/alert/rules/heartbeat.go
+++ b/service/controller/resource/alert/rules/heartbeat.go
@@ -22,7 +22,7 @@ func Heartbeat(obj metav1.Object, installation string) promv1.RuleGroup {
 					"team":                "atlas",
 				},
 				Annotations: map[string]string{
-					"description": "This alert is used to ensure the entire alerting pipeline is functionnal.",
+					"description": "This alert is used to ensure the entire alerting pipeline is functional.",
 				},
 			},
 		},
--- a/service/controller/resource/heartbeat/create.go
+++ b/service/controller/resource/heartbeat/create.go
@@ -68,8 +68,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		r.logger.LogCtx(ctx, "level", "debug", "message", "heartbeat is up to date")
 	}
 
-	// Trigger initial ping to ensure opsgenie alerts when an issue arise.
-	// This check if alertmanager is up and its config is up to date
+	// The initial ping to the heartbeat is there to move the heartbeat from inactive to active.
 	_, err = r.heartbeatClient.Ping(ctx, desired.Name)
 	if err != nil {
 		return microerror.Mask(err)
--- a/service/controller/resource/heartbeat/create.go
+++ b/service/controller/resource/heartbeat/create.go
@@ -68,5 +68,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		r.logger.LogCtx(ctx, "level", "debug", "message", "heartbeat is up to date")
 	}
 
+	// Trigger initial ping to ensure opsgenie alerts when an issue arise.
+	// This check if alertmanager is up and its config is up to date
+	_, err = r.heartbeatClient.Ping(ctx, desired.Name)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
 	return nil
 }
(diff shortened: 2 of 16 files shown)


JosephSalisbury:
would it be possible to separate the copied library code into a separate PR? this is pretty tough to review

omegas27:
@JosephSalisbury  This should help with the review

service/controller/resource/alert/rules/heartbeat.go
```diff
@@ -22,7 +22,7 @@ func Heartbeat(obj metav1.Object, installation string) promv1.RuleGroup {
 					"team":                "atlas",
 				},
 				Annotations: map[string]string{
-					"description": "This alert is used to ensure the entire alerting pipeline is functionnal.",
+					"description": "This alert is used to ensure the entire alerting pipeline is functional.",
```
JosephSalisbury, line 25: <3

service/controller/resource/heartbeat/create.go
```diff
@@ -68,5 +68,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		r.logger.LogCtx(ctx, "level", "debug", "message", "heartbeat is up to date")
 	}
 
+	// Trigger initial ping to ensure opsgenie alerts when an issue arise.
```
JosephSalisbury, line 71: This comment doesn't follow to me, or feels more like 'this is what all the code does'

Language more towards 'the initial ping to the heartbeat is there to move the heartbeat from inactive to active' would make sense to me

service/controller/resource/heartbeatrouting/create.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	cfg, routeUpdated, err := route.EnsureCreated(cfg, cluster, r.installation)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if receiverUpdated || routeUpdated {
```
JosephSalisbury, line 36: Do we need a negation here?

e.g: `if !receiverUpdated || !routeUpdate`? Otherwise it sounds like we update it if we updated it

service/controller/resource/heartbeatrouting/delete.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	cfg, routeUpdated, err := route.EnsureDeleted(cfg, cluster, r.installation)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if receiverUpdated || routeUpdated {
```
JosephSalisbury, line 36: Similarly here

service/controller/resource/heartbeatrouting/create.go
JosephSalisbury: Or having the booleans named the other way around (i.e: receiverNeedsUpate)

JosephSalisbury (commented):
here's some notes yo

service/controller/resource/heartbeat/create.go
omegas27: :+1:

service/controller/resource/heartbeatrouting/create.go
omegas27: I think this could be renamed for sure. Basically, those imply that the alertmanagerconfig struct changed and the configmap needs to be updated. receiverUpdated would mean that one receiver was changed or was added to the list of receiver so we need to update the configmap to reflect the change in alertmanager

omegas27: This is fixed in the latest commit

service/controller/resource/heartbeatrouting/delete.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	cfg, routeNeedsUpdate, err := route.EnsureDeleted(cfg, cluster, r.installation)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	alertManagerConfigMapNeedsUpdate := receiverNeedsUpdate || routeNeedsUpdate
```
JosephSalisbury, line 36: niceeee

JosephSalisbury (approved):
lgtmiiwfy

PR #183, opened 2020-10-22
Add remote write to Cortex

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/13092


omegas27:
image
Tested on Gauss

helm/prometheus-meta-operator/templates/secret.yaml
```diff
@@ -8,5 +8,10 @@ metadata:
 stringData:
   secret.yml: |
     service:
+      prometheus:
+        remoteWrite:
+          basicAuth:
+            username: 10755
```
TheoBrigitte, line 14: Can we also have the username templated ? Nice to have, I am ok if we can't.

TheoBrigitte (approved):
LGTM when tested

TheoBrigitte:
> image
> Tested on Gauss

Why do we loose labels like `customer`, `installation`, `pipeline`, `provider`, and `region`? I think we do need those labels.

helm/prometheus-meta-operator/templates/secret.yaml
omegas27: Right now, not really, we can add that in the helm values but it is not in the installations repo. You can add a PR for it though :)

omegas27:
We are loosing some labels because what is used is the external labels witch we did not really fill per prometheus

helm/prometheus-meta-operator/templates/secret.yaml
TheoBrigitte: Ok fine

TheoBrigitte (approved):

omegas27:
image
tested multiple times on Gauss :D

PR #184, opened 2020-10-26
Add recording rules

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/13092


omegas27:
image
Tested on Gauss

TheoBrigitte (commented):
There are many changes here, this makes it hard to review. Could you please split this up into multiple smaller PRs? please

omegas27:
I had that on my TODO. I could not get to it but I will do it today.

omegas27:
@TheoBrigitte I split some of it into 
https://github.com/giantswarm/prometheus-meta-operator/pull/187
https://github.com/giantswarm/prometheus-meta-operator/pull/188

I will rebase after they are merged

service/controller/resource/alert/create.go
```diff
@@ -0,0 +1,59 @@
...
+
+	rules, err := getRules(obj, r.installation)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring the prometheus rules exists")
+	for _, desiredRule := range rules {
+		ruleName := desiredRule.GetName()
+		r.logger.LogCtx(ctx, "level", "debug", "message", "checking if rule needs to be created or updated", "rule", ruleName)
```
TheoBrigitte, line 27: We do not use separate keys to log additional information, but put everything under the `message` key. see https://github.com/giantswarm/fmt/blob/master/go/log_messages.md

service/controller/resource/alert/resource.go
```diff
@@ -35,66 +44,61 @@ func New(config Config) (*generic.Resource, error) {
...
+	files := bytes.Split(template, []byte("---"))
+
+	var rules []*promv1.PrometheusRule = make([]*promv1.PrometheusRule, 0)
+	for _, file := range files {
+		if len(bytes.TrimSpace(file)) > 0 {
+			var rule promv1.PrometheusRule = promv1.PrometheusRule{}
+
+			if err = yaml.Unmarshal(file, &rule); err != nil {
+				return nil, microerror.Mask(err)
+			}
```
TheoBrigitte, line 95: Given that this might fail when we provide invalid rule files, we should add unit test in order to validate the rule yaml template we are adding. But this can be done in a separate PR.

TheoBrigitte (approved):
Very nice, especially how we just template yaml and converts them back into CRs.

service/controller/resource/alert/create.go
omegas27: Alright, let me fix it before merging if circle ci allows me to

service/controller/resource/alert/resource.go
omegas27: Yes we should :D

omegas27:
Tested on gauss

PR #185, opened 2020-10-27
Fix kube-state-metrics scraping port

omegas27:


TheoBrigitte (approved):
Those ports are quite hidden, but let's go with this.

PR #193, opened 2020-11-04
prefer scheduling prometheus on non master nodes

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/14076

New prometheus eat less memory but still they might kill master node. And this was observed multiple times already.

To prevent this, this PR adds soft node affinity which would prefer not scheduling on master but not prevent it.


omegas27:
I think you forgot to run test -update

omegas27 (approved):

TheoBrigitte:
> I think you forgot to run test -update

yeah thanks <3

PR #192, opened 2020-11-03
Add Atlas rules

omegas27:
Towards: https://github.com/giantswarm/giantswarm/issues/14028

PR #195, opened 2020-11-04
fix wrong path in alert resource

TheoBrigitte:
fixes following error: unknown: html/template: pattern matches no files: `/opt/prometheus/meta-operator/files/templates/rules/**/*.yml`


omegas27:
Oh, I forgot that I fixed it in https://github.com/giantswarm/prometheus-meta-operator/pull/192

PR #196, opened 2020-11-04
fix alert creation

TheoBrigitte:
Fixes the following error: unknown: prometheusrules.monitoring.coreos.com "service-level" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update

This was due to the created object not being stored back into `currentRule` variable, which lead to :

1. the rule being updated right after its creation
2. resourceVersion being empty


JosephSalisbury (approved):
lgtmiiwfy

PR #197, opened 2020-11-05
Revert "Remote write to cortex (#190)"

mig4:
This reverts commit a7cab74b4f8154b7a71693d7ac758acc039c54a3.

Writing to cortex is not ready to be released yet, it has to be disabled
in g8s-prometheus first, but we want to make sure other fixes and
features are 100% first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,15 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
-
-- Add support to remote write to Cortex
-- Added recording rules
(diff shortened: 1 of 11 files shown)


JosephSalisbury (approved):

JosephSalisbury:
as an alternative, we could add a flag to enable / disable cortex support, then we don't need to stress about whether it's on or not?

mig4:
> as an alternative, we could add a flag to enable / disable cortex support, then we don't need to stress about whether it's on or not?

Let's see https://github.com/giantswarm/prometheus-meta-operator/pull/199

TheoBrigitte:
This was fixed in #199.

PR #199, opened 2020-11-05
Make writing to Cortex optional and disable it

mig4:
Writing to cortex is not ready to be released yet, it has to be disabled
in g8s-prometheus first, but we want to make sure other fixes and
features are 100% first.

### DIFF
--- a/helm/prometheus-meta-operator/templates/configmap.yaml
+++ b/helm/prometheus-meta-operator/templates/configmap.yaml
@@ -28,6 +28,8 @@ data:
         {{- range .Values.Installation.V1.Monitoring.Prometheus.Bastions }}
         - {{ . }}
         {{- end }}
+        # remoteWrite:
+        #   url: https://prometheus-us-central1.grafana.net/api/prom/push
         retention:
           duration: 2w
           size: 90GB
--- a/helm/prometheus-meta-operator/templates/configmap.yaml
+++ b/helm/prometheus-meta-operator/templates/configmap.yaml
@@ -28,8 +28,6 @@ data:
         {{- range .Values.Installation.V1.Monitoring.Prometheus.Bastions }}
         - {{ . }}
         {{- end }}
-        remoteWrite:
-          url: https://prometheus-us-central1.grafana.net/api/prom/push
         retention:
           duration: 2w
           size: 90GB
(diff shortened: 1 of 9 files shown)


helm/prometheus-meta-operator/templates/configmap.yaml
```diff
@@ -28,8 +28,6 @@ data:
         {{- range .Values.Installation.V1.Monitoring.Prometheus.Bastions }}
         - {{ . }}
         {{- end }}
-        remoteWrite:
-          url: https://prometheus-us-central1.grafana.net/api/prom/push
```
TheoBrigitte, line 32: Instead of removing this can we please add a toggle flag, something like
```diff
        remoteWrite:	
          url: https://prometheus-us-central1.grafana.net/api/prom/push
+         enable: false
```

So when we need to enable this, its just a matter of toggling this flag and not having to find what was this url.

TheoBrigitte (changes requested):
Good, but just make it a toggle flag please.

helm/prometheus-meta-operator/templates/configmap.yaml
mig4: How about if we comment out the URL instead of removing it?

Cause I do get what you mean, it would be much simpler to be able to toggle it, but adding another setting will translate to adding another CLI flag, and from UX perspective I'm not a fan of patterns like:

```
--remote-write-enable --remote-write-url http://foo
```

cause if I pass `--remote-write-url` it implies I want it enabled.

So I'd slightly prefer if we were toggling it by commenting out, but if you think a flag is better, I'll add it.

TheoBrigitte: ok, fine go with commenting it. I just didn't wanted to loose the value in the end.

mig4:
Tested on gauss:

``` sh
mig in ↕️5 ☸ giantswarm-gauss (monitoring) in st-gs-opsbox with 🈯CP:gauss took 21s 
at 11:26:21 ❯ kubectl --all-namespaces=true get prometheuses -o yaml | grep remoteWrite
🦗
```

helm/prometheus-meta-operator/templates/configmap.yaml
mig4: Done

TheoBrigitte (approved):

PR #202, opened 2020-11-12
set heartbeat client log level to fatal

TheoBrigitte:
heartbeat client logs error messages. Let's remove that because we do take care of those already and this is polluting our logs.
We cannot pass it a logger because it uses a totally different logger (logrus vs micrologger).

```
D 11/12 11:28:59 /apis/cluster.x-k8s.io/v1alpha2/namespaces/default/clusters/c0vce heartbeat checking if heartbeat exists | prometheus-meta-operator/service/controller/resource/heartbeat/create.go:16 | controller=prometheus-meta-operator-cluster-api-controller | event=update | loop=27 | version=53241904
ERRO[2020-11-12T11:29:00.137839877Z] Error occurred with Status code: 404, Message: Heartbeat not found, Took: 0.008000, RequestId: 5e3a44f6-cf20-4005-b3b5-17fe8897b972
D 11/12 11:29:00 /apis/cluster.x-k8s.io/v1alpha2/namespaces/default/clusters/c0vce heartbeat heartbeat does not exist | prometheus-meta-operator/service/controller/resource/heartbeat/create.go:20 | controller=prometheus-meta-operator-cluster-api-controller | event=update | loop=27 | version=53241904
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Set heartbeat client log level to fatal to avoid polluting our logs.
+
 ## [1.9.0] - 2020-11-11
 
 ### Added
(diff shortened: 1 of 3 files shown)


mig4 (commented):
So this gets rid of the message about 404 error, right?

TheoBrigitte:
> So this gets rid of the message about 404 error, right?

Yes, this sneaky line in the middle which is not from us.

mig4 (approved):

PR #206, opened 2020-11-16
fix an issue with bracket rendering

TheoBrigitte:


mig4 (approved):
Cool, I'd just add a comment somewhere (in source or commit desc) to say we changed the delimiters to make it easier to template alerts which already use curly brackets at runtime.

PR #210, opened 2020-11-18
Remove duplicated nginx-ingress-controller target

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/14389

- convert service monitor resource to delete only (removing nginx-ingress-controller service monitor and other)
- move apiserver to scrape config
- move cluster autoscale to scrape config

Applying this fix, does reduce memory slightly.
image

After cleaning data on disk memory dropped significantly
image

Eventually disk data will be cleaned up in maximum 1 week. In case we this high memory we can still clean pvc manually (already did for anteater).


files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -1,4 +1,57 @@
...
+  - source_labels: [__meta_kubernetes_endpoint_port_name]
+    regex: https
+    action: keep
+  - source_labels: [__meta_kubernetes_service_label_giantswarm_io_monitoring]
+    regex: "true"
+    action: drop
+[[ include "_labelingschema" . | indent 2 ]]
+[[ include "_node_role" . | indent 2 ]]
+[[ if ne .ClusterType "control_plane" ]]
+[[ if or (eq .Provider "aws") (eq .Provider "azure") ]]
```
TheoBrigitte, line 22: Replicating previous condition: https://github.com/giantswarm/prometheus-meta-operator/pull/210/files#diff-a9915600bde12cb537e4b1367447e33a4810a28bf1ac74f13fea35b6c84d1d95L65

mig4 (approved):
LGTMIIWFY

PR #218, opened 2020-11-24
set resource limits

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/14389

### DIFF
--- a/service/controller/resource/prometheus/resource.go
+++ b/service/controller/resource/prometheus/resource.go
@@ -163,14 +163,14 @@ func toPrometheus(v interface{}, config Config) (metav1.Object, error) {
 			Replicas: &replicas,
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
-					// cpu: 100m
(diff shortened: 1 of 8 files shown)


service/controller/resource/prometheus/resource.go
```diff
@@ -168,6 +168,12 @@ func toPrometheus(v interface{}, config Config) (metav1.Object, error) {
 					// memory: 5Gi
 					corev1.ResourceMemory: *resource.NewQuantity(5000*1024*1024, resource.BinarySI),
 				},
+				Limits: corev1.ResourceList{
+					// cpu: 100m
+					corev1.ResourceCPU: *resource.NewMilliQuantity(100, resource.DecimalSI),
```
MarcelMue, line 173: This should be much higher IMO

MarcelMue: Like 1 CPU at least

TheoBrigitte: right, I just checked some number to be sure
image

MarcelMue (approved):

omegas27 (approved):

PR #232, opened 2020-11-26
Add verticalpodautoscaler resource

mig4:
Enables `VerticalPodAutoscaler` to manage requests and limits on
the Prometheus `StatefulSet` which helps with scheduling and prevents
Prometheus from consuming all available resources on a node it's
running on.

Closes: https://github.com/giantswarm/giantswarm/issues/14558


mig4:
Tested on _godsmack_ (after installing the [vertical-pod-autoscaler-app](https://github.com/giantswarm/vertical-pod-autoscaler-app) via Helm).

PMO logs:

testing-1 pmo log autoscaler-created 2020-11-27--13-37-59

VPAs:

testing-2 verticalpodautoscalers 2020-11-27--13-50-13

VPA for `godsmack-prometheus` namespace:

testing-3 godsmack-prometheus verticalpodautoscaler 2020-11-27--13-52-07

Prometheus Pod requests and limits in `godsmack-prometheus` namespace:

testing-4 godsmack-prometheus pod req-limits 2020-11-27--13-54-52

helm/prometheus-meta-operator/templates/rbac.yaml
```diff
@@ -149,6 +149,18 @@ rules:
...
+    resources:
+      - verticalpodautoscalers
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
JosephSalisbury, line 163: why do we need `watch`?

service/controller/resource/prometheusautoscaler/resource.go
```diff
@@ -0,0 +1,106 @@
+package prometheusautoscaler
```
omegas27, line 1: This is unrelated to this PR but should we not move all prometheus related resources under prometheus/ so we could have resources/prometheus/prometheus and resources/prometheus/ingress resources/prometheus/autoscaling ?

service/controller/resource/prometheusautoscaler/resource.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	// updateModeAuto := vpa_types.UpdateModeAuto
```
JosephSalisbury, line 72: Do we need these comments? I don't get what they're for

service/controller/resource/prometheusautoscaler/resource.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	return metav1.ObjectMeta{
+		Name:      "prometheus",
+		Namespace: key.Namespace(cluster),
+		Labels:    key.Labels(cluster),
+	}, nil
+}
+
+func getObject(v interface{}) (metav1.Object, error) {
```
omegas27, line 61: Can we rename it to toVerticalPodAutoscaler as we usually do?

service/controller/resource/prometheusautoscaler/resource.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func getObjectMeta(v interface{}) (metav1.ObjectMeta, error) {
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return metav1.ObjectMeta{}, microerror.Mask(err)
+	}
+
+	return metav1.ObjectMeta{
+		Name:      "prometheus",
```
omegas27, line 55: Should we not name it key.ClusterID(cluster) instead?

service/controller/resource/prometheusautoscaler/resource.go
omegas27, line 72: I think you can uncomment those

go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	k8s.io/api v0.19.3
 	k8s.io/apiextensions-apiserver v0.19.3
 	k8s.io/apimachinery v0.19.3
+	k8s.io/autoscaler/vertical-pod-autoscaler v0.9.0
```
TheoBrigitte, line 46: We are currently running 0.8.0. Probably should align this.

service/controller/resource/prometheusautoscaler/resource.go
TheoBrigitte: I don't see how this is more related to prometheus than other resources?

Also I was hopping to see something named `verticalpodautoscaler`.

service/controller/resource/prometheusautoscaler/resource.go
TheoBrigitte: It feels easier to have `getObject` everywhere.

service/controller/resource/prometheusautoscaler/resource.go
```diff
@@ -0,0 +1,106 @@
...
+	vpa := &vpa_types.VerticalPodAutoscaler{
+		ObjectMeta: objectMeta,
+		Spec: vpa_types.VerticalPodAutoscalerSpec{
+			TargetRef: &autoscaling.CrossVersionObjectReference{
+				Kind:       "StatefulSet",
+				Name:       key.PrometheusSTSName(cluster),
+				APIVersion: "apps/v1",
+			},
+			// UpdatePolicy: &vpa_types.PodUpdatePolicy{
+			// 	UpdateMode: &updateModeAuto,
```
TheoBrigitte, line 84: Default mode is `Auto` AFAIK. But let's set it explicitly.

service/controller/resource/prometheusautoscaler/resource.go
```diff
@@ -0,0 +1,106 @@
...
+			// },
+			ResourcePolicy: &vpa_types.PodResourcePolicy{
+				ContainerPolicies: []vpa_types.ContainerResourcePolicy{
+					{
+						ContainerName: key.PrometheusContainerName(),
+						// Mode:             &containerScalingModeAuto,
+						// ControlledValues: &containerControlledValuesRequestsAndLimits,
```
TheoBrigitte, line 91: Please un-comment this line.

TheoBrigitte (changes requested):
I like this a lot. Just few changes here and there.

service/controller/resource/prometheusautoscaler/resource.go
omegas27: As this is not prometheus specific, sure :)

service/controller/resource/prometheusautoscaler/resource.go
omegas27: Then let's align on this

service/controller/resource/prometheusautoscaler/resource.go
mig4: Sure, initially I thought of it as related to Prometheus, but yeah I see, it just creates a VPA resource, it just happens to be for Prometheus. I'll rename it.

helm/prometheus-meta-operator/templates/rbac.yaml
mig4: Well spotted, I just copied from something else and selected too much. Will fix.

service/controller/resource/prometheusautoscaler/resource.go
mig4: No, after I added this code I noticed these are all default values in VPA if not specified, so I commented them out to test if it works without them, and it does. But I've seen Theo mentioned he prefers if we are explicit here, so I'll uncomment the code and leave it in.

service/controller/resource/prometheusautoscaler/resource.go
mig4: Yeah, the `to*` functions in some other resources are from before we split and renamed the struct members to `GetObjectMeta` and `GetDesiredObject`.

service/controller/resource/prometheusautoscaler/resource.go
mig4: If you feel strongly about it I will, but it actually bugs me why we name objects after the cluster ID. They're already in a namespace named after the cluster ID so they won't clash, and with the variable name it means I can't re-use commands from history after switching namespaces.

E.g.:

``` sh
# I set current namespace to foobz-prometheus and get the prometheus resource
kubectl ns foobz-prometheus
kubectl get prometheus/foobz

# then I switch to another prometheus namespace and want to get the prometheus resource there
# but I can't simply select the get command from history because the resource has a different name
kubectl ns barbz-prometheus 
kubectl get prometheus/barbz
```

go.mod
mig4: So, question is do we not want to upgrade to 0.9.0? Reason I'm asking is 0.9.0 added possibility of disabling scaling for individual containers, and I wanted to see if this is something we need with Prometheus.

I'm thinking, Prometheus has 3 containers, and we don't necessarily want to have evictions/re-deployments because of any of the `*-reloader` containers, so we should probably exclude them. I was just going to see if this happens in practice.

service/controller/resource/prometheusautoscaler/resource.go
mig4: Yeah, all these values are defaults, I commented them out to test, but I see the point of being explicit so I'll do that.

service/controller/resource/prometheusautoscaler/resource.go
mig4: Done.

service/controller/resource/prometheusautoscaler/resource.go
mig4: Done.

mig4: Done.

service/controller/resource/prometheusautoscaler/resource.go
mig4: Done.

service/controller/resource/prometheusautoscaler/resource.go
mig4: Done.

helm/prometheus-meta-operator/templates/rbac.yaml
mig4: Done.

go.mod
mig4: OK I just checked that and it does indeed apply recommendations to all containers currently:

testing-5 godsmack-prometheus per-container-req-limits 2020-11-30--12-05-02

WDYT, should we exclude the reloader containers?

service/controller/resource/prometheusautoscaler/resource.go
omegas27: TBH, I would prefer if we are consistent. I think calling it prometheus, or prometheus-vpa or prometheus-cr_type would make sense. We could have prometheus-ingress for example. 
The issue here is that if we change that now, I would like if we add an issue in the board to rename resources and maybe document the pattern we want to be consistent

go.mod
omegas27: Is is possible to enable recommended mode for the reloader and auto for hte prometheus pod? Otherwise, I would agree we should exclude them as they do not really fluctuate

service/controller/resource/prometheusautoscaler/resource.go
mig4: I agree about consistency, let's see if others agree about naming and if so I'll add an issue about renaming other things to make them consistent.

I disagree about `prometheus-cr_type` pattern, you already have to specify type when referencing a resource so you have `vpa/prometheus-vpa` or `VerticalPodAutoscaler/prometheus-vpa` which breaks the [golang no stuttering rule](https://blog.golang.org/package-names).

go.mod
mig4: No, per-container you can only set it to auto(on)/off.

omegas27: I think off is recommended mode but I would ignore those reloaders

omegas27 (approved):
Looks good to me but can we wait for @TheoBrigitte to take another look as well before merging? 

Also, can you update the changelog?

TheoBrigitte (approved):
Looks right, good job.

service/controller/resource/verticalpodautoscaler/resource.go
```diff
@@ -0,0 +1,106 @@
...
+			},
+			UpdatePolicy: &vpa_types.PodUpdatePolicy{
+				UpdateMode: &updateModeAuto,
+			},
+			ResourcePolicy: &vpa_types.PodResourcePolicy{
+				ContainerPolicies: []vpa_types.ContainerResourcePolicy{
+					{
+						ContainerName:    key.PrometheusContainerName(),
+						Mode:             &containerScalingModeAuto,
+						ControlledValues: &containerControlledValuesRequestsAndLimits,
```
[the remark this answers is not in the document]
mig4: Yeah, that won't work because those are consts and in golang you can't take an address of a const.

go.mod
mig4: Done.

service/controller/resource/verticalpodautoscaler/resource.go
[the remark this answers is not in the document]
TheoBrigitte: ok fine

PR #233, opened 2020-11-30
Add SLO classes to SLO alerts

QuentinBisson:
Adding slo classes that were only added to g8s-prometheus

### DIFF
--- a/files/templates/rules/recording/cortex.yml
+++ b/files/templates/rules/recording/cortex.yml
@@ -45,11 +45,11 @@ spec:
     - record: aggregation:giantswarm:cluster_certificate_not_after_seconds
       expr: "min(cert_exporter_not_after) by (cluster_id, cluster_type)"
     - record: aggregation:giantswarm:cluster_release_version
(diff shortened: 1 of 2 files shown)


JosephSalisbury:
I don't get this. What's it do?

omegas27:
This is just updating the service level rules to those in g8s-prometheus as classes were added but in g8s-prometheus and not in pmo.
Also, I removed useless label_replace as those were used for exported_cluster_id which is not used anymore

JosephSalisbury:
my concern is we didn't finish those slo classes at all, so i'm not sure if they're ready to yet to port, i don't think we even tested that PR up in g8s-prom

omegas27:
Well, we might have an issue as this is merged and released

JosephSalisbury:
I was talking about https://github.com/giantswarm/g8s-prometheus/pull/1149

omegas27:
The only ones I updated are from https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus/prometheus-recording-rules/service-level.rules.yml which is merged in master The PR you linked is mostly removing the cluster_type which I am not. We can leave this one open for now as well though if you prefer

PR #234, opened 2020-11-30
Add Inhibition rules

QuentinBisson:
Signed-off-by: Quentin Bisson <quentin@giantswarm.io>


TheoBrigitte (approved):
I trust you here, a diff with g8s-prometheus would have helped the review.

PR #245, opened 2020-12-07
Allow optional metrics path and port

QuentinBisson:


omegas27:
Tested on ghost:
image

This works but this is picking up other components. I think we could add a property to split workload and managed-apps. @giantswarm/team-atlas wdyt?

TheoBrigitte (approved):
Nice, looks good.

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -530,21 +530,26 @@
   - source_labels: [__meta_kubernetes_service_labelpresent_operated_prometheus]
     regex: (true)
     action: drop
-  - source_labels: [__meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring_port]
-    regex: (true)
-    action: keep
-  - source_labels: [__meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring_path]
-    regex: (true)
-    action: keep
```
TheoBrigitte, line 538: We should probably keep those in under not to catch other unwanted targets.

omegas27:
@TheoBrigitte what should I do about the dupes? define a split managed-apps/workload sections? Have only one for now and we will improve later? 

Also, does this means the targets are scraped twice? Or does prometheus scrape them only once if they are in multiple target groups?

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
omegas27: Well, removing those is the main objective of this PR as we want to use defaults if those annotations are not present

TheoBrigitte:
> @TheoBrigitte what should I do about the dupes? define a split managed-apps/workload sections? Have only one for now and we will improve later?

Let's keep the managed-apps and try to keep it as it was, so without those targets.

> Also, does this means the targets are scraped twice? Or does prometheus scrape them only once if they are in multiple target groups?

Targets will be scraped twice and this is bad memory wise.

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
omegas27: If we have both __meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring_path and __meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring_port this means we are not using defaults :D

TheoBrigitte: Ok, but I believe those labels with `keep` action are the ones filtering out the unwanted targets.

omegas27: Yes they are but this means we have to define those annotations even when the values behind it are defaults (metrics and main port).


Maybe we can discuss it a little during planning as this is not a priority but I was thinking we could add an annotation or label to managed apps to know if the app is managed and split the targets in 2 groups (managed-apps and workloads) but we can keep this PR open for when we have time to work on this

omegas27: The monitoring of apps document is then invalid as this explains that the default values are optional when we actually require them.

mig4:
Just for my understanding, what are those unwanted targets, and why/how do they have a `giantswarm.io/monitoring: true` label if they're unwanted?

omegas27:
The unwanted target are those we already have in other sections of the scrape config like net-exporter or chart and app operator and the lastest operator versions

mig4:
> The unwanted target are those we already have in other sections of the scrape config like net-exporter or chart and app operator and the lastest operator versions

Would it make sense to remove them from the other sections? I mean the idea behind using this label was that we'd be able to catch multiple different services with it so we don't have to duplicate configuration.

omegas27:
It would but we need to ensure all versions of the apps are scraped here or we can just ignore the targets that have the monitoring annotations everywhere. In that case, we need to find how to differenciate managed apps from the other workload we run

QuentinBisson:
@giantswarm/team-atlas I tested it on ginger and we have no dups anymore. Also, we collect apps we forgot (AWS collector for starters).

Can we go ahead here?

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -527,36 +548,30 @@
...
+    action: replace
+    target_label: __metrics_path__
+    regex: (.+)
+    # if __meta_kubernetes_service_annotation_giantswarm_io_monitoring_port, we use it as the metrics port
+  - source_labels: [__address__, __meta_kubernetes_service_annotation_giantswarm_io_monitoring_port]
     action: replace
-  - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_pod_name, __meta_kubernetes_service_annotation_giantswarm_io_monitoring_port, __meta_kubernetes_service_annotation_giantswarm_io_monitoring_path]
-    regex: (.*);(.*);(.*);(.*)
+    target_label: __address__
+    regex: ([^:]+)(?::\d+)?;(\d+)
```
TheoBrigitte, line 560: This should work too, why bother if we do not need the second port ? 
`([^:]+).*;(\d+)`

But I'm not strong on this.

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -527,36 +548,30 @@
...
+    replacement: $1:$2
+  - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_pod_name, __address__, __metrics_path__]
+    regex: (.*);(.*);(?:[^:]+:)(\d+);(.*)
```
TheoBrigitte, line 563: Same remark, I believe we could have something simpler like `(.*);(.*);.+:(\d+);(.*)`

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -527,36 +548,30 @@
...
-  - regex: (.*)
-    target_label: __address__
-    replacement: [[ .APIServerURL ]]:443
+    # if __meta_kubernetes_service_annotation_giantswarm_io_monitoring_path is present, we use it as the metrics path
+  - source_labels: [__meta_kubernetes_service_annotation_giantswarm_io_monitoring_path]
+    action: replace
+    target_label: __metrics_path__
+    regex: (.+)
+    # if __meta_kubernetes_service_annotation_giantswarm_io_monitoring_port, we use it as the metrics port
+  - source_labels: [__address__, __meta_kubernetes_service_annotation_giantswarm_io_monitoring_port]
```
TheoBrigitte, line 557: How does this behave when `__meta_kubernetes_service_annotation_giantswarm_io_monitoring_port` is empty?

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -527,36 +548,30 @@
...
-  - source_labels: [__meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring_path]
-    regex: (true)
-    action: keep
-  - regex: (.*)
-    target_label: __address__
-    replacement: [[ .APIServerURL ]]:443
+    # if __meta_kubernetes_service_annotation_giantswarm_io_monitoring_path is present, we use it as the metrics path
+  - source_labels: [__meta_kubernetes_service_annotation_giantswarm_io_monitoring_path]
+    action: replace
+    target_label: __metrics_path__
```
TheoBrigitte, line 554: How does this behave when `__meta_kubernetes_service_annotation_giantswarm_io_monitoring_path` is empty ?

TheoBrigitte (changes requested):
I like the new approach a lot.

Just thought me might simplify those regex, but maybe I'm wrong and that's not needed. Do as you like.

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
QuentinBisson: If the regex does not match, there will be no replacement: https://github.com/prometheus/prometheus/blob/916dbd4c8ac5103ed38a2137b6fd09c83350bda0/pkg/relabel/relabel.go#L211

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
QuentinBisson: Same as above: If the regex does not match, there will be no replacement: prometheus/prometheus@916dbd4/pkg/relabel/relabel.go#L211

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
QuentinBisson: So it will take the default metrics_path value

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
QuentinBisson: I took this from how prometheus operator should be scraping: https://gist.github.com/reachlin/a98b90afcbff4604c90c183a0169474f#file-prometheus-yml-L33  
Basically, the following means:
([^:]+) = address
(?::\d+)? = port if it exists. This specific regex just ensures we do not capture the port value so we can do $1:$2, otherwise, we would need to do $1:$3

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
QuentinBisson: We could, this regex just drops the unused address using ?:. This is merely a convenient way to use regexes :D

QuentinBisson: You can check https://stackoverflow.com/a/3513858 for details on non capturing groups

QuentinBisson:
@TheoBrigitte would it make sense to document the non capturing group as a comment?

TheoBrigitte:
> @TheoBrigitte would it make sense to document the non capturing group as a comment?

I know what a non capturing group is, but I felt its not needed.

QuentinBisson:
I did not mean you did not know what it is, especially as I learned of its existence 3 months ago :D I just asked if we should comment what the regex does :D 
I will try without the capturing group

TheoBrigitte:
> I did not mean you did not know what it is, especially as I learned of its existence 3 months ago :D I just asked if we should comment what the regex does :D
> I will try without the capturing group

I do not see a need to comment this regex, but if you do please comment it.

QuentinBisson:
I am not sure what is worrying you about the __meta_kubernetes_service_annotation_giantswarm_io_* ? I am merely reproducing the prometheus.io/scrape=true default behavior here

QuentinBisson:
I changed the regex to match your preferences

PR #246, opened 2020-12-07
Update architect to 1.0.0

omegas27:


TheoBrigitte (approved):
LGTM

About the failing build, I've seen this error recently locally, but have no fix for it yet.

omegas27:
I will ask in sig-releng, otherwise, we will have to remove the prometheus/prometheus dependency by copying the code

omegas27:
The prometheus maintainers view on this pains me

TheoBrigitte:
> I will ask in sig-releng, otherwise, we will have to remove the prometheus/prometheus dependency by copying the code

I'm not a fan of copying upstream code over.

omegas27:
Me neither, but the prometheus maintainer keep saying this is internal code that we should not really use.

omegas27:
They do not want to provide any kind of semver compability

TheoBrigitte:
> Me neither, but the prometheus maintainer keep saying this is internal code that we should not really use.

I hate when they answer this.

PR #260, opened 2021-01-05
prometheus self monitoring

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/14998

Prometheus monitor itself via `localhost:9090`, decided not to discover prometheus via kubernetes api because this means sorting out rbac access for every TC instance to CP k8s api.

Tested on gauss

CP
image

TC
image


QuentinBisson (approved):
LGTM

PR #269, opened 2021-01-13
remove alert resource

TheoBrigitte:
Towards: giantswarm/giantswarm#14998

We should have done that some long time ago. Removing now because I edited the wrong heartbeat due to this.


QuentinBisson:
We need to move the recording rules to the helm template then. If you remove the resource, the recording rules will never be created

TheoBrigitte:
> We need to move the recording rules to the helm template then. If you remove the resource, the recording rules will never be created

That's not true, currently this resource only delete stuff: https://github.com/giantswarm/prometheus-meta-operator/pull/269/files#diff-f61d7ded3080a566307de1e690027aaf39642c75a76d6dcd89706b42a018a8adL10

I suggest we fix the recording rules in a separate PR.

QuentinBisson:
Oh yes, you are right about this

QuentinBisson (approved):

PR #277, opened 2021-01-15
Fix wrong provider in tests

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/14850


stone-z (approved):
It would be ideal to have correct provider labels to simulate reality, but I guess this will just break when some feature relies on the provider.

PR #279, opened 2021-01-20
fix(#15277): configure prometheus container image tag

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/15277


JosephSalisbury (approved):
lgtmiiwfy

PR #280, opened 2021-01-21
rename control plane & tenant cluster

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/15028

- renamed some alerts
- move controller/control-plane > controller/managementcluster
- removed control plane or tenant cluster notation in some places

### DIFF
--- a/helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
+++ b/helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
@@ -3,7 +3,7 @@ kind: PrometheusRule
 metadata:
   labels:
     {{- include "labels.common" . | nindent 4 }}
-    cluster_type: "control_plane"
+    cluster_type: "management_clustemanagement_cluster"
   name: service-level
   namespace: '{{ include "resource.default.namespace" . }}'
 spec:
--- a/helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
+++ b/helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
@@ -3,7 +3,7 @@ kind: PrometheusRule
 metadata:
   labels:
     {{- include "labels.common" . | nindent 4 }}
-    cluster_type: "management_clustemanagement_cluster"
+    cluster_type: "management_cluster"
   name: service-level
   namespace: '{{ include "resource.default.namespace" . }}'
 spec:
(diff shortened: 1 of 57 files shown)


helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -3,7 +3,7 @@ kind: PrometheusRule
 metadata:
   labels:
     {{- include "labels.common" . | nindent 4 }}
-    cluster_type: "control_plane"
+    cluster_type: "management_clustemanagement_cluster"
```
QuentinBisson, line 6:
```suggestion
    cluster_type: "management_cluster"
```

service/controller/managementcluster/controller.go
```diff
@@ -73,7 +73,7 @@ func NewController(config ControllerConfig) (*Controller, error) {
 		c := controller.Config{
 			K8sClient: config.K8sClient,
 			Logger:    config.Logger,
-			Name:      project.Name() + "-control-plane-controller",
+			Name:      project.Name() + "-management-cluster-controller",
```
QuentinBisson, line 76: We will need to be careful here as we will need to remove the finalizers

QuentinBisson:
Did you update the tests as well?

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
TheoBrigitte: nice typo, thanks.

service/controller/managementcluster/controller.go
TheoBrigitte: oh gosh, almost forgot about those. I guess we can keep a tiny `control-plane-controller` which only delete finalizers.

QuentinBisson: Is there a way in operatorkit to have a controller that only removes finalizers without adding them?

TheoBrigitte: I don't know

TheoBrigitte:
> Did you update the tests as well?

done

QuentinBisson (approved):

service/controller/managementcluster/controller.go
TheoBrigitte: added legacy finalizer resource, which takes care of removing the old finalizers.

PR #290, opened 2021-01-28
Deploy recording rules on all prometheuses

QuentinBisson:


QuentinBisson:
Going with a ping as this is blocking cortex

PR #296, opened 2021-02-01
Remove unecessary label_replace

QuentinBisson:
Remove uncessary label_replace that were needed when we were having exported_cluster_id


mig4 (approved):
I don't quite get those `label_replace`s, are they replacing `cluster_id` label with itself?

QuentinBisson:
They were used to replace exported_cluster_id by cluster id in g8s prometheus but the transformation tool removed those exported_ prefix so relabelling cluster_id to cluster_id makes no sense

PR #299, opened 2021-02-01
fix-some-rules-per-provider

QuentinBisson:
This PR fixes some rules to only be present when they should (control_plane and provider split)


QuentinBisson:
Tested on ginger

TheoBrigitte (approved):
LGTM

PR #309, opened 2021-02-02
Fix missing app label on some target

QuentinBisson:
Fixing the aggregation:kubernetes:up metrics in cortex


JosephSalisbury (approved):
lgtmiiwfy

PR #324, opened 2021-02-08
Update dependencies

mig4:
We currently have some conflicting versions in our dependency graph and
are lagging behind with some major versions of our libraries. This
doesn't cause any runtime issues yet, but it would be good to resolve to
make future updates easier and to unblock dependabot.

Closes https://github.com/giantswarm/giantswarm/issues/15588

## Checklist

I have:

- [x] Described why this change is being introduced
- [x] Separated out refactoring/reformatting in a dedicated PR
- [ ] Updated changelog in `CHANGELOG.md`

PR #328, opened 2021-02-09
Account for replace directives in `go.mod`

mig4:
Replace directives in modules other than main (so e.g. in dependencies)
are ignored by Go toolchain itself, but some of them are important for
building the project (see
https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive
for details).

In our case we want to control the version of `k8s.io/client-go` used
and we want to use a temporary fork of `sigs.k8s.io/cluster-api` to
to match `github.com/giantswarm/apiextensions/v3`,
`github.com/giantswarm/k8sclient/v5` and
`github.com/giantswarm/operatorkit/v4` which use it for compatibility
with K8s 1.18.

Towards: https://github.com/giantswarm/giantswarm/issues/15588

I have:


mig4:
Note: like the others, this targets the branch in #324, it's split out for easier review. This one also makes the builds and tests pass.

New graph after aligning replace directives:

prometheus-meta-operator gomod-graph 5-after-replace-alignment

QuentinBisson (approved):

TheoBrigitte (approved):
Looks about right

PR #330, opened 2021-02-10
fix(15620) invalid prometheus memory usage recording rules

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/15620

The fixed values are visible in the issue

## Checklist

I have:


mig4:
So just for a bit of context, why are the rules incorrect? We used to have `cluster_id` label on `container_memory_working_set_bytes` metric and now we don't?

QuentinBisson:
Sure, so `container_memory_working_set_bytes` comes from cadvisor on the management cluster. As this is scraped by the management cluster prometheus, the cluster_id label will be the management cluster id, not the workload cluster id when we want to have the memory usage of prometheus per cluster and not per management cluster

mig4:
> Sure, so `container_memory_working_set_bytes` comes from cadvisor on the management cluster. As this is scraped by the management cluster prometheus, the cluster_id label will be the management cluster id, not the workload cluster id when we want to have the memory usage of prometheus per cluster and not per management cluster

Got ya. Could you add this, what you just said, as a comment in the file to explain the `label_replace`?

Reading the code, this seemed a bit fishy, but with the explanation it makes sense, or at least I don't know of a cleaner way to do this.

QuentinBisson:
You are right as I was thinking lately that every rule we have should have a comment and a test going with it :D

mig4 (approved):
Cool, thanks for adding comments. LGTM.
Might be good to see if @TheoBrigitte has any ideas here as well.

PR #332, opened 2021-02-10
refactor: remove service monitor resource

QuentinBisson:
Remove the service monitor resource as all the servicemonitor cr have been removed.

## Checklist

I have:


mig4:
This makes me cry 😭

QuentinBisson:
Why is that?

mig4:
> Why is that?

I abhor templating YAML to generate data structures, it's error prone, it's hard to read, to avoid duplication you have to use text-based includes which are not obvious where they come from and have no concept of the overall config structure. Using native objects like `ServiceMonitor` would've been so much better.

Well, fortunately if we want PMO to be useful upstream we'll need to separate all this out and address this anyway, so I'm still hopeful 😉

TheoBrigitte (approved):
Let it go in peace, but might bring it back some time in the future @mig4

QuentinBisson:
@mig4 I know this was not directed at me and thanks for explaining :D
I think we should contribute https://github.com/prometheus-operator/prometheus-operator/issues/2787 to actually achieve that :D

PR #338, opened 2021-02-11
Fix AzureClusterCreationFailed alert

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/15656

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


whites11 (approved):
thx

TheoBrigitte (approved):

fiunchinho:
Mm... how is this working? Months ago we switched the namespace of the metrics to be `azure-operator` https://github.com/giantswarm/azure-collector/blob/master/service/collector/set.go#L13.

Actually, if I go to the prometheus instance in `talos` I only see metrics prefixed with `azure-operator`, no `azure-collector`

image

QuentinBisson:
The metric used for the alert is statusresource_cluster_status so I imagine this is not coming from the azure-collector. Knowing the source of the metric could be useful in the future though

PR #355, opened 2021-02-16
inhibit-updating-10min-after

calvix:
towards https://github.com/giantswarm/giantswarm/issues/15764

To avoid pages that happens right after the update of a cluster.
 I am adjusting the inhibition alert to inhibit the alert 10 mins after the update finished.

We often see an update finished but the error counters are still high or error budged be still depleted right after the update finished and this causes pages that resolve in a few minutes. 

function `changes()` will be equal to 1 in case there was a change of the value over a period of the time (in our case 10 minutes)

So when the update finishes the value of the metric will change from 1 to 0 but the inhibition will still work for 10 minutes. This will have no effect on the start of the inhibition since we don't drop the old rules.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


whites11 (approved):
seems legit

QuentinBisson (approved):

TheoBrigitte (approved):
Nice improvement

PR #368, opened 2021-02-19
firecracker-improve-pod-stuck-aws

calvix:
towards https://github.com/giantswarm/giantswarm/issues/15617

add back alerts about pod stuck to relevant components for firecracker

## Checklist

I have:


QuentinBisson (changes requested):
Do you mind adding exceptions https://github.com/giantswarm/prometheus-meta-operator/blob/ec63a0f409d0f2b34d1a45729e44411103fee387/helm/prometheus-meta-operator/templates/prometheus-rules/management-cluster.rules.yml#L14 for the biscuit alerts then?

calvix:
> Do you mind adding exceptions
> 
> https://github.com/giantswarm/prometheus-meta-operator/blob/ec63a0f409d0f2b34d1a45729e44411103fee387/helm/prometheus-meta-operator/templates/prometheus-rules/management-cluster.rules.yml#L14
> for the biscuit alerts then?

that is a little bit hard as  cluster-operator and vertical-pod-autoscaler can run on different providers and if I set the exception for them then on other providers nobody will get the alert

QuentinBisson:
Should it not be split per provider in that case at least for cluster operator? Also, why are you taking ownership of the vertical pod autoscaler?

QuentinBisson:
Technically, you can just add the alert for celestial in azure.management-cluster.rules.yaml for azure-operator and cluster operator and add those as exception in the biscuti rule as rocket is taking care of kvm managment clusters

QuentinBisson (approved):
Thanks :D

calvix:
@giantswarm/team-celestial you ok with this? Just adding a alert for you that will page incase azure operator or cluster operator will be stuck in pending on azure MC

whites11 (approved):
LGTMIIWFY

PR #382, opened 2021-02-23
move-cluster-auto-scaler-app-install-failure-to-batman

calvix:
These alerts are usually related to app-operator/chart-operator process and we do not feel responsible for a potential fix in case these components do not work properly so transferring the alert to batman as they manage the process.

discussion happened on slack: https://gigantic.slack.com/archives/CRZN9GLJW/p1614068260030900

The alert is now anyway in `severity: notify` so it does not page, I will leave it to you if you wanna make it paging on not.

## Checklist

I have:

towards https://github.com/giantswarm/giantswarm/issues/15613


rossf7 (approved):
Fine to move these back to batman. 👍 

I think we can reduce the number of rules to not need cluster-autoscaler specific rules. But will do that once app-exporter changes are done.

PR #383, opened 2021-02-23
Fix syntax errors in `management-cluster.rules`

mig4:
The expressions of `ManagementClusterPodPending*` alerts used syntax
`pod~!"..."` which is incorrect and causes _prometheus-operator_
admission controller to reject the rules and fail the deployment.
The correct syntax is `pod!~"..."`. This changes the expressions to use
that.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed syntax error in expressions of `ManagementClusterPodPending*` alerts
+
 ## [1.24.0] - 2021-02-23
 
 ### Added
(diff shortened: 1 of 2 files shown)


calvix:
good find, is there a way how to catch this in a CI?

QuentinBisson:
Right now, not really. We do have a story to do this, an rfc even but that's it for now

QuentinBisson:
The validation is done by the prometheus operator but we need to add static validation

mig4:
So just to expand a bit on what Quentin said, we used to have validation via `promtool` in CI but that doesn't work anymore because the rules are wrapped in `PrometheusRule` CRs and IIRC there were some other minor tweaks. _prometheus-operator_ currently doesn't provide an offline tool that can be used to check `PrometheusRule` CRs but it does provide an admission controller that does that online. So we need to look into extracting the necessary bits from the admission controller to be able to run that offline in CI, or to run the rules through an end-to-end test that would pass them through the admission controller.

PR #386, opened 2021-02-24
add catalog label to app-exporter metrics

piontec:
This PR adds one more label ('catalog') to the metrics exported to cortex from `app-exporter`.


QuentinBisson (approved):
LGTM when tested

PR #397, opened 2021-02-26
Fix VPA craziness

QuentinBisson:
Fix VPA craziness by maintaining the status after a pmo restart

Closes https://github.com/giantswarm/giantswarm/issues/15930
Closes https://github.com/giantswarm/giantswarm/issues/15924
Closes https://github.com/giantswarm/giantswarm/issues/15927

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-Fix VPA crazyness
+- Fix equality check on the VPA CR to prevent it being overriden and losing it's status information on every prometheus-meta-operator deployment.
 
 ## [1.24.4] - 2021-02-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+Fix VPA crazyness
+
 ## [1.24.4] - 2021-02-26
 
 ### Changed
--- a/service/controller/resource/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/verticalpodautoscaler/resource.go
@@ -2,13 +2,14 @@ package verticalpodautoscaler
 
 import (
 	"context"
-	"reflect"
 
 	"github.com/giantswarm/k8sclient/v5/pkg/k8sclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	autoscaling "k8s.io/api/autoscaling/v1"
 	v1 "k8s.io/api/core/v1"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
@@ -182,5 +183,5 @@ func hasChanged(current, desired metav1.Object) bool {
 	c := current.(*vpa_types.VerticalPodAutoscaler)
 	d := desired.(*vpa_types.VerticalPodAutoscaler)
 
-	return !reflect.DeepEqual(c.Spec, d.Spec)
+	return !apiequality.Semantic.DeepEqual(c.Spec, d.Spec)
 }
--- a/service/controller/resource/verticalpodautoscaler/create.go
+++ b/service/controller/resource/verticalpodautoscaler/create.go
@@ -14,24 +14,30 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", "creating")
+	r.logger.Debugf(ctx, "checking if vpa cr already exists")
 	current, err := r.vpaClient.AutoscalingV1().VerticalPodAutoscalers(desired.GetNamespace()).Get(ctx, desired.GetName(), metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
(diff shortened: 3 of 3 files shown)


QuentinBisson:
So the DeepEqual was the source of the VPA issue 🤦🏻 

```json
k logs -n monitoring  prometheus-meta-operator-unique-7b9dc8844f-x7tdz -f | grep verticalpodautoscaler
{"caller":"github.com/giantswarm/prometheus-meta-operator/service/controller/resource/verticalpodautoscaler/create.go:17","controller":"prometheus-meta-operator-cluster-api-controller","event":"update","level":"debug","loop":"1","message":"checking if vpa cr already exists","object":"/apis/cluster.x-k8s.io/v1alpha2/namespaces/default/clusters/5n9k5","resource":"verticalpodautoscaler","time":"2021-03-01T10:00:06.630462+00:00","version":"109451505"}
{"caller":"github.com/giantswarm/prometheus-meta-operator/service/controller/resource/verticalpodautoscaler/create.go:17","controller":"prometheus-meta-operator-management-cluster-controller","event":"update","level":"debug","loop":"0","message":"checking if vpa cr already exists","object":"/api/v1/namespaces/default/services/kubernetes","resource":"verticalpodautoscaler","time":"2021-03-01T10:00:06.631554+00:00","version":"94817374"}
{"caller":"github.com/giantswarm/prometheus-meta-operator/service/controller/resource/verticalpodautoscaler/create.go:31","controller":"prometheus-meta-operator-cluster-api-controller","event":"update","level":"debug","loop":"1","message":"checking if vpa cr needs to be updated","object":"/apis/cluster.x-k8s.io/v1alpha2/namespaces/default/clusters/5n9k5","resource":"verticalpodautoscaler","time":"2021-03-01T10:00:06.634667+00:00","version":"109451505"}
{"caller":"github.com/giantswarm/prometheus-meta-operator/service/controller/resource/verticalpodautoscaler/create.go:31","controller":"prometheus-meta-operator-management-cluster-controller","event":"update","level":"debug","loop":"0","message":"checking if vpa cr needs to be updated","object":"/api/v1/namespaces/default/services/kubernetes","resource":"verticalpodautoscaler","time":"2021-03-01T10:00:06.635173+00:00","version":"94817374"}
```

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+Fix VPA crazyness
```
JosephSalisbury, line 13: Can we be a little more specific here?

service/controller/resource/verticalpodautoscaler/resource.go
```diff
@@ -182,5 +183,5 @@ func hasChanged(current, desired metav1.Object) bool {
 	c := current.(*vpa_types.VerticalPodAutoscaler)
 	d := desired.(*vpa_types.VerticalPodAutoscaler)
 
-	return !reflect.DeepEqual(c.Spec, d.Spec)
+	return !apiequality.Semantic.DeepEqual(c.Spec, d.Spec)
```
JosephSalisbury, line 186: Are there other places we need this trick? Do we need to spread it as a best practice?

QuentinBisson: Theoretically, we should everytime we need to check equality between 2 api objects

QuentinBisson: https://github.com/kubernetes/apimachinery/blob/17b79e78ed4ac42a54a132a541fea663fbbd80cd/pkg/api/equality/semantic.go#L27
We should advertise it as best practice but the kickoff got in the way

CHANGELOG.md
QuentinBisson: I hope my new wording is better :D

mig4:
> So the DeepEqual was the source of the VPA issue 🤦🏻

Hi @QuentinBisson, sorry, could you explain that a bit? The logs only seem to show that it's checking if VPA needs to be updated and nothing after that.

The `hasChanged` function only compares the `.Spec` structure, so whatever is in `.Status` shouldn't affect it.

In the changelog you said this happens on deployment of new version of PMO? How? Is PMO deleting resources it manages when Helm upgrades it? (It seems to be that would have more serious consequences cause it would also delete `Prometheus` resources)

BTW, I'm all for this if it's the best practice, I'd just like to understand how does it help with this issue.

QuentinBisson:
Those a good questions :)
If you look at the create.go file, you will see I added some logs so if the vpa resource was either created or updated, the log would have shown it.

The issue with the DeepEqual was not in the different statuses but in the resource policy of the VPA CR. Technically, resource structs equality should not be tested using réflection, but using the Cmp méthod which the Semantic.DeepEqual is using.
For some reason, the resource struct has some internal field populated by one of the kubernetes controllers that is not in our struct so the DeepEqual fails.

The previous pmo code would actually Always update the VPA cr when it is released because hasChanged is always true. As the cr is updated on every pmo release, the status is reset

service/controller/resource/verticalpodautoscaler/resource.go
QuentinBisson: See https://github.com/giantswarm/fmt/pull/134/files

mig4:
> Those a good questions :)
> If you look at the create.go file, you will see I added some logs so if the vpa resource was either created or updated, the log would have shown it.

Right yeah, but was the log you quoted in the message supposed to show those? I triple-checked now and I can't see that 😅 

> The issue with the DeepEqual was not in the different statuses but in the resource policy of the VPA CR. Technically, resource structs equality should not be tested using réflection, but using the Cmp méthod which the Semantic.DeepEqual is using.
> For some reason, the resource struct has some internal field populated by one of the kubernetes controllers that is not in our struct so the DeepEqual fails.
> See [kubernetes/apimachinery@`35be0db`/pkg/api/equality/semantic.go#L30](https://github.com/kubernetes/apimachinery/blob/35be0db31bd6e5635ef6366708dc2b137392f6a2/pkg/api/equality/semantic.go#L30)

Right, I see, or maybe even not that the `.Spec` is changed by a controller but that some nested slice or map has a default that is semantically equal but causes straight comparison to fail. Got it. Thanks for explaining.

> The previous pmo code would actually Always update the VPA cr when it is released because hasChanged is always true. As the cr is updated on every pmo release, the status is reset

OK, cool, let's go with that.

mig4 (approved):

PR #411, opened 2021-03-09
Update architect-orb to 2.3.0

ubergesundheit:
Update architect-orb to 2.1.0 which will use giantswarm.github.io as catalog base url. The old behaviour was to use giantswarm.github.com which will be deprecated in april by GitHub. Towards giantswarm/giantswarm#15898

I did some changes in this PR:

- Update the vendored prometheus/common code to [`6e540be`](https://github.com/prometheus/common/tree/6e540be) because [tests failed](https://app.circleci.com/pipelines/github/giantswarm/prometheus-meta-operator/2098/workflows/cac34602-bf37-480f-b142-ebbaf7b4bde1/jobs/8494/parallel-runs/0/steps/0-118)
- Remove the go-test ci job. The steps included are already run in go-build
- Add a config file for golangci to ignore vendored prometheus/common directory


pkg/prometheus/common/config/config.go
```diff
@@ -24,7 +24,7 @@ type Secret string
 // MarshalYAML implements the yaml.Marshaler interface for Secrets.
 func (s Secret) MarshalYAML() (interface{}, error) {
 	if s != "" {
-		return string(s), nil
+		return "<secret>", nil
```
QuentinBisson, line 27:
```suggestion
		return string(s), nil
```

pkg/prometheus/common/config/http_config_test.go
```diff
@@ -824,6 +968,29 @@ func TestTLSRoundTripperRaces(t *testing.T) {
 	}
 }
 
+func TestHideHTTPClientConfigSecrets(t *testing.T) {
```
QuentinBisson, line 971: Can you remove TestHideHTTPClientConfigSecrets for now?

ubergesundheit: Yes, can you explain why? It works, why bother removing it?

ubergesundheit: Forget above comment :D

QuentinBisson (approved):

PR #412, opened 2021-03-10
Support both CAPI certs and legacy certs

QuentinBisson:
## Checklist

I have:


anvddriesch (approved):
Not too deep into PMO but looks good to me.

PR #428, opened 2021-03-19
Support Legacy and CAPI clusters

QuentinBisson:
## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add support for monitoring vmware clusters
+- Add support to get the API Server URL for both legacy and CAPI clusters
 
 ### Changed 
 
@@ -26,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Avoid alerting for `MatchingNumberOfPrometheusAndCluster` when a cluster is
   being deleted.
+- Fix scrapeconfig to use secured ports for kubernetes control plane components for CAPI clusters
+- Fix scrapeconfig to proxy all calls through the API Server for CAPI clusters
 
 ## [1.24.7] - 2021-03-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fix
 
 - Fix typo in `MatchingNumberOfPrometheusAndCluster` alert
+- Fix scrapeconfig to use secured ports for kubernetes control plane components for CAPI clusters
+- Fix scrapeconfig to proxy all calls through the API Server for CAPI clusters
 
 ## [1.24.8] - 2021-03-18
 
@@ -27,8 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Avoid alerting for `MatchingNumberOfPrometheusAndCluster` when a cluster is
   being deleted.
-- Fix scrapeconfig to use secured ports for kubernetes control plane components for CAPI clusters
-- Fix scrapeconfig to proxy all calls through the API Server for CAPI clusters
+
 
 ## [1.24.7] - 2021-03-18
 
--- a/service/controller/resource/certificates/resource.go
+++ b/service/controller/resource/certificates/resource.go
@@ -139,6 +139,7 @@ func (r *Resource) getSource(ctx context.Context, v interface{}) (*corev1.Secret
 
 		secret, err = r.k8sClient.K8sClient().CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
 		if apierrors.IsNotFound(err) {
+			// fallthrough
 			secret = nil
 		} else if err != nil {
 			return nil, microerror.Mask(err)
--- a/service/controller/resource/certificates/resource.go
+++ b/service/controller/resource/certificates/resource.go
@@ -12,6 +12,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/giantswarm/prometheus-meta-operator/service/key"
 )
@@ -84,6 +85,11 @@ func (r *Resource) getObjectMeta(v interface{}) (metav1.ObjectMeta, error) {
 }
 
 func (r *Resource) getDesiredObject(v interface{}) (*corev1.Secret, error) {
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
 	objectMeta, err := r.getObjectMeta(v)
 	if err != nil {
 		return nil, microerror.Mask(err)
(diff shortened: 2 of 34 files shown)


service/key/key.go
```diff
@@ -221,3 +228,15 @@ func RemoteWriteUsernameKey() string {
 func RemoteWritePasswordKey() string {
 	return "password"
 }
+
+func IsCAPICluster(obj metav1.Object) bool {
```
JosephSalisbury, line 232: What about AWS?

QuentinBisson: AWS is using v1alpha2 :)

JosephSalisbury: Okay, I read this backwards. Cool!

JosephSalisbury (approved):

CHANGELOG.md
```diff
@@ -8,16 +8,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Fix scrapeconfig to proxy all calls through the API Server for CAPI clusters
+
 ## [1.24.8] - 2021-03-18
 
 ### Fix
 
 - Avoid alerting for `MatchingNumberOfPrometheusAndCluster` when a cluster is
   being deleted.
+- Fix scrapeconfig to use secured ports for kubernetes control plane components for CAPI clusters
+- Fix scrapeconfig to proxy all calls through the API Server for CAPI clusters
```
mig4, line 31: Looks like rebase/merge error, like these lines are in two places **Unreleased** and **1.24.8**.

service/controller/resource/certificates/resource.go
```diff
@@ -117,10 +139,15 @@ func (r *Resource) getSource(ctx context.Context, v interface{}) (*corev1.Secret
...
 		secret, err = r.k8sClient.K8sClient().CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
 		if apierrors.IsNotFound(err) {
-			// fall through
+			secret = nil
 		} else if err != nil {
 			return nil, microerror.Mask(err)
 		}
+
+		if secret != nil {
+			// We return the first secret we find
```
mig4, line 148: What's the idea here?

service/controller/resource/resource.go
```diff
@@ -84,7 +84,7 @@ func New(config Config) ([]resource.Interface, error) {
 				},
 				{
 					NameFunc:      key.CAPICertificateName,
-					NamespaceFunc: key.ClusterID,
+					NamespaceFunc: key.CAPICertificateNamespace,
```
mig4, line 87: Does this work? I mean cause for GS clusters the namespace is still the cluster ID, but `CAPICertificateNamespace` returns the namespace where the `Cluster` object resides, which is usually `default`, right? Or am I missing something?

service/controller/resource/scrapeconfigs/resource.go
```diff
@@ -98,12 +100,16 @@ func toSecret(v interface{}, config Config) (*corev1.Secret, error) {
 	var workloadClusterETCDDomain string = ""
 	if "workload_cluster" == key.ClusterType(v) {
 		clusterID := key.ClusterID(cluster)
+		// Try to get the etcd url from the Giant Swarm way
 		service, err := config.K8sClient.K8sClient().CoreV1().Services(clusterID).Get(context.Background(), "master", metav1.GetOptions{})
-		if err != nil {
+		if apierrors.IsNotFound(err) {
+			// fallthrough, we need to get it from capi but ignore it for now
```
mig4, line 106: So, if we ignore it here, where is it being populated?

service/controller/resource/prometheus/test/case-5-cluster-api-v1alpha3.golden
```diff
@@ -23,7 +23,7 @@ spec:
             values:
             - master
   apiserverConfig:
-    host: https://master.baz
+    host: https://:0
```
mig4, line 26: Is this right?

service/controller/resource/certificates/resource.go
QuentinBisson: We have multiple sources for secrets now, the one in the default namespace, or the one for CAPI, once we find one, we return to avoid looping for no reason

CHANGELOG.md
QuentinBisson: Let me fix it

service/controller/resource/resource.go
QuentinBisson: For CAPI clusters, the secret (cluster-id-kubeconfig) is now in the namespace of the cluster cr, not the defautl namespace.
I tested it on all providers first :D

service/controller/resource/scrapeconfigs/resource.go
QuentinBisson: Empty, as we do not have the etcd url for now.

QuentinBisson: We will need to fix it at a later time but by default the etcd's metric endpoint is not used

service/controller/resource/prometheus/test/case-5-cluster-api-v1alpha3.golden
QuentinBisson: Thanks I fixed the configs :D

service/key/key.go
TheoBrigitte, line 232: I do not understand this function, a comment would be welcome to explain what we are doing here.

From what I read I get that we only consider a cluster being CAPI when it's on v1alpha3 and it is not labeled with a giantswarm version. Is this correct?

Could we maybe only check the suffix part `giantswarm.io/version` ? So we are provider agnostic here.

service/key/key.go
```diff
@@ -153,9 +158,11 @@ func AlertManagerKey() string {
 func APIUrl(obj interface{}) string {
 	switch v := obj.(type) {
 	case *v1.Service:
-		return v.Spec.ClusterIP
+		return fmt.Sprintf("%s:443", v.Spec.ClusterIP)
+	case *capiv1alpha3.Cluster: // Support CAPI Clusters
+		return fmt.Sprintf("%s:%d", v.Spec.ControlPlaneEndpoint.Host, v.Spec.ControlPlaneEndpoint.Port)
 	case metav1.Object:
-		return fmt.Sprintf("master.%s", v.GetName())
+		return fmt.Sprintf("master.%s:443", v.GetName())
```
TheoBrigitte, line 165: Nice :+1:

service/controller/resource/scrapeconfigs/resource.go
TheoBrigitte: If this need to be fixed later maybe label this comment with a `TODO`.

service/controller/resource/certificates/resource.go
```diff
@@ -117,10 +139,15 @@ func (r *Resource) getSource(ctx context.Context, v interface{}) (*corev1.Secret
 
 		secret, err = r.k8sClient.K8sClient().CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
 		if apierrors.IsNotFound(err) {
-			// fall through
+			secret = nil
```
TheoBrigitte, line 142: Why do we need this? I would expect `secret` to be `nil` when there is an error.

TheoBrigitte (changes requested):
Haven't review everything.

service/key/key.go
QuentinBisson: I will comment but we only check for Azure labels because only Azure is using v1alpha3 clusters. They set this annotation on the ones the Azure Operator is tracking

service/controller/resource/certificates/resource.go
QuentinBisson: I just do not really like empty methods :D

TheoBrigitte (approved):
Fine with me when tested.
I'm not super happy with both giantswarm and CAPI logic being coupled but I do not have a better suggestion for now, so let's roll.

service/controller/resource/resource.go
mig4: OK, cool, thanks!

mig4 (approved):

service/controller/resource/certificates/resource.go
QuentinBisson: Interestingly, this returns an empty secret by default:

https://github.com/kubernetes/client-go/blob/fa8f4cc307b81b125f6beb2cea6f093ed9e3c069/kubernetes/typed/core/v1/secret.go#L72

QuentinBisson:
Thanks guys :D 
Let's improve on this next week :)

PR #465, opened 2021-03-26
Adds support for calculating maximum CPU

JosephSalisbury:
Similarly to how we specify a maximum amount of memory (to fit within the nodes of a cluster), this PR adds support to specify a maximum amount of CPU.

e.g:
```
$ kubectl get nodes ip-10-0-5-109.eu-central-1.compute.internal -o yaml | grep -A 2 allocatable
        f:allocatable:
          .: {}
          f:attachable-volumes-aws-ebs: {}
--
  allocatable:
    attachable-volumes-aws-ebs: "25"
    cpu: "4"
```
```
$ kubectl -n=gorilla-prometheus get vpa prometheus -o yaml | grep -A 2 maxAllowed
      maxAllowed:
        cpu: "3"
        memory: 14325966Ki
```

## Checklist

I have:

### DIFF
--- a/service/controller/resource/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/verticalpodautoscaler/resource.go
@@ -85,6 +85,10 @@ func (r *Resource) getObject(ctx context.Context, v interface{}) (*vpa_types.Ver
 	minCpu := key.PrometheusDefaultCPU()
 	minMemory := key.PrometheusDefaultMemory()
 
+	maxCpu, err := r.getMaxCPU(ctx)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
 	maxMemory, err := r.getMaxMemory(ctx)
 	if err != nil {
 		return nil, microerror.Mask(err)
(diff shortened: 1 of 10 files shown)


service/controller/resource/verticalpodautoscaler/resource.go
```diff
@@ -135,6 +140,39 @@ func (r *Resource) getObject(ctx context.Context, v interface{}) (*vpa_types.Ver
...
+		}
+
+		for _, n := range nodes.Items[1:] {
+			s, ok := n.Status.Allocatable[v1.ResourceCPU]
+			if ok && nodeCpu.Cmp(s) == -1 {
+				nodeCpu = &s
+			}
+		}
+	}
+	if nodeCpu == nil || nodeCpu.IsZero() {
```
JosephSalisbury, line 164: If `len(nodes.Items) == 0`, then `nodeCpu` could be nil, which would panic is we just called `IsZero` on. Fixed below as well.

JosephSalisbury:
Bam tests fixed

service/controller/resource/verticalpodautoscaler/resource.go
QuentinBisson: Nice

QuentinBisson (approved):
LGTM if tested

PR #478, opened 2021-04-01
Remove promxy resource

QuentinBisson:
Remove promxy resource not needed after  https://github.com/giantswarm/promxy-app/pull/33
## Checklist

I have:


TheoBrigitte:
Is it safe to just delete the resource? Shouldn't we convert it to a delete resource first ?

QuentinBisson:
We could but deploying the promxy app in the newer version should override the configmap anyway

QuentinBisson:
My idea for next week is to actually release pmo without the promxy resource then promxy

TheoBrigitte:
Fine we me then when the config gets overwritten then, I just wanted to make sure we are not leaving anything behind to be cleaned up.

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Split `ManagementClusterCertificateWillExpireInLessThanTwoWeeks` alert per provider.
 
+### Removed 
+
+Remove promxy resource.
```
JosephSalisbury, line 27:
```suggestion
- Remove promxy resource.
```

JosephSalisbury (approved):

PR #485, opened 2021-04-06
Page when prometheus cannot send samples to cortex

QuentinBisson:
This alert will be used to detect communication issues between cortex and prometheus

## Checklist

I have:

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
@@ -26,7 +26,7 @@ spec:
     - alert: PrometheusFailsToCommunicateWithRemoteStorageAPI
       annotations:
         description: '{{`Prometheus can''t communicate with Remote Storage API at {{ $labels.url }}.`}}'
-      expr: rate(prometheus_remote_storage_samples_failed_total{app="prometheus"}[1h]) > 0.10
+      expr: rate(prometheus_remote_storage_samples_failed_total{app="prometheus"}[1h]) > 0.10 or rate(prometheus_remote_storage_samples_total[1h]) == 0
       for: 1h
       labels:
         area: empowerment
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
@@ -22,3 +22,16 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    ## Pages Atlas when prometheus fails to send samples to cortex
+    - alert: PrometheusFailsToCommunicateWithRemoteStorageAPI
+      annotations:
+        description: '{{`Prometheus can''t communicate with Remote Storage API.`}}'
+      expr: rate(prometheus_remote_storage_samples_failed_total{app="prometheus"}[15m]) > 0.25
+      for: 15m
+      labels:
+        area: empowerment
+        severity: page
+        team: atlas
+        topic: observability
+
+        
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
@@ -25,7 +25,7 @@ spec:
     ## Pages Atlas when prometheus fails to send samples to cortex
     - alert: PrometheusFailsToCommunicateWithRemoteStorageAPI
       annotations:
-        description: '{{`Prometheus can''t communicate with Remote Storage API.`}}'
+        description: '{{`Prometheus can''t communicate with Remote Storage API at {{ $labels.url }}.`}}'
       expr: rate(prometheus_remote_storage_samples_failed_total{app="prometheus"}[15m]) > 0.25
       for: 15m
       labels:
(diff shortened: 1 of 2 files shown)


helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
```diff
@@ -22,3 +22,16 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    ## Pages Atlas when prometheus fails to send samples to cortex
+    - alert: PrometheusFailsToCommunicateWithRemoteStorageAPI
+      annotations:
+        description: '{{`Prometheus can''t communicate with Remote Storage API.`}}'
+      expr: rate(prometheus_remote_storage_samples_failed_total{app="prometheus"}[15m]) > 0.25
+      for: 15m
```
mig4, line 30: This is purely for my own understanding, if we have `rate(...[15m])` that means the alert will go to _PENDING_ when error rate was >= 25% over the last 15 minutes, and with `for: 15m` it means it will go to _FIRING_ if this keeps up for another 15 minutes. Is that right? (Just checking as I'm not sure how this works exactly)

mig4 (commented):
Looks good, just a couple of questions/suggestions.

helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
QuentinBisson: Technically, rate(...[15m]) has nothing to do with error rate. It will average the metric over 15 minutes so the alert will fire if for 15 minutes, the rate for 15 minutes is over 25%. 

As tthe rate over 25 minutes is an average, it will change over time.

You can see an example here:
image

QuentinBisson: It appeared during testing that we need to rate over 1 hour due to dropped samples
image


As you can see in peu01, when remote write is working, this rate is at 0:
image

mig4: Ack. I was asking about the interaction between the range `[15m]` and `for: 15m` setting on the alert, as it's one of those things that I'm not quite sure of how it works.

So when there's no samples the `rate()` is 0? Wouldn't that be OK though? As in if there's no samples that probably means Prometheus is having other issues and there's other alerts to cover that case?

QuentinBisson: Technically, prometheus should always send data to cortex as this is another process as this is always scraping. If this is 0, as you say, this is another issue. We could alert on lack of send samples total as well

QuentinBisson: @mig4 do you mind taking another look?

helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
```diff
@@ -22,3 +22,17 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    ## Pages Atlas when prometheus fails to send samples to cortex
+    - alert: PrometheusFailsToCommunicateWithRemoteStorageAPI
+      annotations:
+        description: '{{`Prometheus can''t communicate with Remote Storage API at {{ $labels.url }}.`}}'
+      expr: rate(prometheus_remote_storage_samples_failed_total{app="prometheus"}[1h]) > 0.10 or rate(prometheus_remote_storage_samples_total[1h]) == 0
```
mig4, line 29: I'm a bit concerned that this could be too noisy:

- isn't error rate of 10% quite low? I mean even if the error rate is > 50%, as long as it's succeeding eventually, that's OK, no?
- if there are no samples that means we are likely already getting paged with heartbeat or whatever as Prometheus is having issues; wouldn't this just mean now we'd be getting two alerts at the same time instead of 1?

Also, does having the range of `1h` not mean that if this fires and the cause is fixed, it'll take another hour for this to resolve?

Any chance you could test some scenarios like that?

QuentinBisson: From what I saw, the usual error rate is below 20% so 10% was a nice number. Sadly, under 10% can mean it will never succeed as cortex only supports 24h old samples and prometheus currently does not support dropping samples. There is a design doc about this but this might not make it before prometheus 3.0.0.

There could be a lack of samples for other reasons, like prometheus taking it's time to read the wal and so on. It could be that we will get paged twice but this is still unclear so I preferred to page twice now and remove this if this is too noisy. 

Remember this will only page during working hours :D

It does not completely mean that,because the rate drops with every passing minute. From what I saw, samples are retried every hour or so so the rate of 1h is needed.

Honestly, I'm not sure I can test such a complex setup without setting a proxy to cut the connection to cortex for more than 24h and I honestly think this does not deserve this effort

JosephSalisbury, line 29: Why do we have the `app="prometheus"` selector on one metric, but no the other?

QuentinBisson: I think we can just remove it

JosephSalisbury (approved):

PR #488, opened 2021-04-06
add alert for HA master down for too long

anvddriesch:
Towards: https://github.com/giantswarm/giantswarm/issues/15833

While HA Masters makes one master going down not such a big deal, we still want to get notified if this is the case for extended periods of time. (Is 1h appropriate here? We can also do 30 minutes instead.)

## Checklist

I have:

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/aws.workload-cluster.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/aws.workload-cluster.rules.yml
@@ -92,6 +92,22 @@ spec:
         severity: page
         team: firecracker
         topic: kubernetes
+    - alert: WorkloadClusterHAMasterDownForTooLong
+      annotations:
+        description: '{{`Master node in HA setup is down for a long time.`}}'
+        opsrecipe: master-node-missing/
+      expr: kubernetes_build_info{app="kubelet"} and on(cluster_id) sum(kubernetes_build_info{app="kubelet", role="master"}) by (cluster_id) == 2
+      for: 1h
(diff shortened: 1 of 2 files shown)


helm/prometheus-meta-operator/templates/prometheus-rules/aws.workload-cluster.rules.yml
```diff
@@ -92,6 +92,22 @@ spec:
         severity: page
         team: firecracker
         topic: kubernetes
+    - alert: WorkloadClusterHAMasterDownForTooLong
+      annotations:
+        description: '{{`Master node in HA setup is down for a long time.`}}'
+        opsrecipe: master-node-missing/
+      expr: kubernetes_build_info{app="kubelet"} and on(cluster_id) sum(kubernetes_build_info{app="kubelet", role="master"}) by (cluster_id) == 2
+      for: 1h
```
njuettner, line 100: I would lower it to 30m, 1h seems pretty long to me 🙂

njuettner (commented):
Nice 👍🏻, only one small nit. Maybe we can set the downtime to 30min. WDYT? We already exclude cluster upgrades so I think 30 minutes is reasonable.

helm/prometheus-meta-operator/templates/prometheus-rules/aws.workload-cluster.rules.yml
anvddriesch: Sure, let's go with 30 minutes.

njuettner (approved):

PR #490, opened 2021-04-07
Add Missing Workload App labels

QuentinBisson:
## Checklist

I have:

### DIFF
--- a/files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
+++ b/files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
@@ -558,6 +558,33 @@
     action: drop
 [[ include "_common" . | indent 2 ]]
 [[ include "_labelingschema" . | indent 2 ]]
+[[ if eq .ClusterType "workload_cluster" ]]
+  metric_relabel_configs:
+  - source_labels: [deployment]
+    regex: (.+)
+    target_label: workload_type
+    replacement: deployment
+  - source_labels: [daemonset]
+    regex: (.+)
+    target_label: workload_type
+    replacement: daemonset
+  - source_labels: [statefulset]
+    regex: (.+)
+    target_label: workload_type
+    replacement: statefulset
+  - source_labels: [deployment]
+    regex: (.+)
+    target_label: workload_name
+    replacement: ${1}
+  - source_labels: [daemonset]
+    regex: (.+)
+    target_label: workload_name
+    replacement: ${1}
+  - source_labels: [statefulset]
+    regex: (.+)
+    target_label: workload_name
+    replacement: ${1}
+[[ end ]]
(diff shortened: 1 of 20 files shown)


tomahawk28 (approved):
Wonderful, Thanks for the changes!

rossf7 (approved):
LGTM Thx! @QuentinBisson

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -558,6 +558,33 @@
     action: drop
 [[ include "_common" . | indent 2 ]]
 [[ include "_labelingschema" . | indent 2 ]]
+[[ if eq .ClusterType "workload_cluster" ]]
+  metric_relabel_configs:
```
rossf7, line 562: @tomahawk28 SInce we discussed in planning I thought I'd post what I found earlier.

cc @QuentinBisson 


I think when the managed apps alerts were introduced the workload_type and workload_name labels were being set by PCC. So when we stopped using PCC they started appearing as empty

https://github.com/giantswarm/prometheus-config-controller/blob/9d8e47c0e244c4fa75a4043395bfa1840a6d1922/service/controller/v1/prometheus/prometheus.go#L152-L157

Dumb question but why are we using `metric_relabel_configs` instead of `relabel_configs` here? Could that be why the labels are not appearing?

QuentinBisson: The only difference is that this adds the labels on the metrics while relabel configs adds it to the targets. Can you try to create a PR to test it?

tomahawk28: The document said that it had no difference between `metric_relabel_configs` and `relabel_configs`. The only difference is that the former ignores the `up` metrics when the latter applies. 

I checked the metrics again but it doesn't look like `metric_relabel_configs` are working at the moment. (I can still see the deployment metrics with `deployment` label)
https://prometheus.g8s.alpaca.ore.aws.k8s.3stripes.net/pna01/graph?g0.expr=%7Bstatefulset!%3D%22%22%7D&g0.tab=1&g0.stacked=0&g0.range_input=1h

@QuentinBisson Can you maybe check the syntax again?

tomahawk28: And one more thing; Why I can't see `pna01-prometheus/kube-state-metrics-pna01/0` job in prometheus's target list? https://prometheus.g8s.alpaca.ore.aws.k8s.3stripes.net/pna01/targets

QuentinBisson: kube-state-metrics is located under pna01-prometheus/workload-pna01/0

QuentinBisson: so that's why :D

PR #492, opened 2021-04-08
Fix(16537) Prometheus Storage alert

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/16537
Closes https://github.com/giantswarm/giantswarm/issues/15871

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add workload type and name labels for `ManagedAppBasicError*` alerts
 - Add alert for master node in HA setup down for too long.
 
+### Fixed
+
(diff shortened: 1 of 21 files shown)


QuentinBisson:
Tested on gauss:
image

JosephSalisbury:
Could you write a little about _why_ we need this? I don't follow it

QuentinBisson:
We had only one alert related to storage space too low on the management cluster that was paging biscuit for the prometheus containers.
We splitted the alert using prometheus in the mountpoint but it was not sufficient as the alert was paging both biscuit and atlas.

This PR will fix this situation by alerting atlas for prometheus volumes and biscuit for the rest

JosephSalisbury (approved):

JosephSalisbury:
gotcha, i'd be missed the difference in priority class matching

PR #494, opened 2021-04-09
Add aggregation for Docker actions

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/16599

I want to see how many pulls we're doing (as part of reducing overall registry usage), so adding the aggregation. New metric should add time series on the the order of O(nodes), which is fine.

## Checklist

I have:


QuentinBisson (approved):
LGTM

helm/prometheus-meta-operator/templates/recording-rules/cortex.rules.yml
```diff
@@ -25,6 +25,8 @@ spec:
     # Spot Instances being used
     - expr: count(sum(aws_operator_ec2_instance_status{lifecycle != ""}) by (ec2_instance, lifecycle)) by  (lifecycle)
       record: aggregation:aws:instance_lifecycle
+    - expr: sum(engine_daemon_image_actions_seconds_count) by (cluster_id, cluster_type, action)
```
kopiczko, line 28: Is it possible/does it make sense to narrow it down to pulls only?

PR #505, opened 2021-04-13
Remove absent from KSM alert

MarcelMue:
Towards https://github.com/giantswarm/giantswarm/issues/16178

This was just different from the other `Down` metrics. In all others we do not check for `absent`. Could it therefor be that the absent part leads to false pages?

My idea was to just merge this and try it.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
+- Removed check for `absent` metric in `KubeStateMetricsDown` alert.
+
 ## [1.30.0] - 2021-04-12
 
 ### Removed
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml
@@ -44,7 +44,7 @@ spec:
       annotations:
         description: '{{`KubeStateMetrics ({{ $labels.instance }}) is down.`}}'
         opsrecipe: kube-state-metrics-down/
-      expr: up{app="kube-state-metrics"} == 0 or absent(up{app="kube-state-metrics"}) == 1
+      expr: up{app="kube-state-metrics"} == 0
       for: 15m
       labels:
         area: kaas


ericgraf (approved):

mig4:
So IIRC we added `absent()` because if the target is discovered dynamically (like based on labels like `giantswarm.io/monitoring`), you won't get an `up == 0` when it's down, it just won't be there.

I.e. Prometheus doesn't explicitly know that there should be a `kube-state-metrics` target, it finds everything with a `giantswarm.io/monitoring` label (or similar) and usually `kube-state-metrics` happens to be part of this set.

That would mean removing the `absent()` would prevent us from getting the alert at all.

But, this seems to have changed somehow and I'm a little lost in how the targets are configured now via the `additionalScrapeConfigs` mechanism, @QuentinBisson could you double-check my thinking here? Is that still the case, or should Prometheus now expect KSM to be there and `up == 0` is enough?

MarcelMue:
@giantswarm/team-atlas What is your opinion here?
Removing `absent` does cause the alert to fire if KSM is not installed. But we are not doing the same for any other `down` alert. Should we change this in general?

QuentinBisson:
I think @mig4 is correct here: https://serverfault.com/a/938251

Currently, if ksm is , 1.2.0 it will be a specific target in the scrape config https://github.com/giantswarm/prometheus-meta-operator/blob/138ea90ae04c7bebcfdb64e511386ec9343d103b/files/templates/scrapeconfigs/additional-scrape-configs.template.yaml#L530 but for versions after 1.2.0, it will be discovered automatically through labels. 

Are the false negative for versions after 1.2.0? 

Maybe we could try to use absent_over_time instead of absent then?

QuentinBisson:
The solution proposed upstream is actually to use https://github.com/helm/charts/blob/7e45e678e39b88590fe877f159516f85f3fd3f38/stable/prometheus-operator/templates/prometheus/rules/kubernetes-absent.yaml#L46

Maybe we just need to change our target down alerts to `absent(up{app="app_name"} == 1)` instead?

mig4:
> The solution proposed upstream is actually to use [helm/charts@`7e45e67`/stable/prometheus-operator/templates/prometheus/rules/kubernetes-absent.yaml#L46](https://github.com/helm/charts/blob/7e45e678e39b88590fe877f159516f85f3fd3f38/stable/prometheus-operator/templates/prometheus/rules/kubernetes-absent.yaml#L46)
> 
> Maybe we just need to change our target down alerts to `absent(up{app="app_name"} == 1)` instead?

Yeah, this has occurred to me, I think I'd be OK with that actually, however I don't think it makes much difference, the way it is just happens to also catch the case when Prometheus knows about the target but can't reach it.

With regards to the mentioned [PM](https://github.com/giantswarm/giantswarm/issues/16178), Marcel had a hypothesis that this could've been caused by Prometheus being unstable / restarting often at the time. I think it makes sense as that would've prevented the inhibiting alert from reaching AlertManager.

So when it comes to this I'd argue we should wait and see if this happens again before making any changes.

MarcelMue:
My plan right now:

Add a new alert which only checks if the metric is missing. 
Work on this PR again in the follow up.

The main problem with pages seem to be the missing metrics in a combination with inhibitions not working. Putting this in a separate alert will make tracking of the issue easier.

MarcelMue:
Updated this PR after #517 was merged.

MarcelMue:
PTAL

stone-z (approved):
Different problems --> different alerts makes sense to me

PR #517, opened 2021-04-19
Add kube-state-metrics missing alert

MarcelMue:
## Checklist

This is a preparation for https://github.com/giantswarm/prometheus-meta-operator/pull/505

We are currently getting paged by `kube-state-metrics` down quite a bit for different reasons. This will make it easier for oncallers to quickly identify why the problem is happening.

I have:

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml
@@ -56,6 +56,22 @@ spec:
         severity: page
         team: ludacris
         topic: observability
+    - alert: KubeStateMetricsMissing
+      annotations:
+        description: '{{`KubeStateMetrics ({{ $labels.instance }}) is missing.`}}'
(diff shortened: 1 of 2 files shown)


stone-z (approved):

helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml
```diff
@@ -56,6 +56,22 @@ spec:
         severity: page
         team: ludacris
         topic: observability
+    - alert: KubeStateMetricsMissing
+      annotations:
+        description: '{{`KubeStateMetrics ({{ $labels.instance }}) is missing.`}}'
+        opsrecipe: kube-state-metrics-down/
+      expr: absent(up{app="app_name"} == 1)
```
QuentinBisson, line 63: I think the app name is not correct here :D

MarcelMue: Argh - fix incoming

PR #528, opened 2021-04-22
Clean up puma targets

QuentinBisson:
Puma's Prometheus is currently really slow. This PR should help reduce the load by removing not working targets 

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - Remove `PodLimitAlmostReachedAWS` and `EBSVolumeMountErrors` alerts as they were not used.
+- Removed Puma's blackbox exporters as the blackbox exporter is not installed there.
 
 ## [1.32.1] - 2021-04-22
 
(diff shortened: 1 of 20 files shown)


QuentinBisson:
This definitely sped up puma's prometheus. @giantswarm/team-atlas wdyt?

QuentinBisson:
No issues to report on puma since I deployed this PR

TheoBrigitte (commented):
I find it strange to have per installation config

QuentinBisson:
I am not a big fan but we have this for dragon and dinosaur already and this is only temporary till all the kvm control plane are running the same hive version :( 
Theoretically, we could have this in the installation repo but having this from outside does not seem worth it right now. WDYT?

TheoBrigitte:
> I am not a big fan but we have this for dragon and dinosaur already and this is only temporary till all the kvm control plane are running the same hive version :(
> Theoretically, we could have this in the installation repo but having this from outside does not seem worth it right now. WDYT?

Ok fine then, but let's make it a TODO and track somewhere when this should be removed, otherwise I'm afraid it will stay forever.

QuentinBisson:
Clean up issue : https://github.com/giantswarm/giantswarm/issues/16921

QuentinBisson:
I added the todo comments and linked the issue

QuentinBisson:
Seems unneeded

PR #542, opened 2021-05-06
fix InhibitionKubeStateMetricsDown not firing longe enough

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/16803

This PR fixes an issue with `InhibitionKubeStateMetricsDown` which does not fires long enough and result in `KubeStateMetricsMissing` firing too early.

* `KubeStateMetricsMissing` fires until at 11:55:35

image

* `InhibitionKubeStateMetricsDown` fires until 11:53:01, which gives ~2mn for the alert to fire.

image

* fixed `InhibitionKubeStateMetricsDown` fires until at 11:55:35, which gives no time for the alert to fire.

image

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/inhibit.all.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/inhibit.all.rules.yml
@@ -22,7 +22,7 @@ spec:
     - alert: InhibitionKubeStateMetricsDown
       annotations:
         description: '{{`KubeStateMetrics ({{ $labels.instance }}) is down.`}}'
-      expr: up{app="kube-state-metrics"} == 0 or absent(up{app="kube-state-metrics"}) == 1
+      expr: up{app="kube-state-metrics"} == 0 or absent(up{app="kube-state-metrics"} == 1)
       labels:
         area: kaas
         kube_state_metrics_down: "true"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix ManagedApp alert
+- Fix `InhibitionKubeStateMetricsDown` not firing long enough
 
 ## [1.33.0] - 2021-04-27
 


helm/prometheus-meta-operator/templates/prometheus-rules/inhibit.all.rules.yml
```diff
@@ -22,7 +22,7 @@ spec:
     - alert: InhibitionKubeStateMetricsDown
       annotations:
         description: '{{`KubeStateMetrics ({{ $labels.instance }}) is down.`}}'
-      expr: up{app="kube-state-metrics"} == 0 or absent(up{app="kube-state-metrics"}) == 1
+      expr: up{app="kube-state-metrics"} == 0 or absent(up{app="kube-state-metrics"} == 1)
```
QuentinBisson, line 25: I fail to see how this fixes it. Could you explain why that works?

TheoBrigitte: `absent(up{app="kube-state-metrics"})` check if `up{app="kube-state-metrics"}` is missing, but it might be zero and present. While `absent(up{app="kube-state-metrics"} == 1)` check for its absence while value is 1. So it fires also when value is zero, and this aligns with `KubeStateMetricsMissing` [expression](https://github.com/giantswarm/prometheus-meta-operator/blob/32841dfa3a8a8a8d391667099657f7cf4454cd04/helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml#L63)

TheoBrigitte: This inhibition is essentially a combination of the expression for the alert it needs to inhibit. 

```yaml
    - alert: KubeStateMetricsDown
      expr: up{app="kube-state-metrics"} == 0

    - alert: KubeStateMetricsMissing
      expr: absent(up{app="kube-state-metrics"} == 1)

    - alert: InhibitionKubeStateMetricsDown
      expr: up{app="kube-state-metrics"} == 0 or absent(up{app="kube-state-metrics"}) == 1
```

MarcelMue (approved):
Works for me

helm/prometheus-meta-operator/templates/prometheus-rules/inhibit.all.rules.yml
QuentinBisson: Thanks Theo

QuentinBisson (approved):

PR #545, opened 2021-05-06
inhibit KubeStateMetricsDown and KubeStateMetricsMissing

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/16803

Inhibition for both alert was wrong, this probably caused the false pages.

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml
@@ -52,7 +52,7 @@ spec:
         cancel_if_cluster_status_creating: "true"
         cancel_if_cluster_status_deleting: "true"
         cancel_if_cluster_status_updating: "true"
-        kube_state_metrics_down: "true"
+        cancel_if_any_kube_state_metrics_down: "true"
         severity: page
         team: ludacris
         topic: observability
@@ -68,7 +68,7 @@ spec:
         cancel_if_cluster_status_creating: "true"
         cancel_if_cluster_status_deleting: "true"
         cancel_if_cluster_status_updating: "true"
-        kube_state_metrics_down: "true"
+        cancel_if_any_kube_state_metrics_down: "true"
         severity: page
         team: ludacris
         topic: observability
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- inhibit KubeStateMetricsDown and KubeStateMetricsMissing
+
 ## [1.34.0] - 2021-05-06
 
 ### Changed


helm/prometheus-meta-operator/templates/prometheus-rules/up.all.rules.yml
```diff
@@ -52,7 +52,7 @@ spec:
         cancel_if_cluster_status_creating: "true"
         cancel_if_cluster_status_deleting: "true"
         cancel_if_cluster_status_updating: "true"
-        kube_state_metrics_down: "true"
+        cancel_if_any_kube_state_metrics_down: "true"
```
stone-z, line 55: Not totally sure how this is wired up with the old/new setup, but if this inhibition name comes from [here](https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus/templates/alertmanager/configmap.yaml) then I think the `any` here is incorrect

QuentinBisson: The inhibitions are coming from prometheus-meta-operator now as g8s is gone but cancel_if_any_kube_state_metrics_down does not exist there. We changed it to cancel_if_mc_kube_state_metrics_down but the value is never set  for some reason

MarcelMue: I am _so confused_

I can only find one instance of the inhibition here:
https://github.com/giantswarm/prometheus-meta-operator/blob/f115e2ddf34959bb39c842c927002dd0d578fbfc/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus-meta-operator.rules.yml#L43

What should the correct value be? I feel like I don't get it at all right now :D

QuentinBisson: Oh wait, I was wrong, we still configure inhibitions in g8s-prometheus ... 
I will see myself out now :D

QuentinBisson: Sorry for your brain @MarcelMue

QuentinBisson: @TheoBrigitte when do we set kube_state_metrics_down to true now?

TheoBrigitte: Inhibition are still wired in alertmanager AFAIK, so I added it there.

TheoBrigitte: `kube_state_metrics_down` is set by `InhibitionKubeStateMetricsDown`

QuentinBisson: Thanks,  Could you just add a comment here to explain for the future?

TheoBrigitte: Feels super weird to have a comment here about this specific label, while we have many other inhibition label all over the place. Skipping comment and moving on.

PR #548, opened 2021-05-11
Add alertmanager-dashboard alert

njuettner:
Introducing alert for `alertmanager-dashboard` deployment not running, which pages only in business hours.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add alert `alertmanager-dashboard` not satisfied.
+
 ## [1.34.1] - 2021-05-10
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


anvddriesch (approved):
All good if this is tested. I am just wondering if there isn't already and app-rule that would cover karma?

njuettner:
> All good if this is tested. I am just wondering if there isn't already and app-rule that would cover karma?

What do you mean? IIRC you have to specify the cluster id and the installation in case a specific deployment only runs in a certain place, is there another way how to do that (I'm not the a PRO in alert-manager 😄 )?

anvddriesch:
There is an alert called `WorkloadClusterAppFailed` and I thought that one might also page for Karma.

njuettner:
Hmm from checking the expression it seems it's only looking for `giantswarm` namespace so it would ignore karma because it's in `monitoring` namespace

anvddriesch:
Ah okay. In that case let's go with this :) That will also page us directly which is probably better.

PR #549, opened 2021-05-11
Add alerts for alloc orphan and time wait sockets

ericgraf:
## Checklist

I have:

This is related to https://github.com/giantswarm/giantswarm/issues/14756. 

This change introduces 3 alerts. One for number of allocated ports, one for orphaned ports, and one for time wait state ports.

This is to help early identify when the node runs out of ports and can start having issues. The limits are set at 90% and 80% in order early notify before there are errors

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
@@ -75,3 +75,37 @@ spec:
         team: rocket
         {{- end }}
         topic: network
+    - alert: HighNumberOfAllocatedSockets
+      annotations:
+        description: '{{`Over 90 percent of sockets are allocated on {{ or $labels.pod_name $labels.instance }}.`}}'
+      # This threshold value is based off net.ipv4.ip_local_port_range = 32768 60999
+      expr: node_sockstat_TCP_alloc > ( 0.9 * ( 60999 - 32768 ) )
+      for: 5m
+      labels:
+        area: kaas
+        severity: page
+        team: ludacris
+        topic: network
+    - alert: HighNumberOfOrphanedSockets
+      annotations:
+        description: '{{`Over 80 percent of ports are orphaned {{ or $labels.pod_name $labels.instance }}.`}}'
+      # This threshold value is based off net.ipv4.ip_local_port_range = 32768 60999
+      expr: node_sockstat_TCP_orphan > ( 0.8 * ( 60999 - 32768 ) )
+      for: 5m
+      labels:
+        area: kaas
+        severity: page
+        team: ludacris
+        topic: network
+    - alert: HighNumberOfTimeWaitSockets
+      annotations:
+        description: '{{`Over 80 percent of available ports are in time wait state on {{ or $labels.pod_name $labels.instance }}.`}}'
+      # This threshold value is based off net.ipv4.ip_local_port_range = 32768 60999
+      expr: node_sockstat_TCP_orphan > ( 0.8 * ( 60999 - 32768 ) )
+      for: 5m
+      labels:
+        area: kaas
+        severity: page
+        team: ludacris
+        topic: network
+
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
@@ -77,34 +77,40 @@ spec:
         topic: network
     - alert: HighNumberOfAllocatedSockets
       annotations:
-        description: '{{`Over 90 percent of sockets are allocated on {{ or $labels.pod_name $labels.instance }}.`}}'
+        description: '{{`Over 90 percent of sockets are allocated on {{ $labels.instance }}.`}}'
+        opsrecipe: high-port-usage/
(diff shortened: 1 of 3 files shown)


ericgraf:
@giantswarm/team-ludacris 
Couple of points:
- I am not crazy about hard coding a threshold number, but I'm not sure how to expose the kernel parameter as a variable here. net.ipv4.ip_local_port_range doesn't seem to have change often. I wasn't able to find anywhere where we explicitly change it.
- I set the team to only ludacris with the intent of changing it once we know how often this alert actually fires. I searched on a few customer clusters and we were far from the threshold I set. 

WDYT?

ericgraf:
I also didn't add an alert for socket memory . To know the proper value we would need to get the value of  `tcp_mem` kernel parameter. Adding an alert for socket memory might be able to detect socket memory leaks.

helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
```diff
@@ -75,3 +75,37 @@ spec:
         team: rocket
         {{- end }}
         topic: network
+    - alert: HighNumberOfAllocatedSockets
+      annotations:
+        description: '{{`Over 90 percent of sockets are allocated on {{ or $labels.pod_name $labels.instance }}.`}}'
```
stone-z, line 80: Can you explain the pod name OR instance logic? I'm guessing the pod name would be the node exporter pod, but wouldn't this metric always be at the node level?

stone-z (commented):
I'm down to try this out and see what we actually learn from these alerts. My only request would be that there should be an ops recipe describing what the heck this means and what to do (if we need to do anything) because with area oncall others will also be getting these pages if they fire.

helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
ericgraf: Fixed.
Thanks for pointing this out.

ericgraf: Fixed.
Thanks for pointing this out.

helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
```diff
@@ -75,3 +75,43 @@ spec:
...
+      for: 5m
+      labels:
+        area: kaas
+        cancel_if_outside_working_hours: "true"
+        severity: page
+        team: ludacris
+        topic: network
+    - alert: HighNumberOfTimeWaitSockets
+      annotations:
+        description: '{{`Over 80 percent of available ports are in time wait state on {{ $labels.instance }}.`}}'
```
MarcelMue, line 106: Okay here is my question: Why alert if 80% of ports are in wait but 90% allocated? What's the difference between wait and allocated?

Like will it not anyway hit 90% allocated if 80% are in wait?

stone-z: > Like will it not anyway hit 90% allocated if 80% are in wait?

I think not necessarily as it would depend on workload behavior? Every workload could `FIN` at once and hit 80% tw without having 90% allocated overall, IIUC.

But my feedback might actually be the opposite of Marcel's -- why 80% here? Might that be too high to actually catch cases where the workload is "leaking" sockets? Or maybe we just care about extreme cases? I'm fine either way as this is exploratory, just curious if there's a reason for 80.

helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
```diff
@@ -75,3 +75,43 @@ spec:
...
+        opsrecipe: high-port-usage/
+      # This threshold value is based off net.ipv4.ip_local_port_range = 32768 60999
+      expr: node_sockstat_TCP_orphan > ( 0.8 * ( 60999 - 32768 ) )
```
stone-z, line 109: This is the wrong metric, copied from above I assume.

helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
ericgraf: Ports go into a `Time Wait (tw)` state when they are closed from the server side . Before the port can be reallocated, the kernel needs to wait for any client packets that might still be in transit or make sure it receives an ACK that the connection is closed on the client side. This is to prevent the packet from being routed to a new socket that might reuse the same port. If this count gets really high it usually indicates there are things that can be done to correct it. 

80% threshold is too high, I simply picked an arbitrary conservative number that is lower than 90%. It does makes more sense to lower it, I'll change it to 50%. If 50% of the total allocatable ports are stuck waiting and not allowed to be used, there are things that can be done to improve the socket handling.

helm/prometheus-meta-operator/templates/prometheus-rules/network.all.rules.yml
ericgraf: Yep my mistake.

ericgraf:
I also lowered the total orphan port alert threshold. Might still need to be adjusted overtime.

stone-z (approved):
I'm happy to try this out now :+1:

MarcelMue (approved):

PR #553, opened 2021-05-14
Increases duration for prometheus low disk space alert

JosephSalisbury:
This alert resolves itself very often - Prometheus does disk cleanup, so let's give it more time to do so before alerting

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Increased duration for `PrometheusPersistentVolumeSpaceTooLow` alert
+
 ## [1.35.0] - 2021-05-12
 
 ### Added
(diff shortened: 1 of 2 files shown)


JosephSalisbury:
e.g:

Screenshot 2021-05-14 at 14 41 22

you can see disk cleanup occuring, 10m is way too small a duration

QuentinBisson:
6h seems like a lot of time though. If there is a storage issue, prometheus would stop working first. Should we make it notify or remote the alert altogether?

JosephSalisbury:
yeah i flip-flopped a lot about the actual time - what about `1h`?

QuentinBisson:
That sounds better to me :D

JosephSalisbury:
@QuentinBisson ptala

TheoBrigitte (approved):
Feels like prometheus is doing some cleaning every 2h, but 1h looks fine anyway.

QuentinBisson (approved):

PR #556, opened 2021-05-18
Merge AlertManager global HTTP config with generated receivers

mig4:
AlertManager doesn't do merging itself, which means that if we specify
HTTP config in the receivers (e.g. to specify auth credentials), then
the global config is not used for those receivers.

We want to preserve any options set in the global config which may
specify things like a proxy server to use, which affects the entire
environment.

Towards: https://github.com/giantswarm/giantswarm/issues/17238

## Checklist

I have:

### DIFF
--- a/service/controller/resource/heartbeatrouting/receiver/receiver.go
+++ b/service/controller/resource/heartbeatrouting/receiver/receiver.go
@@ -28,16 +28,9 @@ func toReceiver(cfg alertmanagerconfig.Config, cluster metav1.Object, installati
 			return alertmanagerconfig.Receiver{}, err
 		}
 		httpConfig = httpConfigCp
-		// httpConfig = cfg.Global.HTTPConfig
 	}
 	httpConfig.BasicAuth = &promcommonconfig.BasicAuth{
 		Password: promcommonconfig.Secret(opsgenieKey),
-	}
-
-	if cfg.Global != nil && cfg.Global.HTTPConfig != nil {
-		if reflect.DeepEqual(cfg.Global.HTTPConfig, httpConfig) {
-			return alertmanagerconfig.Receiver{}, fmt.Errorf("HTTPConfig not cloned correctly")
-		}
 	}
 
 	r := alertmanagerconfig.Receiver{
--- a/service/controller/resource/heartbeatrouting/receiver/receiver.go
+++ b/service/controller/resource/heartbeatrouting/receiver/receiver.go
@@ -25,7 +25,7 @@ func toReceiver(cfg alertmanagerconfig.Config, cluster metav1.Object, installati
 	if cfg.Global != nil && cfg.Global.HTTPConfig != nil {
 		httpConfigCp, err := cloneHttpConfig(cfg.Global.HTTPConfig)
 		if err != nil {
-			return alertmanagerconfig.Receiver{}, err
+			return alertmanagerconfig.Receiver{}, microerror.Mask(err)
 		}
 		httpConfig = httpConfigCp
 	}
@@ -105,12 +105,12 @@ func getReceiver(cfg alertmanagerconfig.Config, receiver alertmanagerconfig.Rece
 func cloneHttpConfig(hc *promcommonconfig.HTTPClientConfig) (*promcommonconfig.HTTPClientConfig, error) {
 	s, err := yaml.Marshal(hc)
 	if err != nil {
-		return nil, err
+		return nil, microerror.Mask(err)
 	}
 	r := &promcommonconfig.HTTPClientConfig{}
 	err = yaml.Unmarshal(s, r)
 	if err != nil {
-		return nil, err
+		return nil, microerror.Mask(err)
 	}
 	return r, nil
 }
--- a/service/controller/resource/heartbeatrouting/receiver/receiver.go
+++ b/service/controller/resource/heartbeatrouting/receiver/receiver.go
@@ -5,6 +5,7 @@ import (
 	"net/url"
 	"reflect"
 
+	"gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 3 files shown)


mig4:
Is cloning via YAML Marshal/Unmarshal roundtripping too dirty? This structure has a couple levels of nested pointers to structs, some of them from stdlib (like `URL`). If not a hack like this I'd be tempted to use a cloning lib like https://github.com/jinzhu/copier or propose an implementation of `DeepCopy` upstream.

service/controller/resource/heartbeatrouting/receiver/receiver.go
```diff
@@ -14,24 +15,39 @@ import (
...
 	}
 
+	httpConfig := &promcommonconfig.HTTPClientConfig{}
+	if cfg.Global != nil && cfg.Global.HTTPConfig != nil {
+		httpConfigCp, err := cloneHttpConfig(cfg.Global.HTTPConfig)
+		if err != nil {
+			return alertmanagerconfig.Receiver{}, err
+		}
+		httpConfig = httpConfigCp
+		// httpConfig = cfg.Global.HTTPConfig
```
TheoBrigitte, line 31: This line can be removed

service/controller/resource/heartbeatrouting/receiver/receiver.go
```diff
@@ -14,24 +15,39 @@ import (
...
+	}
+	httpConfig.BasicAuth = &promcommonconfig.BasicAuth{
+		Password: promcommonconfig.Secret(opsgenieKey),
+	}
+
+	if cfg.Global != nil && cfg.Global.HTTPConfig != nil {
+		if reflect.DeepEqual(cfg.Global.HTTPConfig, httpConfig) {
```
TheoBrigitte, line 38: Aren't two struct holding identical data but at different address equal ?
This is something that should be test and not done at runtime, because I do not see how copy can fail at runtime.

service/controller/resource/heartbeatrouting/receiver/receiver.go
```diff
@@ -92,3 +108,16 @@ func getReceiver(cfg alertmanagerconfig.Config, receiver alertmanagerconfig.Rece
...
 	return nil, -1
 }
+
+func cloneHttpConfig(hc *promcommonconfig.HTTPClientConfig) (*promcommonconfig.HTTPClientConfig, error) {
+	s, err := yaml.Marshal(hc)
+	if err != nil {
+		return nil, err
+	}
+	r := &promcommonconfig.HTTPClientConfig{}
+	err = yaml.Unmarshal(s, r)
```
TheoBrigitte, line 118: That's a good and simple way of doing cloning, nice :)

service/controller/resource/heartbeatrouting/receiver/receiver_test.go
```diff
@@ -88,6 +88,40 @@ func TestEnsureReceiver(t *testing.T) {
...
+						WebhookConfigs: []*alertmanagerconfig.WebhookConfig{
+							&alertmanagerconfig.WebhookConfig{
+								URL: &alertmanagerconfig.URL{
+									URL: u,
+								},
+								HTTPConfig: &promcommonconfig.HTTPClientConfig{
+									BasicAuth: &promcommonconfig.BasicAuth{
+										Password: "secret-key",
+									},
+								},
```
TheoBrigitte, line 112: Where is the result of this test with the merged proxy url ?

TheoBrigitte (commented):
I'm now wondering: why don't we just set the proxy url per provider instead of doing this merge thingy ?

service/controller/resource/heartbeatrouting/receiver/receiver.go
mig4: Oh yeah, sorry this was just me testing to see if the cloning works. I'll remove this.

service/controller/resource/heartbeatrouting/receiver/receiver.go
mig4: Ditto, I only used it to make sure that the cloning works (e.g. this fails if I uncomment the line `httpConfig = cfg.Global.HTTPConfig`, but works with `httpConfig = httpConfigCp`), we don't need this at runtime so I'll remove this.

service/controller/resource/heartbeatrouting/receiver/receiver.go
mig4: Cool, I'll go ahead with this approach :smile:

service/controller/resource/heartbeatrouting/receiver/receiver_test.go
mig4: If I'm reading this correctly, the way these test are structured is we don't specify the expected result, we just specify the source and that we expect something to change.

I'd be open to refactoring this a bit to be a bit more explicit and specify the expected result, WDYT?

mig4:
> I'm now wondering: why don't we just set the proxy url per provider instead of doing this merge thingy ?

We could do that too, but we'll need the global proxy setting in AM anyway for the receivers we don't manage through PMO (like all the Slack ones), so I thought it'd be more consistent to use the same source of the settings here. But yeah, right now it's the same value, so the effect would be the same.

service/controller/resource/heartbeatrouting/receiver/receiver.go
mig4: Done.

service/controller/resource/heartbeatrouting/receiver/receiver.go
mig4: Done.

mig4:
Tested on _ghost_, works OK, I can see the `proxy_url` setting in the generated receiver config and I can see AlertManager connections to OpsGenie going through the proxy:

```
10.0.128.103:49058: clientconnect
10.0.128.103:49054: clientconnect
10.0.128.103:49056: clientconnect
::ffff:10.0.128.103:49058: Set new server address: hooks.slack.com:443
::ffff:10.0.128.103:49056: Set new server address: api.opsgenie.com:443
::ffff:10.0.128.103:49054: Set new server address: hooks.slack.com:443
::ffff:10.0.128.103:49058: serverconnect
  -> ('hooks.slack.com', 443)
::ffff:10.0.128.103:49056: serverconnect
  -> ('api.opsgenie.com', 443)
::ffff:10.0.128.103:49054: serverconnect
  -> ('hooks.slack.com', 443)
::ffff:10.0.128.103:49054: serverdisconnect
  -> ('hooks.slack.com', 443)
10.0.128.103:49054: clientdisconnect
```

service/controller/resource/heartbeatrouting/receiver/receiver_test.go
TheoBrigitte: This test ensure, that the whole update logic is correct and does update only when it needs to.
Maybe write a different test for this, because here we want to test that the proxy url is propagated from global to receiveirs, its a different test IMO.

TheoBrigitte:
> > I'm now wondering: why don't we just set the proxy url per provider instead of doing this merge thingy ?
> 
> We could do that too, but we'll need the global proxy setting in AM anyway for the receivers we don't manage through PMO (like all the Slack ones), so I thought it'd be more consistent to use the same source of the settings here. But yeah, right now it's the same value, so the effect would be the same.

Yeah you're right, let's use the same piece of code this will be easier.

service/controller/resource/heartbeatrouting/receiver/receiver_test.go
mig4: Done.

mig4:
We tested this on _beaver_ now and it works OK :tada: 

@TheoBrigitte PTAL, if the code looks all right we can go ahead and merge. Thanks!

service/controller/resource/heartbeatrouting/receiver/receiver_test.go
```diff
@@ -179,6 +146,60 @@ func TestEnsureReceiver(t *testing.T) {
 	}
 }
 
+func TestEnsureReceiverWithProxy(t *testing.T) {
+	// The HTTPConfig from global config should be merged when creating
+	// HTTPConfig for the receiver
```
TheoBrigitte, line 151: Suggestion for a better comment, making it a valid go comment is my main point here.
```suggestion
// TestEnsureReceiverWithProxy ensures the global HTTPConfig is merged into receiver HTTPConfig.
func TestEnsureReceiverWithProxy(t *testing.T) {
```

mig4: Lol, I meant to do that and it slipped my mind, thanks!

TheoBrigitte (approved):
Beautiful <3

PR #558, opened 2021-05-19
Bump prometheus default value to 2.27.1

QuentinBisson:
Upgrade prometheus to 2.27.1

## Checklist

I have:

### DIFF
--- a/main.go
+++ b/main.go
@@ -115,7 +115,7 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().StringSlice(f.Service.Prometheus.Bastions, make([]string, 0), "Address of the bastions.")
 	daemonCommand.PersistentFlags().Bool(f.Service.Prometheus.Storage.CreatePVC, false, "Should the operator create a PVC for storage.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Storage.Size, "100Gi", "Storage size for prometheus.")
-	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Version, "v2.24.0", "Prometheus container image version.")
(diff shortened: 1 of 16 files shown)


main.go
```diff
@@ -115,7 +115,7 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().StringSlice(f.Service.Prometheus.Bastions, make([]string, 0), "Address of the bastions.")
 	daemonCommand.PersistentFlags().Bool(f.Service.Prometheus.Storage.CreatePVC, false, "Should the operator create a PVC for storage.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Storage.Size, "100Gi", "Storage size for prometheus.")
-	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Version, "v2.24.0", "Prometheus container image version.")
+	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Version, "v2.27.1", "Prometheus container image version.")
```
TheoBrigitte, line 118: I feel like we should also align the go dependency here.

Also please make sure communication with alertmanager still works because of this upstream change

```
[CHANGE] Alerting: Using Alertmanager v2 API by default. #8626
```
see https://github.com/prometheus/prometheus/releases/tag/v2.26.0

We probably have to change it here from `v1` to `v2`: https://github.com/giantswarm/prometheus-meta-operator/blob/master/service/controller/resource/alertmanagerconfig/resource.go#L26

QuentinBisson: I updated it and it is working:
image


BTW, v1 is going away in 0.22

TheoBrigitte (approved):
Nice

PR #560, opened 2021-05-21
add atlas and installation tags on Heartbeats

TheoBrigitte:


service/controller/resource/heartbeat/resource.go
```diff
@@ -78,6 +79,15 @@ func toHeartbeat(v interface{}, installation string, pipeline string) (*heartbea
 		return nil, microerror.Mask(err)
 	}
 
+	// They need to be sorted alphabetically to avoid unecessary heartbeat update
```
JosephSalisbury, line 82:
```suggestion
	// They need to be sorted alphabetically to avoid unnecessary heartbeat update
```

JosephSalisbury (approved):

PR #563, opened 2021-05-21
update PrometheusFailsToCommunicateWithRemoteStorageAPI alert

TheoBrigitte:
Prometheus servers running in china seems to never have sent data to cortex, and the alert about remote write never fired for them

image

So I updated the alert to take into account the `prometheus_remote_storage_metadata_retried_total` metric

image


QuentinBisson (approved):

TheoBrigitte:
btw, this will page for China cluster.

PR #564, opened 2021-05-21
Add Proxy URL to remote write

QuentinBisson:
## Checklist

I have:

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -4,6 +4,7 @@ package service
 
 import (
 	"context"
+	"os"
 	"sync"
 
 	"github.com/giantswarm/k8sclient/v5/pkg/k8sclient"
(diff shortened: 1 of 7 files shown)


service/service.go
```diff
@@ -146,7 +146,9 @@ func New(config Config) (*Service, error) {
 			RemoteWriteURL:      config.Viper.GetString(config.Flag.Service.Prometheus.RemoteWrite.URL),
 			RemoteWriteUsername: config.Viper.GetString(config.Flag.Service.Prometheus.RemoteWrite.BasicAuth.Username),
 			RemoteWritePassword: config.Viper.GetString(config.Flag.Service.Prometheus.RemoteWrite.BasicAuth.Password),
+			ProxyURL:            config.Viper.GetString(config.Flag.Service.Installation.ProxyURL),
```
mig4, line 149: I thought we agreed in refinement to pass the value through from PMO env variables to avoid having another flag as it's unlikely you'd want PMO to use a proxy but Prometheus to not use it or use a different one. Kinda like they do it in [kvm-operator](https://github.com/giantswarm/kvm-operator/blob/be5d19eddf8cae39b11976c35b79185d6d645e6d/service/controller/cloudconfig/master_template.go#L55-L57).

I'm not completely opposed to adding a flag or anything, just wondering why?

Also, if going the flag route, I'd suggest renaming it to `PrometheusProxyURL` to be clear that this is for Prometheus and not for PMO itself.

QuentinBisson: We can definitely get it from an env variable and I will fix it but I cannot name it PrometheusProxyURL as it will be used for alertmanager as well :D

QuentinBisson:
Tested on beaver

QuentinBisson:
@giantswarm/team-atlas can you take another look?

mig4 (approved):

PR #569, opened 2021-05-26
increase-time-for-etcd-alerts

calvix:
towards https://github.com/giantswarm/giantswarm/issues/17199

these seem to page when the cluster is upgrading from single master to multi-master HA, but resolve afterward

I think we should be safe to increase the duration for the alerts, specially fro the DB size
and if the Etcd has really "no leader" we would get paged by tons of other alerts as API would be basically dead

## Checklist

I have:


whites11 (approved):
LGTM

anvddriesch (approved):

PR #573, opened 2021-05-26
Rename config names

QuentinBisson:
## Checklist

I have:


TheoBrigitte (commented):
Why do we need to pass all those new `Promtheus*` values around?

QuentinBisson:
Sorry for not explaining. Basically, this is to namespace prometheus values because they conflict with alertmanager properties. They both have a Version, an Address and so on. Eventually, i'm thinking WE should use objets instead but I thought this would do for now

TheoBrigitte (approved):
Ok, thanks for the explanation.

Yeah it does feel silly to pass all those values around. Maybe we should just pass the entire flag struct instead. Let's refactor this later.

mig4:
> Yeah it does feel silly to pass all those values around. Maybe we should just pass the entire flag struct instead. Let's refactor this later.

BIG +1 from me, let's make some re-usable structs, it's so annoying that when you add a flag you need to add it in dozen different places.

PR #574, opened 2021-05-27
Namespace codebase between monitoring and monitoring

QuentinBisson:
This PR is namespacing monitoring and alerting code to prepare for the addition of the alertmanager code

## Checklist

I have:


mig4:
>  Described why this change is being introduced

So, why do we need that? 😄 It seems there are still dependencies between the resources, and in fact maybe we'd want to generalise some resources to re-use them between controllers (like we have/had with certificate resource)?

QuentinBisson:
Yes we can try to reuse resources. I am just thinking namespacing will help as having to see all resources in one place makes it hard to see everything. Having a proper namespacing makes it easier in my opinion.
Ideally, I would not reuse the resource entirely but create a service for specific things like creating an ingress though

TheoBrigitte (approved):

PR #581, opened 2021-06-01
Add Alertmanager

QuentinBisson:
This PR adds support for prometheus operator managed alertmanager.
This adds the alertmanager resource  which creates alertmanager CRs as well as the alertmanagerconfigsecret resource which creates the mininal mandatory configuration alertmanager needs. 
I copied the config from g8s-prometheus and got it working but we might need to clean it up some more

This PR also adds support to override Prometheus Log level when we need to debug it.

## Checklist

I have:


.nancy-ignore
```diff
@@ -1,4 +1,4 @@
 # Ignore CVE fixed in prometheus 2.7.1 as the tag version is using a newer prometheus version
 CVE-2019-3826
 # Ignore CVE jwt-go before 4.0.0-preview1 allows attackers to bypass intended access restrict...https://github.com/dgrijalva/jwt-go/issues/290
-CVE-2020-26160 until=2021-06-01
+CVE-2020-26160
```
TheoBrigitte, line 4: This was fixed in master. I finally merged https://github.com/giantswarm/prometheus-meta-operator/pull/538

service/controller/resource/alerting/alertmanager/resource.go
```diff
@@ -0,0 +1,200 @@
...
+											"master",
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
```
TheoBrigitte, line 176: This is a bit overkill for now as we only have 1 replica, but good for the future when we move to HA alertmanager.

service/controller/resource/alerting/alertmanager/resource.go
```diff
@@ -0,0 +1,200 @@
...
+					TopologyKey:       "kubernetes.io/hostname",
+					WhenUnsatisfiable: corev1.ScheduleAnyway,
+					LabelSelector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{
+							"app.kubernetes.io/name": "alertmanager",
+						},
+					},
+				},
+			},
+			PriorityClassName: "prometheus",
```
TheoBrigitte, line 188: This name feels super weird now. We should change it to something like `monitoring` but you can leave this for now.

service/controller/resource/alerting/alertmanagerconfigsecret/test/case-0-cluster-api.golden
```diff
@@ -0,0 +1,446 @@
...
+receivers:
+- name: root
+
+- name: heartbeat
+  webhook_configs:
+  - url: https://api.opsgenie.com/v2/heartbeats/test-heartbeat/ping
+    http_config:
+      basic_auth:
+        password: opsgenie-key
+    send_resolved: false
```
TheoBrigitte, line 123: Those top/installation level heartbeat are gone, heartbeat are now dynamically configure by PMO. We should remove this.

TheoBrigitte (changes requested):
That's fairly large, I hope I didn't missed anything.

service/controller/resource/alerting/alertmanager/resource.go
QuentinBisson: I agree. I was planning on going with HA first but the gossip protocol was not working as expected but I kept the topology

.nancy-ignore
QuentinBisson: Let me rebase master then :)

service/controller/resource/alerting/alertmanager/resource.go
QuentinBisson: I agree and we should do it also for the namespace and so on but this will require a lot of things that I would rather not have in this PR

service/controller/resource/alerting/alertmanagerconfigsecret/test/case-0-cluster-api.golden
QuentinBisson: You are right

service/controller/resource/alerting/alertmanager/resource.go
TheoBrigitte: Yeah fine, keep it.

mig4 (changes requested):
Dude, I'm sorry but this is way too big. You ticked these off in the description:

> - [x] Described why this change is being introduced
> - [x] Separated out refactoring/reformatting in a dedicated PR

yet there are both refactoring changes, reformatting changes and functionality changes. I understand that when you're hacking you need to try this or that to get things going, but please understand that by dumping everything like that you're putting the burden on the reviewer to go through, figure out what's the reason for each change and see if there are any issues which becomes a monstrous task given how much context you need to build up and hold in your head to do this. That means it takes time, and if I as a reviewer don't have spare hours to spend on it now that means I'm putting it off.

This is not a race (at least for me), I'd much prefer if we agreed to take time to prepare not only the code, but also the presentation to make it easier for reviewers. They are here to help you catch issues or make improvements before they get deployed.

As for the contents of the changes, from my skimming through this my main question is would we not want to migrate the routes/receivers to [AlertManagerConfigSpec](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#alertmanagerconfigspec) resources?

I get the feeling that if we don't do it straight away we're never going to do it, same as with `ServiceMonitors`, and I'm really not keen on dealing with more templated YAML.

QuentinBisson:
@mig4 I understand that you would think that but I already did my best to scratch any refactoring from there. Sure, there is a change from PrometheusVersion to Version and Labels to PrometheusLabels. The only reason I refactored them is that it made sense in the scope of this PR and I would honestly not do the change it not in a PR

This PR is already the MVP to get alertmanager running with the less amount of changes as described at the top of my PR.
My hacking/draft PR is 3 times bigger than this so I configured the minimum for it to work to not have a broken master branch. I can remove the 2 renaming if you want but TBH, I would not call this refactoring

Regarding your last point, sure, I did it in my main PR branch (we have one alertmanager config per cluster) and we can definitely add one for the team routes but we will need a basic config as notification templates can only be defined using the alertmanagerconfigsecret.

mig4:
> @mig4 I understand that you would think that but I already did my best to scratch any refactoring from there. Sure, there is a change from PrometheusVersion to Version and Labels to PrometheusLabels. The only reason I refactored them is that it made sense in the scope of this PR and I would honestly not do the change it not in a PR

There's that, there's changing `app` label to `app.kubernetes.io/name`, there's reformatting of struct slices where you remove the types from elements (e.g. `corev1.NodeSelectorTerm`), and those of course cause changes to the `.golden` files in tests, and I'm not sure if any of this has anything to do with adding AlertManager?

> This PR is already the MVP to get alertmanager running with the less amount of changes as described at the top of my PR.
> My hacking/draft PR is 3 times bigger than this so I configured the minimum for it to work to not have a broken master branch. I can remove the 2 renaming if you want but TBH, I would not call this refactoring.

It is OK to split a PR into a series where individual ones are not _"working"_, i.e. maybe not like they're breaking the build but introducing changes that on their own do nothing, but add boilerplate for another more significant change. One example could be adding CLI flags, as there's currently a lot of boilerplate associated with them, etc. You can also set base of a PR to a branch from another PR and then merge them one-by-one in order, etc.

We had a session a while back that touched on some techniques people use to split PRs, if you would find it useful we can come up with some concrete examples and schedule something again?

Anyway, sure let's get this through as-is, no point splitting it now when we already went through it, but my main question is, do you see where I'm coming from? You complained that we don't do reviews fast enough but believe me, if it was easier, I'd go through them faster.

I really really highly recommend reading [Refactor and make changes in different commits](http://jakegoulding.com/blog/2012/11/04/refactor-and-make-changes-in-different-commits/).

> Regarding your last point, sure, I did it in my main PR branch (we have one alertmanager config per cluster) and we can definitely add one for the team routes but we will need a basic config as notification templates can only be defined using the alertmanagerconfigsecret.

Yeah I know, there's no way to map global config into prometheus-operator API yet. That's fine. Just asking the question for now because you had to introduce the config here, if this was more or less copied-and-pasted from g8s-prometheus, we can do the refactoring in a separate PR. If you think the difference between removing the template and adding an object is minimal then it would make sense to do it here. WDYT?

BTW, it makes me very happy that you have this change in WIP, now I'm less worried that we're never going to change that and stay with templated YAML forever. Thank you :heart:

QuentinBisson:
Hopefully, once the flag PR is merged, this should be easier to review :D. I can still split the resources if needed

files/templates/alertmanager/alertmanager.yaml
```diff
@@ -0,0 +1,462 @@
+global:
```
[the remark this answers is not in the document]
QuentinBisson:
```diff
colordiff helm/g8s-prometheus/templates/alertmanager/configmap.yaml /home/quentin/go/src/github.com/giantswarm/prometheus-meta-operator/files/templates/alertmanager/alertmanager.yaml
1d0
< 
3,10d1
<   {{- with .Values.Installation.V1 }}
<   {{- if .Proxy }}
<   {{- if .Proxy.Enabled }}
<   http_config:
<     proxy_url: {{ if .Secret.Proxy.HTTPS }}{{ .Secret.Proxy.HTTPS }}{{ else }}{{ .Secret.Proxy.HTTP }}{{ end }}
<   {{- end }}
<   {{- end }}
<   {{- end }}
12,13c3,7
< 
<   slack_api_url: {{ .Values.Installation.V1.Secret.Alertmanager.Slack.ApiUrl }}
---
> [[- if .ProxyURL ]]
>   http_config:
>     proxy_url:  [[ .ProxyURL ]]
> [[- end ]]
>   slack_api_url: [[ .SlackApiURL ]]
16c10
< - /etc/alertmanager-templates/*.tmpl
---
> - '*.tmpl'
20d13
< 
24d16
< 
26d17
< 
28,34c19
<   - receiver: alertmanager_failing_slack
<     match:
<       alertname: AlertManagerNotificationsFailing
<       integration: opsgenie
<     continue: false
< 
<   # Team Ops Opsgenie
---
>     # Team Ops Opsgenie
63c48
<   {{ if eq .Values.Installation.V1.Provider.Kind "aws" }}
---
>   [[ if eq .Provider "aws" ]]
65c50
<   {{ else if eq .Values.Installation.V1.Provider.Kind "azure" }}
---
>   [[ else if eq .Provider "azure" ]]
67c52
<   {{ else if eq .Values.Installation.V1.Provider.Kind "kvm" }}
---
>   [[ else if or (eq .Provider "kvm") (eq .Provider "vmware") ]]
69c54
<   {{ end }}
---
>   [[ end ]]
132,171d116
< - name: alertmanager_failing_slack
<   slack_configs:
<   - channel: '#alertmanager_failing'
<     send_resolved: true
<     actions:
<     - type: button
<       text: ':green_book: OpsRecipe'
<       url: 'https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/{{`{{ (index .Alerts 0).Annotations.opsrecipe }}`}}'
<       style: '{{`{{ if eq .Status "firing" }}primary{{ else }}default{{ end }}`}}'
<     - type: button
<       text: ':mag: Query'
<       url: '{{`{{ (index .Alerts 0).GeneratorURL }}`}}'
<     - type: button
<       text: ':grafana: Dashboard'
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
<     - type: button
<       text: ':no_bell: Silence'
<       url: '{{`{{ template "__alert_silence_link" . }}`}}'
<       style: '{{`{{ if eq .Status "firing" }}danger{{ else }}default{{ end }}`}}'
< 
< - name: project_customer_slack
<   slack_configs:
<   - channel: {{ .Values.Installation.V1.Slack.ProjectName }}
<     send_resolved: true
<     actions:
<     - type: button
<       text: ':green_book: OpsRecipe'
<       url: 'https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/{{`{{ (index .Alerts 0).Annotations.opsrecipe }}`}}'
<       style: '{{`{{ if eq .Status "firing" }}primary{{ else }}default{{ end }}`}}'
<     - type: button
<       text: ':mag: Query'
<       url: '{{`{{ (index .Alerts 0).GeneratorURL }}`}}'
<     - type: button
<       text: ':grafana: Dashboard'
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
<     - type: button
<       text: ':no_bell: Silence'
<       url: '{{`{{ template "__alert_silence_link" . }}`}}'
<       style: '{{`{{ if eq .Status "firing" }}danger{{ else }}default{{ end }}`}}'
< 
186c131
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
209c154
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
217,219c162
<   {{- if .Values.Installation.V1.Infra.TestingEnvironment }}
<   - channel: '#alert-batman-test'
<   {{- else }}
---
>   [[- if eq .Pipeline "stable" ]]
221c164,166
<   {{- end }}
---
>   [[- else ]]
>   - channel: '#alert-batman-test'
>   [[- end ]]
236c181
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
259c204
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
267,269c212
<   {{- if .Values.Installation.V1.Infra.TestingEnvironment }}
<   - channel: '#alert-celestial-test'
<   {{- else }}
---
>   [[- if eq .Pipeline "stable" ]]
271c214,216
<   {{- end }}
---
>   [[- else ]]
>   - channel: '#alert-celestial-test'
>   [[- end ]]
286c231
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
294,296c239
<   {{- if .Values.Installation.V1.Infra.TestingEnvironment }}
<   - channel: '#alert-firecrack-test'
<   {{- else }}
---
>   [[- if eq .Pipeline "stable" ]]
298c241,243
<   {{- end }}
---
>   [[- else ]]
>   - channel: '#alert-firecracker-test'
>   [[- end ]]
313c258
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
321,323c266
<   {{- if .Values.Installation.V1.Infra.TestingEnvironment }}
<   - channel: '#alert-ludacris-test'
<   {{- else }}
---
>   [[- if eq .Pipeline "stable" ]]
325c268,270
<   {{- end }}
---
>   [[- else ]]
>   - channel: '#alert-ludacris-test'
>   [[- end ]]
340c285
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
348,350c293
<   {{- if .Values.Installation.V1.Infra.TestingEnvironment }}
<   - channel: '#alert-rocket-test'
<   {{- else }}
---
>   [[- if eq .Pipeline "stable" ]]
352c295,297
<   {{- end }}
---
>   [[- else ]]
>   - channel: '#alert-rocket-test'
>   [[- end ]]
367c312
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
375,381c320,321
<   - api_key: {{ .Values.Installation.V1.Secret.Alertmanager.Opsgenie.ApiKey }}
<     tags: "{{`{{ (index .Alerts 0).Labels.alertname }},{{ (index .Alerts 0).Labels.cluster_type }},{{ (index .Alerts 0).Labels.severity }},{{ (index .Alerts 0).Labels.team }}`}},{{ .Values.Installation.V1.Provider.Kind }},{{ .Values.Installation.V1.Monitoring.Prometheus.ClusterLabel }},
<     {{- if .Values.Installation.V1.Infra.TestingEnvironment }}
<     testing
<     {{- else }}
<     stable
<     {{- end }}"
---
>   - api_key: [[ .OpsgenieKey ]]
>     tags: "{{`{{ (index .Alerts 0).Labels.alertname }},{{ (index .Alerts 0).Labels.cluster_type }},{{ (index .Alerts 0).Labels.severity }},{{ (index .Alerts 0).Labels.team }}`}},[[ .Provider ]],[[ .Installation ]],[[ .Pipeline ]]"
385,387c325
<   {{- if .Values.Installation.V1.Infra.TestingEnvironment }}
<   - channel: '#alert-test-{{ .Values.Installation.V1.Name }}'
<   {{- else }}
---
>   [[- if eq .Pipeline "stable" ]]
389c327,329
<   {{- end }}
---
>   [[- else ]]
>   - channel: '#alert-test-[[ .Installation ]]'
>   [[- end ]]
404c344
<       url: '{{ .Values.Installation.V1.Monitoring.Grafana.Address }}/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
---
>       url: '[[ .GrafanaAddress ]]/{{`{{ (index .Alerts 0).Annotations.dashboard }}`}}'
416d355
< 
419c358
<     cluster_id: '{{ .Values.Installation.V1.Name }}'
---
>     cluster_id: '[[ .Installation ]]'
422d360
< 
427d364
< 
433d369
< 
439d374
< 
445d379
< 
451d384
< 
457d389
< 
463d394
< 
469d399
< 
475d404
< 
481d409
< 
487d414
< 
493d419
< 
499d424
< 
505d429
< 
511d434
< 
517d439
< 
523d444
< 
529d449
< 
535d454
< 
540d458
< 
```

files/templates/alertmanager/notification-template.yaml
```diff
@@ -0,0 +1,52 @@
+{{ define "__alertmanager" }}AlertManager{{ end }}
```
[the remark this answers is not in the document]
QuentinBisson: The file is the same here :)

mig4:
Hmm, something broke in tests. Any ideas?

mig4:
Oh, does this need to be rebased on master? Or after merging #592 ?

QuentinBisson:
It should be better now @mig4

service/controller/resource/alerting/alertmanager/resource.go
```diff
@@ -0,0 +1,200 @@
...
+
+func getObjectMeta(v interface{}) (metav1.ObjectMeta, error) {
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return metav1.ObjectMeta{}, microerror.Mask(err)
+	}
+
+	return metav1.ObjectMeta{
+		Name:      key.ClusterID(cluster),
+		Namespace: key.Namespace(cluster),
```
mig4, line 68: Shouldn't this remain in `monitoring` NS? I mean, it's still a "global" resource, and AFAIK `$INSTALLATION-prometheus` is for resources specific to the Prometheus instance dedicated to the MC.

service/controller/resource/alerting/alertmanagerconfigsecret/resource.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func getObjectMeta(v interface{}, config Config) (metav1.ObjectMeta, error) {
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return metav1.ObjectMeta{}, microerror.Mask(err)
+	}
+
+	return metav1.ObjectMeta{
+		Name:      fmt.Sprintf("alertmanager-%s", config.Installation),
```
mig4, line 95: I'd have a preference to call it just `alertmanager` due to same points as for the namespace - i.e. this isn't an AlertManager instance specific to the MC, but a "global" one for MC and all WCs, right?

service/controller/resource/alerting/alertmanagerconfigsecret/resource.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	notificationTemplate, err := ioutil.ReadFile(path.Join(templateDirectory, notificationTemplatePath))
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	alertmanagerConfigSecret, err := toData(v, config)
```
mig4, line 111: A nitpick, but it confused me a bit as `toData` returns the contents of the config file to be included in the secret, not the secret itself.

```suggestion
	alertmanagerConfig, err := toData(v, config)
```

service/controller/resource/alerting/alertmanagerconfigsecret/resource.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, microerror.Mask(err)
+	}
+
+	return data, nil
+}
+
+func getTemplateData(cluster metav1.Object, config Config) (*TemplateData, error) {
+
+	var proxyURL *string = nil
+	if !strings.Contains(config.NoProxy, "api.opsgenie.com") {
```
mig4, line 151: This is a bit of an edge case (i.e. I don't see why would someone want to do this, but) but with this if `NoProxy == "api.opsgenie.com"` we won't include proxy URL in the global config, but we still need it for slack and possibly other receivers.

Personally, I'd just skip the `NoProxy` check as it's more meant to filter out services internal to the cluster and AM only talks to external services. But if you think it would be prudent to have it, we'll need to check all URLs AM might try to talk to.

service/key/key.go
```diff
@@ -84,6 +84,14 @@ func EtcdSecret(obj interface{}) string {
 	return Secret()
 }
 
+func AlertmanagerLabels(cluster metav1.Object) map[string]string {
+	return map[string]string{
+		"app.kubernetes.io/name":       "alertmanager",
+		"app.kubernetes.io/managed-by": project.Name(),
+		"app.kubernetes.io/instance":   ClusterID(cluster),
```
mig4, line 91: Same comment as resource name/namespace.

service/service.go
```diff
@@ -230,12 +232,22 @@ func New(config Config) (*Service, error) {
...
+			AlertmanagerAddress:     config.Viper.GetString(config.Flag.Service.Alertmanager.Address),
+			AlertmanagerBaseDomain:  config.Viper.GetString(config.Flag.Service.Alertmanager.BaseDomain),
+			AlertmanagerCreatePVC:   config.Viper.GetBool(config.Flag.Service.Alertmanager.Storage.CreatePVC),
+			AlertmanagerLogLevel:    config.Viper.GetString(config.Flag.Service.Alertmanager.LogLevel),
+			AlertmanagerStorageSize: config.Viper.GetString(config.Flag.Service.Alertmanager.Storage.Size),
+			AlertmanagerVersion:     config.Viper.GetString(config.Flag.Service.Alertmanager.Version),
+			GrafanaAddress:          config.Viper.GetString(config.Flag.Service.Grafana.Address),
+			SlackApiURL:             config.Viper.GetString(config.Flag.Service.Slack.ApiURL),
+			SlackProjectName:        config.Viper.GetString(config.Flag.Service.Slack.ProjectName),
+			OpsgenieKey:             config.Viper.GetString(config.Flag.Service.Opsgenie.Key),
```
mig4, line 244: Weren't these already merged?

mig4 (changes requested):

QuentinBisson:
@mig4 regarding AlertManagerNotificationsFailing, we discussed it in slack as it is not used and the project_customer_slack as well

service/controller/resource/alerting/alertmanager/resource.go
QuentinBisson: I can always move it there of course

service/controller/resource/alerting/alertmanagerconfigsecret/resource.go
QuentinBisson: How about alertmanager-config?

service/controller/resource/alerting/alertmanagerconfigsecret/resource.go
QuentinBisson: I see a few cases where this could apply. For example, we could proxy to an internal server for messaging but not opsgenie. This might not be a usecase we have now so I think we can remove it here for now

service/service.go
QuentinBisson: I did not add them to the controllers

files/templates/alertmanager/notification-template.yaml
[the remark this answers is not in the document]
TheoBrigitte: Nice

mig4:
> @mig4 regarding AlertManagerNotificationsFailing, we discussed it in slack as it is not used and the project_customer_slack as well

Sorry, I must've missed it. You mean `AlertManagerNotificationsFailing` is not used? I thought that indicates when AlertManager can't talk to OpsGenie, what do you use for that? Just heartbeat and then investigate?

QuentinBisson:
As you can see in g8s-prometheus, the AlertmanagerNotificationsFailing is only configured to route to a hidden slack channel that almost none of atlas are in:
https://github.com/giantswarm/g8s-prometheus/blob/f5ad29317b878a7d41f4dc17b00ed263cced1dc4/helm/g8s-prometheus/templates/alertmanager/configmap.yaml#L34
https://github.com/giantswarm/g8s-prometheus/blob/f5ad29317b878a7d41f4dc17b00ed263cced1dc4/helm/g8s-prometheus/templates/alertmanager/configmap.yaml#L138 

The AlertmanagerNotificationsFailing alert still exists and is still routed to atlas :)

mig4 (changes requested):
OK one more minor tweak I think, looks good otherwise.

mig4:
Not going to block on this or anything, but just out of curiosity, do we need to generate those resources programmatically at all? I.e. would it work to just have them in the Helm chart?

QuentinBisson:
Well, I did not think of this but  yes we could create them using the helm chart instead. One of the advantages with code is that the labels and so on are defined in on place and not in the helm chart and the code and pmo will always ensure alertmanager runs while helm would not if someone manually deletes it.
We can always switch it to the helm chart if we think it's better though

mig4 (approved):
OK, sounds good.

mig4:
Hey @QuentinBisson I just noticed a whole bunch of extra files appeared in this PR before you merged it, including code that creates Ingress programmatically. What's that about?

QuentinBisson:
I'm not really sure
I'm trying to understand what happened

QuentinBisson:
Can we remove one commit from the history : https://github.com/giantswarm/prometheus-meta-operator/pull/581/commits/37a9c9da26def9dcf27b4496dbf5908c927c4b39 ? It was probably left over on my laptop when I created the other branch and as I force pushed, it was pushed upstream as well :(

QuentinBisson:
https://github.com/giantswarm/prometheus-meta-operator/pull/610

mig4:
Ah, right yeah that'd explain it. Thanks for finding and fixing, approved #610 😄

QuentinBisson:
THanks for noticing it so fast :D

PR #583, opened 2021-06-02
Fix memory limit to avoid going over node capacity

mig4:
VPA takes the maxAllowed to be based on memory request, that means the
request can be set to almost max memory available to the node, but since
the limit is calculated based on the request and multiplied by 1.2 it
means the limit can go higher than available memory.

Prevent this by lowering the maxAllowed value by the request-limit
coefficient.

Towards: https://github.com/giantswarm/giantswarm/issues/17350

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fix
+
+- Fixed the way VPA `maxAllowed` parameter for memory is calculated so that we
+  avoid going over node memory capacity with the memory limit (`maxAllowed` is
+  used for request and limit is that multiplied by 1.2).
+
 ## [1.38.0] - 2021-05-28
 
 ### Changed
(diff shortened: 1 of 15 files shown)


TheoBrigitte:
Tested on `argali` :heavy_check_mark: 

#### Before 

request: 30G
limit: 36G

```
$ k describe po|grep -i Limits -A5
    Limits:
      cpu:     6420m
      memory:  35592873162480m
    Requests:
      cpu:        4280m
      memory:     29680050585
```
```
$ k describe vpa|grep -i max -A5
      Max Allowed:
        Cpu:     7
        Memory:  29680050585
      Min Allowed:
        Cpu:           100m
        Memory:        1073741824
```

#### After

request: 25G
limit: 30G

```
$ k describe vpa|grep -i max -A5
      Max Allowed:
        Cpu:     7
        Memory:  24733375487
      Min Allowed:
        Cpu:           100m
        Memory:        1073741824
```
```
$ k describe po|grep -i Limits -A5
    Limits:
      cpu:     6420m
      memory:  29680050565972m
    Requests:
      cpu:        4280m
      memory:     24733375487
```

TheoBrigitte (approved):
:ok_hand:

PR #588, opened 2021-06-04
Moves node-exporter to use SLO

JosephSalisbury:
Closes https://github.com/giantswarm/giantswarm/issues/14336

## Checklist

I have:


helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -36,6 +36,16 @@ spec:
         class: MEDIUM
         service: kubelet
       record: raw_slo_errors
+    - expr: count(up{app="node-exporter", role!="bastion"}) by (cluster_type, cluster_id)
+      labels:
+        class: MEDIUM
+        service: node-exporter
+      record: raw_slo_requests
+    - expr: sum((up{app='node-exporter', role!="bastion"} * -1) + 1) by (cluster_type, cluster_id)
```
MarcelMue, line 44: Can you explain what this does? I am also confused why this is here and not under `service-level-apps`

JosephSalisbury: This counts the number of node-exporters that are down.

up == 1 if the node-exporter is up, and up == 0 if it's down. With errors, we want to have a count of all the node-exporters that are down. So, * -1 to get -1 for up node-exporters and 0 for down node-exporters, and then add 1 to get 0 for up node-exporters, and 1 for down node-exporters. Then, sum them all up to get the number of node-exporters that are down. Math!

This isn't under `service-level-apps` because that is a different SLO implementation that we're planning on removing in future.

MarcelMue: Okay makes sense to me - is there a way to inhibit this by now? Like this should not fire if the kubelet SLO fires.

JosephSalisbury: Not yet, but we can totally bring it in without too much pain

MarcelMue (approved):
Fine with me.

mig4 (approved):
SLO 🧠 Nice 👍

JosephSalisbury:
fyi @giantswarm/sig-monitoring

QuentinBisson (approved):

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
QuentinBisson: Could you add what you added here as a comment to explain the SLI ?

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -56,14 +56,34 @@ spec:
...
       record: slo_target
 
+      # -- node-exporter
+    - expr: "count(up{app="node-exporter", role!="bastion"}) by (cluster_type, cluster_id)"
+      labels:
+        class: MEDIUM
+        area: kaas
+        service: node-exporter
+      record: raw_slo_requests
+    - expr: "sum((up{app='node-exporter', role!="bastion"} * -1) + 1) by (cluster_type, cluster_id)"
```
QuentinBisson, line 66: Can you just describe what this SLI does in a comment? Otherwise, it's fine for me to merge

calvix (approved):
that is some deep magig

mig4 (approved):

PR #589, opened 2021-06-04
Add prometheus log level

QuentinBisson:
This changes adds the possibility to override the prometheus log level to investigate issues more easily

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add the prometheus log level option
+
 ### Fixed
 
 - Fixed the way VPA `maxAllowed` parameter for memory is calculated so that we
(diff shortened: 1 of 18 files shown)


mig4 (commented):
The change is fine, the question is does this make sense to configure globally?

What I mean is that with this, in a scenario where you want to debug something, you have to change the PMO ConfigMap and PMO will roll this out to all the Prometheuses on an installation.

For me it's rather rare to be investigating all Prometheuses at the same time, usually problems are more localised, so what I'd do is scale down PMO and add the `logLevel` setting to the specific Prometheus I'm investigating.

WDYT? Do you have a use-case for changing log-level of all Prometheuses?

QuentinBisson:
I understand where you are going with this but I do think scaling down PMO to change the log level is not ideal. We should probably define sone kind of clustermonitoring CR that defines this in the future.

For now,  I still think this is useful to debug things at the installation level and it will definitely be useful for the logging story

JosephSalisbury:
This would be a case where having a separate CR instead of using the cluster CRs would be neat (we could then add log level there)

JosephSalisbury:
I should read all of Quentin's comment before reading :D

QuentinBisson:
I need to get better at commenting as well (i.e. TLDRs and so on)

mig4 (approved):
Fine with me then. Thanks!

PR #590, opened 2021-06-04
Clean up and rename a few internal variables

QuentinBisson:
This changes introduces refactoring to preparing for the addition of alertmanager and fixes some leftovers

## Checklist

I have:

### DIFF
--- a/service/controller/resource/monitoring/prometheus/resource.go
+++ b/service/controller/resource/monitoring/prometheus/resource.go
@@ -32,7 +32,6 @@ type Config struct {
 	Customer          string
 	Installation      string
 	Pipeline          string
-	PrometheusVersion string
 	Provider          string
 	Region            string
 	Registry          string
@@ -40,9 +39,11 @@ type Config struct {
 	RetentionDuration string
 	RetentionSize     string
 	RemoteWriteURL    string
-	HTTPProxy         string
-	HTTPSProxy        string
-	NoProxy           string
+	Version           string
+
+	HTTPProxy  string
+	HTTPSProxy string
+	NoProxy    string
 }
 
 func New(config Config) (*generic.Resource, error) {
@@ -78,7 +79,7 @@ func getObjectMeta(v interface{}) (metav1.ObjectMeta, error) {
 	return metav1.ObjectMeta{
 		Name:      key.ClusterID(cluster),
 		Namespace: key.Namespace(cluster),
-		Labels:    key.Labels(cluster),
+		Labels:    key.PrometheusLabels(cluster),
 	}, nil
 }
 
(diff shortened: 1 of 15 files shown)


service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -234,13 +235,13 @@ func toPrometheus(v interface{}, config Config) (metav1.Object, error) {
...
 			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
-				corev1.TopologySpreadConstraint{
+				{
 					MaxSkew:           1,
 					TopologyKey:       "kubernetes.io/hostname",
 					WhenUnsatisfiable: corev1.ScheduleAnyway,
 					LabelSelector: &metav1.LabelSelector{
 						MatchLabels: map[string]string{
-							"app": "prometheus",
+							"app.kubernetes.io/name": "prometheus",
```
QuentinBisson, line 244: This is the default in prometheus operator 0.47.0+

service/controller/resource/monitoring/prometheus/resource_test.go
```diff
@@ -23,13 +23,13 @@ func TestPrometheus(t *testing.T) {
...
 		Pipeline:          "testing",
 		Provider:          "provider",
-		PrometheusVersion: "v2.27.1",
 		Region:            "onprem",
 		Registry:          "quay.io",
 		StorageSize:       "50Gi",
 		RetentionDuration: "2w",
 		RetentionSize:     "45Gi",
 		RemoteWriteURL:    "http://grafana/api/prom/push",
+		Version:           "v2.27.1",
```
mig4, line 32: What's the reasoning behind this one? Asking cause just `Version` is not self-explanatory and could be confusing.

mig4 (commented):
Nice, thank you Quentin! 😄

service/controller/resource/monitoring/prometheus/resource_test.go
QuentinBisson: I only renamed it for the prometheus resource config and it is still called prometheus version everywhere else. Otherwise, we should theoretically name RemoteWriteURL PrometheusRemoteWriteURL and so on. I think the redundancy is not useful here

TheoBrigitte (approved):
Fine with me

QuentinBisson:
I will let michael time to get back before merging :D

service/controller/resource/monitoring/prometheus/resource_test.go
mig4: Oh, because the full path of this thing is `prometheus.Config.Version`, right, my bad, didn't notice that.

mig4 (approved):

PR #593, opened 2021-06-08
include managed apps in service level metrics

anvddriesch:
## Checklist

I have:


helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -19,6 +19,11 @@ spec:
         class: HIGH
         service: api-server
       record: raw_slo_errors
+    - expr: "scalar(vector(0.999))"
+      labels:
+        class: HIGH
+        service: api-server
+      record: slo_target
```
anvddriesch, line 26: This is the SLO we want for a given service. I added this because it is different for managed apps.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -36,57 +41,76 @@ spec:
...
-    - record: monitoring:managed_apps:number_of_pods_unavailable
-      expr: |
+    - expr: |
         kube_deployment_status_replicas_unavailable
         and on(deployment,cluster_id,cluster_type,namespace) 
         kube_deployment_labels{label_giantswarm_io_monitoring_basic_sli='true'}
+      labels:
+        class: MEDIUM
+        service: managed-apps
+      record: raw_slo_errors
```
anvddriesch, line 66: `raw_slo_errors` for managed apps are the same metrics as recorded before

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -36,57 +41,76 @@ spec:
...
+    - expr: |
+        kube_daemonset_status_desired_number_scheduled
+        and on(daemonset,cluster_id,cluster_type,namespace)
+        kube_daemonset_labels{label_giantswarm_io_monitoring_basic_sli='true'}
+      labels:
+        class: MEDIUM
+        service: managed-apps
+      record: raw_slo_requests
```
anvddriesch, line 83: We add the `raw_slo_requests` for managed apps here since they were not explicitly recorded before

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -36,57 +41,76 @@ spec:
...
+        kube_statefulset_labels{label_giantswarm_io_monitoring_basic_sli='true'}
+      labels:
+        class: MEDIUM
+        service: managed-apps
+      record: raw_slo_requests
+    - expr: "scalar(vector(0.995))"
+      labels:
+        class: MEDIUM
+        service: managed-apps
+      record: slo_target
```
anvddriesch, line 104: As you can see the target is different from `kubelet` and `api-server`. Now the issue is how to include this metric in the `ServiceLevelBurnRateTooHigh` alert. (Right now it is explicitely 0.001 but it should be 1-`slo_target (service)`

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -36,57 +41,76 @@ spec:
...
-    - expr: sum(raw_slo_errors) by (service, cluster_type, cluster_id)
-      record: slo_errors
-    - expr: sum(sum_over_time(raw_slo_errors[5m])) by (cluster_type, cluster_id, service, class) / sum(sum_over_time(raw_slo_requests[5m])) by (cluster_type, cluster_id, service, class)
-      record: slo_errors_per_request:ratio_rate5m
-    - expr: sum(sum_over_time(raw_slo_errors[30m])) by (cluster_type, cluster_id, service, class) / sum(sum_over_time(raw_slo_requests[30m])) by (cluster_type, cluster_id, service, class)
-      record: slo_errors_per_request:ratio_rate30m
-    - expr: sum(sum_over_time(raw_slo_errors[1h])) by (cluster_type, cluster_id, service, class) / sum(sum_over_time(raw_slo_requests[1h])) by (cluster_type, cluster_id, service, class)
-      record: slo_errors_per_request:ratio_rate1h
-    - expr: sum(sum_over_time(raw_slo_errors[6h])) by (cluster_type, cluster_id, service, class) / sum(sum_over_time(raw_slo_requests[6h])) by (cluster_type, cluster_id, service, class)
-      record: slo_errors_per_request:ratio_rate6h
```
anvddriesch, line 50: these recording rules moved downwards for some reason

JosephSalisbury:
i'd break this into two PRs - one to add support for variable objective values, and then another to rebase managed apps

anvddriesch:
makes sense!

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
tomahawk28: Are you moving this because these recordings should run later than other rules?
https://github.com/giantswarm/prometheus-meta-operator/pull/593/files#diff-f2635a7abec77c51bb996d248798f925b42a22eb713a28b021e65785042488cfR65

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -36,6 +47,103 @@ spec:
...
+    - expr: |
+        kube_statefulset_status_replicas
+        and on(statefulset,cluster_id,cluster_type,namespace, workload_name) 
+        kube_statefulset_labels{label_giantswarm_io_monitoring_basic_sli='true'}
+      labels:
+        class: MEDIUM
+        area: managed-apps
+        service: {{ $labels.workload_name }}
+      record: raw_slo_requests
+    - expr: sum by (service) (raw_slo_errors{area="managed-apps"} - raw_slo_errors{area="managed-apps"}) + 0.05 * 36
```
anvddriesch, line 141: this is a hack in order to get each value of "service" for managed apps with the same slo value. Let me know if you have a more elegant way

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
anvddriesch: True, that makes sense!

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
QuentinBisson: Can you explain the hack?

anvddriesch: Yeah.
So we want to have "0.05 * 36" as slo_target for all managed app services. (That is currently used in the managed service slo alerts) 
All the service names can be found in the `raw_slo_errors` metrics above (We use the "area=managed-apps' label to only get the applicable ones). 
We then substract this metric from itself so that it is always 0. Then we sum by `service` so that there is only one row for each service. The sum is also 0. Because of that we can just add "0.05 * 36" and receive the value we need.

QuentinBisson: Alright it make sense.
Now I can see a world where we define SLA using kubernetes annotations :D

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -48,45 +143,7 @@ spec:
...
-    # to our error budget, which is (30d * (1 - SLO)) (converted to seconds).
-    # Keep in mind that this can be >1 - it just means that our downtime in the last 30d was bigger than
-    # the error budget permits.
-    - record: monitoring:managed_apps:service_level:primary:error_budget_used
-      expr: |
-        monitoring:managed_apps:service_level:primary:helper_time_unavailable_longer_than_5m_over_30d/
-        (30*24*60*(1-scalar(monitoring:managed_apps:service_level:primary:slo)))
-    - record: monitoring:managed_apps:service_level:primary:slo
-      expr: "scalar(vector(0.995))"
+    - expr: slo_target*scalar(min(slo_burnrate_high))
```
anvddriesch, line 146: the `scalar(min())` here is needed to turn `slo_burnrate_high` into a single scalar value for multiplication. It does not change anything because slo_burnrate_high always has the same value.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
kopiczko: Is it ok to create recording rule with the same name and labels twice?

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
anvddriesch: Yeah. Jose beat me to it https://github.com/giantswarm/prometheus-meta-operator/pull/594/files

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -50,6 +50,62 @@ spec:
...
+        label_replace(kube_statefulset_status_replicas
+        and on(statefulset,cluster_id,cluster_type,namespace, workload_name) 
+        kube_statefulset_labels{label_giantswarm_io_monitoring_basic_sli='true'}, "service", "$1", "workload_name", "(.*)" )
+      labels:
+        class: MEDIUM
+        area: managed-apps
+      record: raw_slo_requests
+      # -- 99.5% availability
+    - expr: sum by (service) (raw_slo_errors{area="managed-apps"} - raw_slo_errors{area="managed-apps"}) + 1-0.995
+      record: slo_target
```
kopiczko, line 107: That's sneaky :D

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
anvddriesch: I think so. These metrics were called `monitoring:managed_apps:number_of_pods_unavailable` before and defined in exactly the same way.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
anvddriesch: :D It was the only thing I could come up with that would add all the services with the right `slo_target` without explicitly listing them.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
JosephSalisbury: How does that work? Does one overwrite the other? I don't get how you could have multiple time series with the exact same labels

piontec: They don't have the same labels. Only the first one has `daemonset` label key, the 2nd one `deployment` label key.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -50,6 +50,62 @@ spec:
         service: kubelet
       record: slo_target
 
+    # -- managed-apps
+    # record when pods of a daemonset with label "label_giantswarm_io_monitoring_basic_sli" are down
+    - expr: |
+        label_replace(kube_daemonset_status_number_unavailable
+        and on(daemonset,cluster_id,cluster_type,namespace, workload_name)
+        kube_daemonset_labels{label_giantswarm_io_monitoring_basic_sli='true'}, "service", "$1", "workload_name", "(.*)" )
```
piontec, line 58: Why do we copy `workload_name` to `service`? Is it a convention? I have a doubt: what if a single workload exposes multiple k8s API `Service` objects? Is not misleading?

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
anvddriesch: How I understand it the difference is the `workload_type` label from the original series. It can be `deployment`, `daemonset` or `statefulset`.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
piontec: I actually don't get these 2 lines... what do you want to do here? Insert target for all existing services?

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
kopiczko: Alright so the labels defined there are extra labels on top of what the underlying metric already has. All good then.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
kopiczko: It basically emits `slo_target` of 0.005 for all services from managed-apps area.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
anvddriesch: We use the `service` label to identify each service in the slo_ metrics (service="kubelet" for example). From my understanding, `workload_name` contains the name of the managed-app in question so like this we get a separate request and error metric for each of them.

anvddriesch:
> I actually don't get these 2 lines... what do you want to do here? Insert target for all existing services?
Basically, combine the 'service` labels of each managed app from above with the target_slo so we can use it later.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
anvddriesch: Exactly. I just do this stupid math with substracting from itself to make sure it is 0 before adding the value I want.

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
QuentinBisson: The service here is referring to the SLO (Service Level Objective) and has nothing to do with the Kubernetes Service. We could theoretically have a service for efk-curator and efk-clusters for example

QuentinBisson: For now, renaming workload name should be sufficient but you might want different SLOs per service inside an app for example

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
QuentinBisson: Can you document it in the rule?

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
```diff
@@ -36,6 +44,74 @@ spec:
...
+        area: managed-apps
+      record: raw_slo_requests
+      # -- 99.5% availability
+    - expr: sum by (service) (raw_slo_errors{area="managed-apps"} - raw_slo_errors{area="managed-apps"}) + 1-0.995
+      record: slo_target
+
+      # -- generic stuff
+      # -- standard burnrates based on https://sre.google/workbook/alerting-on-slos/#6-multiwindow-multi-burn-rate-alerts
+    - expr: "vector(36)"
+      record: slo_burnrate_high
```
QuentinBisson, line 112: We merged this already but should this not be named slo_burnate_fast and slow instead of high and low as this is how fast we burn our slos?

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
QuentinBisson: We need to find something else for such cases. The service name or component could be coming from an annotation for example

helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml
fiunchinho: I understand "high rate" like we are burning fast, so to me it sounds similar.

QuentinBisson: I can accept this :D I'm just checking :)

PR #595, opened 2021-06-11
Lower kubelet SLO target to 99%

kopiczko:
We thought about it in the SLO hive and we think that single kubelet being down on 100 nodes cluster at all times should not be a problem. That should translate well to other cluster sizes. If anything we should probably lower the SLO for smaller clusters.


calvix:
is the difference significant in most of cluster sizes?

kopiczko:
> is the difference significant in most of cluster sizes?

You have 10x more time to burn trough the budget.

calvix (approved):

anvddriesch (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Changed
 
+- Lower `kubelet` SLO 99%.
```
marians, line 17:
```suggestion
- Lower `kubelet` SLO from 99.9% to 99%.
```

PR #596, opened 2021-06-11
Add ServiceLevelBurnRateTicket alert

kopiczko:
The idea behind that is that we still don't have the alert that creates a ticket. Why not having an alert that suggests opening one during the office hours?

I deployed it to ginger to check if it deploys. It does. I checked handful of clusters, but I couldn't find a hit.


QuentinBisson:
I do think this makes sense but I will let the other atlassians give it a look

TheoBrigitte (approved):
I trust you on the math here. I like the poor's man ticket alerting style ;)

kopiczko:
The math is blindly taken from the SRE book.

PR #604, opened 2021-06-14
Add term to not count api-server errors for transitioning clusters

anvddriesch:
Towards https://github.com/giantswarm/giantswarm/issues/17638

The goal here is to not count api-server errors while a cluster is transitioning so that we do not eat up the error budget and get false positives once the transition is complete.
Note that this will not make a difference for workload-clusters until the cluster-status metrics are fixed by atlas.

## Checklist

I have:


njuettner (approved):
lgtm if that works 🙂

TheoBrigitte (commented):
Why no using inhibitions instead of complex expression for this ?

JosephSalisbury:
What about for HA masters? It feels weird to not have them alert if there's issues during upgrade

anvddriesch:
Right now we use the inhibition during upgrades but it feels wrong. We are still counting the errors and it still counts towards the error budget even though it is an expected downtime.
I have not considered the HA Masters clusters separately since we currently don't (inhibition works for everything). Maybe I can come up with something.

calvix (approved):
that looks fancy , never seen ignoring() before

anvddriesch:
While this is not perfect I want to try it out for now and iterate once the inhibition issue is resolved

PR #605, opened 2021-06-14
Add Alertmanager TLS ingress

QuentinBisson:
## Checklist

I have:

### DIFF
--- a/flag/service/alertmanager/alertmanager.go
+++ b/flag/service/alertmanager/alertmanager.go
@@ -1,11 +1,10 @@
 package alertmanager
 
 type Alertmanager struct {
-	Address    string
-	BaseDomain string
-	LogLevel   string
-	Storage    AlertmanagerStorage
-	Version    string
+	Address  string
+	LogLevel string
+	Storage  AlertmanagerStorage
+	Version  string
 }
 
 type AlertmanagerStorage struct {
--- a/helm/prometheus-meta-operator/templates/ingress-oauth.yaml
+++ b/helm/prometheus-meta-operator/templates/ingress-oauth.yaml
@@ -18,7 +18,17 @@ spec:
           serviceName: oauth2-proxy
           servicePort: 4180
         path: /oauth2
+  - host: {{ .Values.Installation.V1.Monitoring.Alertmanager.Host }}
+    http:
+      paths:
+      - backend:
+          serviceName: oauth2-proxy
+          servicePort: 4180
+        path: /oauth2
   tls:
   - hosts:
     - {{ .Values.Installation.V1.Monitoring.Prometheus.Host }}
     secretName: prometheus-tls
+  - hosts:
+    - {{ .Values.Installation.V1.Monitoring.Alertmanager.Host }}
+    secretName: alertmanager-tls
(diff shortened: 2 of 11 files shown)


TheoBrigitte (approved):

helm/prometheus-meta-operator/templates/ingress-oauth.yaml
```diff
@@ -18,7 +18,17 @@ spec:
           serviceName: oauth2-proxy
           servicePort: 4180
         path: /oauth2
+  - host: {{ .Values.Installation.V1.Monitoring.Alertmanager.Host }}
+    http:
+      paths:
+      - backend:
+          serviceName: oauth2-proxy
+          servicePort: 4180
+        path: /oauth2
```
mig4, line 27: Just out of curiosity, why add this here instead of keeping everything AlertManager related in `alertmanager-ingress.yaml`?

flag/service/alertmanager/alertmanager.go
```diff
@@ -1,11 +1,10 @@
 package alertmanager
 
 type Alertmanager struct {
-	Address    string
-	BaseDomain string
```
mig4, line 5: What is this `BaseDomain` and why is it being removed? I don't quite follow how this relates to the ingress changes?

QuentinBisson: So I added it earlier because I was creating the ingress programmatically as they were in the installation-prometheus namespace but now as alertmanager is moved to the monitoring one, we can do without this so I removed it here as it was originally for this PR. Do you want me to split it out?

helm/prometheus-meta-operator/templates/ingress-oauth.yaml
QuentinBisson: Because they do not share the same annotations :)

flag/service/alertmanager/alertmanager.go
mig4: Ah OK, I get it. No that's OK, with your explanation it makes sense.

QuentinBisson: It's also my bad for not explaining :(

helm/prometheus-meta-operator/templates/ingress-oauth.yaml
mig4: Well, but if this is essentially moving from https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus/templates/alertmanager/ingress.yaml then it does have the `tls-acme` annotation.

I don't know, I'll leave this up to you to decide, the reason I asked is that the `ingress-oauth.yaml` already conflates settings for Prometheuses and Promxy just because they happen to share a domain. It breaks the single-responsibility principle and means that e.g. Promxy cannot be deployed without PMO.

But since we're moving AlertManager into PMO anyway, then it doesn't matter anyway.

mig4 (approved):

QuentinBisson:
Sadly, I'm not sure we can do things differently as we also have this in g8s-prometheus for oauth:
https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus/templates/alertmanager/ingress-oauth2-proxy.yaml
Maybe we just need to remove promxy from the url? How about we change that to monitoring.baseDomain in the future? It would remove friction with prometheus even more

PR #606, opened 2021-06-14
Change severity of ChartOperatorDown alert to notify

rossf7:
Fixes https://github.com/giantswarm/giantswarm/issues/17610

This alert paging is usually a symptom of the workload cluster API server being down rather than anything actionable we can fix. Most recent example is from clusters being upgraded to HA.

Discussed with @tomahawk28 and we both think it's better to downgrade to notify.

## Checklist

I have:


webwurst (approved):

tomahawk28 (approved):
LGTM!

PR #608, opened 2021-06-14
Add heartbeat routing for all clusters

QuentinBisson:
This PRs adds the route and receiver for each prometheus. After this, we can remove our dependencies on alertmanager and prometheus code as we will not need to update g8s-prometheus alermanager config

## Checklist

I have:


service/controller/resource/alerting/alertmanager/resource.go
```diff
@@ -173,6 +173,9 @@ func toAlertmanager(v interface{}, config Config) (metav1.Object, error) {
 					},
 				},
 			},
+			AlertmanagerConfigNamespaceSelector: &metav1.LabelSelector{
+				MatchLabels: key.NamespaceLabels(),
```
QuentinBisson, line 177: We match the alertmanagerconfigs in namespaces managed by prometheus-meta-operator

service/controller/resource/alerting/alertmanagerconfigsecret/resource.go
```diff
@@ -111,7 +111,6 @@ func toSecret(v interface{}, config Config) (*corev1.Secret, error) {
 		Data: map[string][]byte{
 			key.AlertmanagerKey():         alertmanagerConfig,
 			"notification-templates.tmpl": notificationTemplate,
-			"opsgenie.key":                []byte(config.OpsgenieKey),
```
QuentinBisson, line 114: It needs to be in the monitornig namespace now so it's in it's own secret

service/controller/resource/alerting/alertmanagerrouting/resource.go
```diff
@@ -0,0 +1,157 @@
...
+							Key: "opsgenie.key",
+						},
+					},
+				},
+				SendResolved: &sendResolved,
+			},
+		},
+	}
+
+	alertmanagerConfig := &monitoringv1alpha1.AlertmanagerConfig{
```
QuentinBisson, line 128: With this, we can remove our dependency on alertmanager code

service/controller/resource/namespace/resource.go
```diff
@@ -49,7 +49,8 @@ func getObjectMeta(v interface{}) (metav1.ObjectMeta, error) {
 	}
 
 	return metav1.ObjectMeta{
-		Name: key.Namespace(cluster),
+		Name:   key.Namespace(cluster),
+		Labels: key.NamespaceLabels(),
```
mig4, line 53: Nice one, I like it.
Did you test upgrade? I assume adding labels should be fine, but I remember some weirdness when changing namespace resources, so just double-checking.

service/controller/resource/alerting/alertmanagerrouting/resource.go
mig4: Hell yeah 🎉 🚀

service/controller/resource/alerting/alertmanagerrouting/resource.go
```diff
@@ -0,0 +1,157 @@
...
+
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	labels := make(map[string]string)
+	for k, v := range key.AlertmanagerLabels(cluster) {
+		labels[k] = v
+	}
```
mig4, line 91: Why this? And not just using return value of `key.AlertmanagerLabels(cluster)` directly?

mig4 (changes requested):
Love it ❤️ 
Just one minor tweak in the `alertmanagerrouting` resource.

service/controller/resource/alerting/alertmanagerrouting/resource.go
QuentinBisson: It's not even used :D 
This is some leftover code

service/controller/resource/generic/create.go
```diff
@@ -46,8 +46,6 @@ func updateMeta(c, d metav1.Object) {
 	d.SetCreationTimestamp(c.GetCreationTimestamp())
 	d.SetDeletionTimestamp(c.GetDeletionTimestamp())
 	d.SetDeletionGracePeriodSeconds(c.GetDeletionGracePeriodSeconds())
-	d.SetLabels(c.GetLabels())
-	d.SetAnnotations(c.GetAnnotations())
```
mig4, line 50: Was this intended?

mig4 (changes requested):
Looking good.
I see you're still working on this, let me know when you're done testing and I'll take another look.

service/controller/resource/generic/create.go
QuentinBisson: I'm on it as the namespace change is making some of it not work anymore :D
Yes it is, otherwise, we will never be able to update labels and annotations :( 
I am thinking we can reenable it in a subsequent release

PR #614, opened 2021-06-16
Add metrics-server deployment to monitored WC components

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/17712

## Checklist

I have:

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/deployment.workload-cluster.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/deployment.workload-cluster.rules.yml
@@ -22,6 +22,17 @@ spec:
         severity: page
         team: batman
         topic: releng
+    - alert: WorkloadClusterDeploymentNotSatisfiedLudacris
+      annotations:
+        description: '{{`Deployment {{ $labels.namespace}}/{{ $labels.deployment }} is not satisfied.`}}'
+        opsrecipe: workload-cluster-deployment-not-satisfied/
+      expr: kube_deployment_status_replicas_unavailable{cluster_type="workload_cluster", deployment="metrics-server"} > 0
+      for: 30m
+      labels:
+        area: kaas
+        severity: page
+        team: ludacris
+        topic: observability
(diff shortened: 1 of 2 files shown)


helm/prometheus-meta-operator/templates/prometheus-rules/deployment.workload-cluster.rules.yml
```diff
@@ -22,6 +22,17 @@ spec:
         severity: page
         team: batman
         topic: releng
+    - alert: WorkloadClusterDeploymentNotSatisfiedLudacris
```
MarcelMue, line 25: Can we maybe somehow make this an SLO thingy?

stone-z: Maybe? It is a single pod in the deployment, this would fire if it is down for 30min. I guess the upside is SLO would also catch a flappy metrics server?

MarcelMue: I think using this PR as a reference should be pretty straight forward:
https://github.com/giantswarm/prometheus-meta-operator/pull/588/files

LMK if we can pair on it or something

stone-z: So after looking into it a bit more, I think the better approach would be to create some class of our components which we want to be SLO alerts (like [managed apps](https://github.com/giantswarm/prometheus-meta-operator/blob/master/helm/prometheus-meta-operator/templates/recording-rules/service-levels.rules.yml)). There doesn't seem to be a clean way to do it for a single component at this point (because IIUC alert routing is still not done by the usual team label but based on the component when the alert reaches opsgenie).
So I'd suggest we go with this for now so we've at least got it in place and then move our alerts to SLO as part of a bigger conversation about which and how.

MarcelMue (approved):
Fine to do this now and follow up later with a SLO conversion thingy

stone-z:
Tried this out on `gauss` by repeatedly killing the WC metrics-server. Alert is pending as expected

PR #615, opened 2021-06-16
Add `PrometheusHighMemoryUsage` alert

mig4:
To page us during working hours if Prometheus container memory usage
(specifically the working set size) grows beyond 95% of total memory on
the node it's running on. This is dangerously close to OOM crash-loop
territory and likely needs to be addressed by either cleaning up some
data or upgrading to nodes with more memory.

Closes https://github.com/giantswarm/giantswarm/issues/17350

## Checklist

I have:

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
@@ -34,3 +34,33 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: PrometheusHighMemoryUsage
+      annotations:
+        description: |-
+          {{`Prometheus in pod {{ $labels.namespace }}/{{ $labels.pod }} is using >95% of node's memory.
+          VALUE = {{ $value }}
+          LABALE = {{ $labels }}`}}
+        opsrecipe: TODO/
+      # `node` label is sometimes a name and sometimes IP, and it was different
+      # between those metrics on kvm installations, so extract IP to be able to match
+      expr: |
+        sum by (node_ip, pod) (
+          label_replace(
+            container_memory_working_set_bytes{pod=~"^prometheus-.+-\\d+$", container="prometheus"},
+            "node_ip", "$1", "instance", "(.*):.*"
+          )
+        )
+        / on (node_ip) group_left
+        label_replace(
+          node_memory_MemTotal_bytes,
+          "node_ip", "$1", "instance", "(.*):.*"
+        )
+        * 100
+        > 95
+      for: 5m
+      labels:
+        area: empowerment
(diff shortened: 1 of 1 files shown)


mig4:
Still need to add a changelog, but let me know what you think about the approach.

I tried looking into alerting based on recent restarts but for some reason `process_start_time_seconds` is not available for Prometheuses. Still, alerting on the actual symptom that happens just before restart loops might be better?

helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
```diff
@@ -34,3 +34,33 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: PrometheusHighMemoryUsage
+      annotations:
+        description: |-
+          {{`Prometheus in pod {{ $labels.namespace }}/{{ $labels.pod }} is using >95% of node's memory.
+          VALUE = {{ $value }}
```
QuentinBisson, line 41: I would imagine VALUE and LABALE should be removed here?

helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
```diff
@@ -34,3 +34,33 @@ spec:
...
+          LABALE = {{ $labels }}`}}
+        opsrecipe: TODO/
```
QuentinBisson, line 43: Could you redirect to a page that will exist in the future?

helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
```diff
@@ -34,3 +34,33 @@ spec:
...
+      # `node` label is sometimes a name and sometimes IP, and it was different
```
QuentinBisson, line 44: I thought we fixed it with the instance label?

helm/prometheus-meta-operator/templates/prometheus-rules/prometheus.rules.yml
```diff
@@ -34,3 +34,33 @@ spec:
...
+      # between those metrics on kvm installations, so extract IP to be able to match
+      expr: |
+        sum by (node_ip, pod) (
```
QuentinBisson, line 47: The rule makes sense to me :)

QuentinBisson: I'm wondering if we should not alert only if the memory is increasing after 95 % instead of when it's above.

mig4:
Closing this as per https://github.com/giantswarm/giantswarm/issues/17350#issuecomment-863333218

PR #621, opened 2021-06-17
Moves SLO alert to opt-out

JosephSalisbury:
Closes https://github.com/giantswarm/giantswarm/issues/17734

To avoid having services alerting that we don't - e.g: in the case where we're still testing a service's service level out, this adds support to opt-in to the service level alert, so we don't get unwanted pages.

See https://gigantic.slack.com/archives/C02MS174K/p1623828191186600 for more context.

## Checklist

I have:

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/service-level.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/service-level.rules.yml
@@ -14,7 +14,7 @@ spec:
         description: '{{`Service level burn rate is too high for {{ $labels.service }} service.`}}'
         opsrecipe: service-level-burn-rate-too-high/
       expr: |
-        slo_errors_per_request:ratio_rate1h{service=~"api-server|kubelet"}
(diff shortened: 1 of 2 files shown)


rossf7:
@JosephSalisbury I added Halo too.

helm/prometheus-meta-operator/templates/prometheus-rules/service-level.rules.yml
```diff
@@ -14,13 +14,17 @@ spec:
         description: '{{`Service level burn rate is too high for {{ $labels.service }} service.`}}'
         opsrecipe: service-level-burn-rate-too-high/
       expr: |
+        slo_errors_per_request:ratio_rate1h{service=~"api-server|kubelet"}
```
QuentinBisson, line 17: I understand why we do this but I'm not sure I like the idea because this will be a nightmare. Would it not make more sense to opt out by area and/or service instead ?

QuentinBisson: I think it should be easier to opt-in that to opt-out

rossf7 (approved):

TheoBrigitte (approved):

PR #627, opened 2021-06-21
Polish CoreDNSMaxHPAReplicasReached alert

pipo02mix:
When max and min number of replicas are the same don't alert. After deploying the new alert we have been paged for clusters where they fix number of max/min replicas. I don't see the point of alert in such a case

## Checklist

I have:


stone-z (approved):
Yeah that makes sense

MarcelMue (approved):

PR #631, opened 2021-06-21
Added/fixed azure WC creation and update failure alerts

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/17763

This PR adds a new alert for azure workload cluster which upgrade takes too long and fixes the alert for creation taking too long.

## Checklist

I have:

### DIFF
--- a/helm/prometheus-meta-operator/templates/prometheus-rules/azure.management-cluster.rules.yml
+++ b/helm/prometheus-meta-operator/templates/prometheus-rules/azure.management-cluster.rules.yml
@@ -23,9 +23,9 @@ spec:
         severity: notify
         team: celestial
         topic: azure
-    - alert: AzureClusterUpdateFailed
+    - alert: AzureClusterUpgradeFailed
       annotations:
-        description: '{{`Cluster {{ $labels.cluster_id }} update is taking longer than expected.`}}'
+        description: '{{`Cluster {{ $labels.cluster_id }} upgrade is taking longer than expected.`}}'
         opsrecipe: cluster-update-failed/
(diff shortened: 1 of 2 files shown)


helm/prometheus-meta-operator/templates/prometheus-rules/azure.management-cluster.rules.yml
```diff
@@ -15,16 +15,26 @@ spec:
...
-      expr: (statusresource_cluster_status{app=~"azure-operator.*", status="Creating"} == 1 or cluster_operator_cluster_status{app=~"cluster-operator.*", status="Updating", provider="azure"} == 1)
+      expr: azure_operator_cluster_status(status="Creating"} == 1
       for: 40m
       labels:
         area: kaas
         cancel_if_outside_working_hours: "true"
         severity: notify
         team: celestial
         topic: azure
+    - alert: AzureClusterUpdateFailed
```
tuommaki, line 26: Would it make sense to unify the terminology and `s/update/upgrade/` ?

We usually talk about upgrades when the cluster version is bumped up. Updating cluster gives me an association to cluster configuration change, such as node count change, node pool creation / deletion, node size change etc. that doesn't affect the used release.

WDYT?

MarcelMue (approved):
Generally LGTM - not very opinionated on the naming.

helm/prometheus-meta-operator/templates/prometheus-rules/azure.management-cluster.rules.yml
whites11: renamed

PR #639, opened 2021-06-22
Removes ServiceLevelBurnRateTicket alert

JosephSalisbury:
This was too unactionable to be useful

## Checklist

I have:


calvix (approved):
can't say if it was useless, but for now, it triggered a lot of false-positive in clusters that changed nodes(rolling or scaling down)

PR #654, opened 2021-06-24
Fix node role labels

QuentinBisson:
## Checklist

Closes https://github.com/giantswarm/giantswarm/issues/2485

I have:


QuentinBisson:
It seems it can only be achieved using a join

PR #664, opened 2021-07-06
Trim workload-cluster KSM metrics

mig4:
Change the scrape configuration to only preserve metrics from
`kube-system` and `giantswarm` namespaces on workload clusters to cut
down on a massive amount of metrics coming from _kube-state-metrics_
on clusters with a lot of activity.

Closes: https://github.com/giantswarm/giantswarm/issues/17980

## Checklist

I have:


mig4:
Tested on _alpaca/cna01_.

### Before

metrics by app (`sort_desc(count({__name__=~".+"})by(app))`):

before metrics-by-app 2021-07-06--13-55-07

metrics by namespace (`sort_desc(count({__name__=~".+"})by(namespace))`):

before metrics-by-namespace 2021-07-06--14-08-18

### After

metrics by app:

after metrics-by-app 2021-07-06--15-02-27

metrics by namespace

after metrics-by-namespace 2021-07-06--15-04-04

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -655,6 +655,10 @@
   - source_labels: [__name__]
     regex: (nginx_ingress_controller_bytes_sent_bucket|nginx_ingress_controller_request_duration_seconds_bucket|nginx_ingress_controller_request_size_bucket|nginx_ingress_controller_response_duration_seconds_bucket|nginx_ingress_controller_response_size_bucket)
     action: drop
+  # on workload clusters only keep metrics from namespaces we manage
+  - source_labels: [namespace]
+    regex: (kube-system|giantswarm.*)
+    action: keep
```
TheoBrigitte, line 661: Wait this apply to every cluster, could we re-use the condition we saw ... somewhere ... ?

mig4: It's in the workload job, so I assumed it's only WCs? Or do you mean you'd want to limit it further?

mig4: But I think you're right, I don't actually see anything that would limit this to WCs only.

TheoBrigitte: That's a MC job too. I'm not sure we should limit this on MCs.

mig4: Yeah, you're right. OK leave it with me, I'll play around with the conditionals and test this a bit more. Thanks for checking, and well spotted!

mig4: Confirmed on _gorgoth_. Before this change:

prometheus-mc workload-job-apps-from-other-namespaces before-change 2021-07-13--16-49-29

After:

prometheus-mc workload-job-apps-from-other-namespaces after-change-1 2021-07-13--16-57-52

mig4: OK with commit 9470f0700a636a85abe14e8706ed3d00f6baa7b4 we're back to:

prometheus-mc workload-job-apps-from-other-namespaces after-change-2 2021-07-14--09-49-42

JosephSalisbury:
does this mean we have targets in other namespaces, but we don't care about their metrics? what are they? why are we scraping them?

QuentinBisson:
By default, we get ksm metrics for all namespaces, including customers. The issue if we restrict the scope to giantswarm and kube-system is that we will not be collecting the KSM metrics for customer installed apps as they are in different namespaces :(

JosephSalisbury:
Can we do something with labels? Don't all the KSMs we install for customers have a certain label?

My general thinking is to not scrape unnecessary targets at all, rather than drop the metrics later.

QuentinBisson:
I think the managed app have the service-type: managed label but we would need halo to confirm. Also, it could make sense to have it aligned with the giantswarm.io/service-type pattern ?

mig4:
> does this mean we have targets in other namespaces, but we don't care about their metrics? what are they? why are we scraping them?

The `workload` target doesn't currently filter by namespace. You can get an idea of the metrics we're getting from the before screenshots, e.g. on that CI cluster the biggest offender is Harbor - 213287 metrics. There's also a ton of metrics from KSM regarding creating/deleting ConfigMaps and other resources.

mig4:
> My general thinking is to not scrape unnecessary targets at all, rather than drop the metrics later.

I'm not sure how to do that, should I move this filter from `metric_relabel_configs` to `relabel_configs`?

mig4:
Testing on _gremlin_ as there's a WC there with linkerd and some test workloads.

change-2 step-0 before prometheus-wc gremlin-x2oo6 workload-job-metrics-by-namespace 2021-07-22--15-38-21

But after moving the filter from `metric_relabel_configs` to `relabel_configs` I'm getting no metrics at all:

change-2 step-1 after prometheus-wc gremlin-x2oo6 workload-job-metrics-by-namespace 2021-07-22--16-24-39

Confirmed the same thing on _alpaca_.

JosephSalisbury:
From that above list, would we still want metrics from the `linkerd2-app` namespace? (as one example). I think that's an example of a managed app we deploy, which we'd still want metrics for.

QuentinBisson:
Yes, otherwise, you will break all the managed apps alerts

mig4:
OK so this isn't going to work, closing it as explained in the [comment on the original ticket](https://github.com/giantswarm/giantswarm/issues/17980#issuecomment-885575652).

PR #675, opened 2021-07-30
Update prom to 2.28.1

QuentinBisson:
Upgrading prometheus to the latest version

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Add the service label in the alert templates for the `ServiceLevelBurnRateTooHigh` alert.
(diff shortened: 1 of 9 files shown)


JosephSalisbury (approved):
lgtmiiwfy - do we need to coordinate this w/ prometheus-operator version?

mig4 (approved):
Ditto.

mig4:
> do we need to coordinate this w/ prometheus-operator version?

It seems there's usually enough compatibility there that it works OK without it, but yeah it's not a supported configuration, so it would be good to update _prometheus-operator_ to 0.49.0 (see https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/operator/defaults.go).

QuentinBisson:
I am planning on testing it first anyway :) But we can ask Halo to update the app

PR #687, opened 2021-08-09
Add kyverno as scrape endpoint

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/18354

Scrapes kyverno so we can get better metrics about it

## Checklist

I have:


stone-z:
My understanding of prometheus's intended behavior with `endpoints` is that it will discover all ports on the backing pod. This means it also finds port `9443` on the kyverno pods, which don't serve metrics and are never `up`. 

image


I'm not sure how to tell prometheus to ignore those altogether, would appreciate any hints. I don't think it's the end of the world, but probably inefficient to have them in there so I'd like to remove them if possible.

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -399,15 +420,9 @@
...
     target_label: instance
-[[ if eq .ClusterType "management_cluster" ]]
-  - source_labels: [__meta_kubernetes_service_name]
-    regex: chart-operator-unique
-    action: keep
-[[ else ]]
   - source_labels: [__meta_kubernetes_service_name]
     regex: chart-operator
     action: keep
-[[ end ]]
```
stone-z, line 410: The service for `chart-operator` should no longer use `-unique` on MCs.

stone-z:
Would appreciate a look from @giantswarm/team-atlas -- is there a way to tell prometheus to completely ignore one port on one of the pods it is scraping and not even see it as a target? Or does it not actually matter?

MarcelMue (approved):
LGTM

QuentinBisson:
It looks good to me but before I approve, what do you think about using a ServiceMonitor in the Kyverno chart instead ?

stone-z:
@QuentinBisson I'm generally open to it, but I tried using the labels/annotations as well as a ServiceMonitor and that seemed to not work. Will try it again today and go with this if it still doesn't work

stone-z:
Well, I've learned a lot about our monitoring setup from this process :smile: https://github.com/giantswarm/kyverno-app/pull/51

stone-z:
Closing in favor of https://github.com/giantswarm/kyverno-app/pull/51

PR #688, opened 2021-08-10
bump alertmanager to 0.22.2

TheoBrigitte:
We are now using alertmanager 0.22.2, so PMO needs to use the same version internally to avoid any problem.

This PR was created by following the procedure written here https://github.com/giantswarm/prometheus-meta-operator/blob/master/README.md#how-to-update-upstream-code

## Checklist

I have:


go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/giantswarm/micrologger v0.5.0
 	github.com/giantswarm/operatorkit/v4 v4.3.1
 	github.com/giantswarm/versionbundle v0.2.0
-	github.com/go-kit/log v0.1.0
+	github.com/go-kit/kit v0.10.0
```
mig4, line 17: So this is needed after all?

TheoBrigitte: that's required by alertmanager/config: https://github.com/prometheus/alertmanager/blob/44f8adc06af5101ad64bd8b9c8b18273f2922051/config/coordinator.go#L21

mig4: Right, got it.

TheoBrigitte:
This was tested successfully on ginger.

PR #705, opened 2021-09-01
drop v1alpha2

njuettner:
AWS only uses `v1alpha2` till now but we switched to `v1alpha3` as stored version.
CRD Webhook converts all `v1alpha2` CR's to `v1alpha3`.

Dropping `v1alpha2` is needed for migration.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Drop `v1alpha2` scheme.
+
 ## [1.50.0] - 2021-08-16
 
 ### Changed
--- a/pkg/unittest/input/case-0-cluster-api.golden
+++ b/pkg/unittest/input/case-0-cluster-api.golden
@@ -1,4 +1,0 @@
-apiVersion: cluster.x-k8s.io/v1alpha2
-kind: Cluster
-metadata:
-  name: bob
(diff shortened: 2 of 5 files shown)


TheoBrigitte:
This looks alright. There might be some test results to remove here and there. But my main point would be to be sure this is tested.

mig4:
Hey, thanks, this was the next thing we were going to look into 😁 Thanks for kicking this off.

So, AFAIK the webhook won't convert CRDs and since we have to tell operatorkit which version to watch, i.e.:

``` go
fn = func() runtime.Object { return new(capiv1alpha2.Cluster) }
```

the webhook won't work for that.

Are there any MCs that still contain any v1alpha2 `Cluster` CRs?

mig4:
> Are there any MCs that still contain any v1alpha2 `Cluster` CRs?

I just checked `opsctl kubectl -i $all_installations -- --all-namespaces=true get clusters.cluster.x-k8s.io -o custom-columns=NAMESPACE:.metadata.namespace,NAME:.metadata.name,APIVERSION:.apiVersion` and I can't see any, so :+1: should be good to go. Nice!

mig4 (approved):

njuettner:
Sorry for catching up, I was AFK yesterday.

Yeah the CRD conversion webhook is in place on all AWS installations for v1alpha2 <-> v1alpha3 conversion. v1alpha3 is kept as the storage version. So any call without specifically apiVersion is returning v1alpha3.

I saw that the `prometheus-meta-operator` is still watching on `v1alpha2`CRs based on the information from managedFields.

I'll deploy it manually in gauss and will double check. Thank you for having a look on that PR 🤗

njuettner:
updates on `v1alpha3`

```  
- apiVersion: cluster.x-k8s.io/v1alpha3
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .: {}
          v:"operatorkit.giantswarm.io/prometheus-meta-operator-cluster-api-controller": {}
    manager: prometheus-meta-operator
    operation: Update
    time: "2021-09-03T12:47:43Z"
```

PR #711, opened 2021-09-07
reconcile v1alpha4 cluster

TheoBrigitte:
Changed the logic to only reconcile v1alpha3 cluster, as the webhook conversion will do the conversion job for us.

The main change is that v1alpha3 is always used whether or not its the storage crd version. 

This solves all cases where AWS run v1alpha2 and v1alpha3 (as storage) and Azure runs v1alpha3 and v1alpha4 (as storage)

```
$ k --context giantswarm-ginger get crd clusters.cluster.x-k8s.io -oyaml|yq -y '.spec.versions[]|{name, storage}'
name: v1alpha2
storage: false
---
name: v1alpha3
storage: true
 k --context giantswarm-ghost get crd clusters.cluster.x-k8s.io -oyaml|yq -y '.spec.versions[]|{name, storage}'
name: v1alpha3
storage: false
---
name: v1alpha4
storage: true
```


TheoBrigitte:
Tested on both aws/ginger and azure/ghost. Cluster can be created and deleted. :heavy_check_mark:

QuentinBisson:
Can we not reconcile v1alph4 instead of v1alpha3 ?

TheoBrigitte:
> Can we not reconcile v1alph4 instead of v1alpha3 ?

I don't think we should. How does this work on aws where only v1alpha3 and v1alpha2 are available?

QuentinBisson:
Is v1alpha2 still available though ?

TheoBrigitte:
Yes and v1alpha4 is not yet on aws installations AFAIK.

QuentinBisson:
Good to know :D

QuentinBisson (approved):

PR #736, opened 2021-10-05
Remove oauth2-proxy ingress

QuentinBisson:
Remove the oauth2-proxy ingress as it is moved to https://github.com/giantswarm/config/pull/513

## Checklist

I have:


TheoBrigitte (changes requested):
promxy oauth authentication won't work without this.

I think we got mistaken here this ingress should be moved to promxy-app and not oauth2-proxy I believe.

QuentinBisson:
The idea was for this ingress to be installed with oauth2-proxy directly so we would have the ingress but it's not currently doable due to the single host in oauth2-proxy

PR #762, opened 2021-11-30
Avoid defaulting node role

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/18779

Among all exporters that are discovered using the prometheus' `kubernetes` discovery plugin, only the ones using the `node` role have the node's labels available.
We rely on the nodes' "role" label to detect the role of the k8s node so we can't reliably detect this information.

This PR removes the defaulting feature to avoid the wrong default.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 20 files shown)


QuentinBisson (approved):
LGTM but why would the label be missing ?

whites11:
> LGTM but why would the label be missing ?

I tried to explain in the PR description. The `Node` resource labels are only copied to the metrics when the `role` used to discover prometheus targets is `node`. So all the pods that use `endpoint` will never have such label

PR #763, opened 2021-11-30
Inhibition no workers

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/546

This PR enables a new inhibition leveraging https://github.com/giantswarm/prometheus-rules/blob/0b1f77f8b1b5a840ad47ff7bb0599b929c737283/helm/prometheus-rules/templates/alerting-rules/inhibit.all.rules.yml#L32 and silencing any alerts labeled `cancel_if_has_no_workers` when the cluster has no worker nodes

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add new inhibition for clusters without workers.
+
 ### Changed
 
 - Upgrade alertmanager to v0.23.0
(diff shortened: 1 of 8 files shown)


fiunchinho (approved):
I don't know how this works

whites11:
> I don't know how this works

This is a feature of alert manager. You can say "I want you to silence alerts with labels as defined in `target_match` IF there is another alert with labels as defined in `source_match` AND the two alerts have matching label values for all labels defined in `equals`. This is the glue between the inhibition alert we emit when there are no workers (on the prometheus side) and alertmanager

fiunchinho:
> > I don't know how this works
> 
> This is a feature of alert manager. You can say "I want you to silence alerts with labels as defined in `target_match` IF there is another alert with labels as defined in `source_match` AND the two alerts have matching label values for all labels defined in `equals`". This is the glue between the inhibition alert we emit when there are no workers (on the prometheus side) and alertmanager

That makes sense, thank you.

PR #765, opened 2021-12-07
Adding  support for updating app label based on monitoring annotation

ced0ps:
## Change description
- Towards https://github.com/giantswarm/giantswarm/issues/20083
- Adds relabel config for app that copies app name annotation to prometheus app label, towards helping managed app monitoring

## Checklist

I have:

- [x] tested on gauss

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add new inhibition for clusters without workers.
-- Relabeling for `__meta_kubernetes_annotation_giantswarm_io_monitoring_app_label` 
+- Add relabeling for `__meta_kubernetes_annotation_giantswarm_io_monitoring_app_label`
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add new inhibition for clusters without workers.
+- Relabeling for `__meta_kubernetes_annotation_giantswarm_io_monitoring_app_label` 
 
 ### Changed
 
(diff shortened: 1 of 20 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add new inhibition for clusters without workers.
+- Relabeling for `__meta_kubernetes_annotation_giantswarm_io_monitoring_app_label`
```
ced0ps, line 13:
```suggestion
- Add relabeling for `__meta_kubernetes_annotation_giantswarm_io_monitoring_app_label`
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add new inhibition for clusters without workers.
+- Add relabeling for `__meta_kubernetes_annotation_giantswarm_io_monitoring_app_label`
```
QuentinBisson, line 14: Could you explain what is wrong with the current app label ?

ced0ps: The current app label gets the name of the service that points to the metrics endpoint.
With kong as an example you get something like `kong-app-kong-app-service-metrics` as the name.
If the chart is built in a way that let's you change the name of a charts components, e.g. `fullNameOverride` or similar you can end up with an app like `big-dumb-ape-big-dumb-ape-metrics`.
We tried to find a label or annotation that holds a name that can reliably be linked to an app. The closest would be the label `helm.sh/chart=kong-app-2.3.4` which is used in most upstream helm charts we checked. But this isn't as clean, since we would need to fuzzy match because of the version #.
Since we do have to add other monitoring labels to the metrics service, this is the most reliable solution we could come up with.

QuentinBisson: Alright, in that case, why not call is app_name instead of app_label ?

JosephSalisbury: we went with a _lot_ of bikeshedding already, went with 'label' over 'name' because we reckoned it makes it clearer that this is explicitly for overriding the app label in prometheus

QuentinBisson: ack

QuentinBisson (approved):

PR #803, opened 2022-01-14
Set ingress class name

QuentinBisson:
## Checklist

I have:


helm/prometheus-meta-operator/templates/ingress-oauth.yaml
```diff
@@ -1,23 +1,43 @@
...
+            port:
+              number: 9090
+          {{- else }}
           serviceName: oauth2-proxy
           servicePort: 4180
+          {{- end }}
         path: /oauth2
+        {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
+        pathType: ImplementationSpecific
+        {{- end }}
```
TheoBrigitte, line 40: Why do we need a condition for this?

service/controller/resource/monitoring/ingress/resource.go
```diff
@@ -114,6 +118,7 @@ func toIngress(v interface{}, config Config) (metav1.Object, error) {
 		},
 		ObjectMeta: objectMeta,
 		Spec: networkingv1beta1.IngressSpec{
+			IngressClassName: &ingressClassName,
```
TheoBrigitte, line 121: How does this works on 1.18 k8s?

QuentinBisson: I'm trying to deploy it on 1.20 first :D

QuentinBisson: Yes this is not working

QuentinBisson: I wanted to see if we could use chartutil to get the capabilities but it looks like we cannot support v1beta1 and v1 easily enough :(

TheoBrigitte: We should know the kubernetes version we are working against, but then we need to maintain both
`v1beta1.Ingress` and `v1.Ingress` objects in Go also :face_exhaling:

helm/prometheus-meta-operator/templates/ingress-oauth.yaml
TheoBrigitte: I think not

service/controller/resource/monitoring/ingress/resource.go
QuentinBisson: I think maintaining the 2 would not be hard tbh but that means we need to move out of the generic resource and tests :( I'm not sure I want to go that way today :D

TheoBrigitte: Maybe we can find faith together tomorrow and pair on this

QuentinBisson: Tested on gorgoth with 1.18

QuentinBisson:
Tested on Gorgoth, gauss, ginger and gremlin

TheoBrigitte (approved):
Nice

QuentinBisson:
Thanks buddy :)

PR #820, opened 2022-02-10
Scrape node-exporter trough apiserver proxy

kopiczko:
Towards https://github.com/giantswarm/roadmap/issues/725

### DIFF
--- a/files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
+++ b/files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
@@ -502,7 +502,7 @@
 # node-exporter
 - job_name: [[ .ClusterID ]]-prometheus/node-exporter-[[ .ClusterID ]]/0
   honor_labels: true
-  scheme: http
+  scheme: https
   kubernetes_sd_configs:
   - role: pod
 [[ include "_apiserver" . ]]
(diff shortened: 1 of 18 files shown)


kopiczko:
Seems to work on openstack:

Screenshot 2022-02-10 at 19 25 12

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -500,42 +502,23 @@
...
-    replacement: $1:$2
-    action: replace
-  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scheme]
-    regex: (https?)
-    target_label: __scheme__
-    replacement: $1
-    action: replace
-  - source_labels: [__meta_kubernetes_service_annotationpresent_giantswarm_io_monitoring, __meta_kubernetes_service_labelpresent_giantswarm_io_monitoring]
-    regex: .*(true).*
-    action: drop
```
TheoBrigitte, line 538: I reckon this is kind of important otherwise we get duplicate targets from node-exporter app which is using the `giantswarm.io/monitoring` annotation.

TheoBrigitte (commented):
Looks alright, but I thought there was supposed to be something specific to openstack, where is this?

kopiczko:
> I thought there was supposed to be something specific to openstack, where is this?

That a good call. I forgot about that. I was testing only on the MC. I will add it in a single PR because I don't want to test everything twice for other providers.

kopiczko:
Tested in AWS/Azure/KVM. I'll make changes for openstack MCs later.

PR #823, opened 2022-02-23
node-exporter target: drop namespace,container, and pod label. keep node label.

TheoBrigitte:
This PR fixes an issue where duplicate series are found for node-exporter. This happen because we label the target with its pod name, which results in producing multiples series for the same node over time when pods are replaced.

The prometheus memory percentage aggregation rely on the fact that nodes are unique.

image

#### Before

Screenshot_2022-02-23_14-14-17

#### After

image


QuentinBisson (approved):
LGTM

PR #470, opened 2021-03-29
Move ingress to v1

QuentinBisson:
This needs to be tested

## Checklist

I have:

- [x] Described why this change is being introduced
- [x] Separated out refactoring/reformatting in a dedicated PR
- [x] Updated changelog in `CHANGELOG.md`

PR #826, opened 2022-03-01
Support ingress v1 by default

QuentinBisson:
## Checklist

Towards https://github.com/giantswarm/giantswarm/issues/17206#issuecomment-1054495336

I have:

I will test it tomorrow but it should solve the issue as long as we set the value for the ingress version on kvm installations


tfussell:
I guess https://github.com/giantswarm/prometheus-meta-operator/pull/470 is superseded by this PR.

service/controller/legacy/controller.go
```diff
@@ -33,6 +33,7 @@ type ControllerConfig struct {
 	Provider                string
 	Region                  string
 	Registry                string
+	IngressApiVersion       string
```
tfussell, line 36: In Go we usually capitalize abbreviations so this should be `IngressAPIVersion`. See `AlertmanagerCreatePVC` below for example.

main.go
```diff
@@ -115,6 +115,7 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Installation.Region, "", "Region where the installation is located.")
 	daemonCommand.PersistentFlags().String(f.Service.Installation.Customer, "", "Customer the installation belongs to.")
 	daemonCommand.PersistentFlags().String(f.Service.Installation.Registry, "", "Container image registry.")
+	daemonCommand.PersistentFlags().String(f.Service.Installation.IngressApiVersion, "networking.k8s.io/v1", "Installation supported ingress api version")
```
tfussell, line 118:
```suggestion
	daemonCommand.PersistentFlags().String(f.Service.Installation.IngressApiVersion, "networking.k8s.io/v1", "Installation supported ingress API version")
```

tfussell:
I guess we should go ahead and make the corresponding PR in `giantswarm/config` overriding this to `v1beta1` for all installations that don't support `v1` yet before merging this.

QuentinBisson:
Regarding config, I think this should be for all KVM installations. @kopiczko as Thomas is not online, can you confirm ?

kopiczko:
> Regarding config, I think this should be for all KVM installations. @kopiczko as Thomas is not online, can you confirm ?

Can't we use capabilities to set that?

kopiczko:
You should be able to add something like below here https://github.com/giantswarm/prometheus-meta-operator/blob/master/helm/prometheus-meta-operator/templates/configmap.yaml

```

service:
  ...
  installation:
    ...
{{- if not .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
    apiVersion: "networking.k8s.io/v1beta1"
{{- end }}
```

QuentinBisson:
@kopiczko omg thank you :)

QuentinBisson:
cc @TheoBrigitte

QuentinBisson:
Tested on CAPO 1.22 clusters.

TheoBrigitte (approved):
Sounds good to me when tested.

helm/prometheus-meta-operator/templates/configmap.yaml
```diff
@@ -80,5 +80,10 @@ data:
         region: onprem
         {{- end }}
         registry: {{ .Values.registry.domain }}
+        {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
```
JosephSalisbury, line 83: for my understanding - could we query the api server for this, instead of having it configurable?

QuentinBisson: Yes helm checks that using the kubernetes API. If we were to add that to our codebase, this would basically mean importing helm code in pmo

PR #838, opened 2022-03-08
remove aws/azure/kvm/vmware/openstack collections

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/21069

Let's only use shared app collection rather than every provider collection.


TheoBrigitte:
shared app collection do not work.

PR #840, opened 2022-03-15
do no trail right whitespaces

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/21294

Removing trailing right whitespaces result in invalid/broken config and crashlooping PMO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Migrate to rbac/v1 from rbac/v1beta1.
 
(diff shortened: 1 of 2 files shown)


TheoBrigitte:
There's another one [there ](https://github.com/giantswarm/prometheus-meta-operator/blob/b4a85f033b208db6f8f53e280955fbc5fcbc7031/helm/prometheus-meta-operator/templates/ingress-oauth.yaml#L3) but I am not sure we should remove this or not

QuentinBisson:
I think we need to keep this one

TheoBrigitte:
I think this worked so far, so let's keep it.

PR #846, opened 2022-03-18
Add missing alertmanager flags

QuentinBisson:
Adding flags in preparation for some follow up prs

## Checklist

I have:

### DIFF
--- a/main.go
+++ b/main.go
@@ -121,10 +121,11 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Opsgenie.Key, "", "Opsgenie Key used for API authentication.")
 
 	daemonCommand.PersistentFlags().String(f.Service.Alertmanager.Address, "", "Address to access Alertmanager UI.")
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -121,10 +121,11 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Opsgenie.Key, "", "Opsgenie Key used for API authentication.")
 
 	daemonCommand.PersistentFlags().String(f.Service.Alertmanager.Address, "", "Address to access Alertmanager UI.")
+	daemonCommand.PersistentFlags().String(f.Service.Alertmanager.BaseDomain, "", "Base domain to create Alertmanager Ingress resources under.")
```
TheoBrigitte, line 124: Can't we re-use prometheus base domain? I think they are the same.

QuentinBisson: We could but this means we provide less flexibility in the end

QuentinBisson:
@TheoBrigitte and @JosephSalisbury what do you think we should do here ?

PR #849, opened 2022-03-21
Fix notification template

QuentinBisson:
## Checklist

I have:


TheoBrigitte (commented):
Was this tested on both alertmanagers?
I reckon we needed that because of helm templating

QuentinBisson:
I tested it on the new one yes :)

QuentinBisson:
As this is not templated using helm it's not needed

PR #852, opened 2022-03-22
Add heartbeat webhook config

QuentinBisson:
## Checklist

I have:


service/controller/resource/alerting/heartbeatwebhookconfig/resource.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, microerror.Mask(err)
+	}
+
+	labels := make(map[string]string)
+	for k, v := range key.AlertmanagerLabels() {
+		labels[k] = v
+	}
+
+	sendResolved := false
+	urlAddress, err := url.Parse(fmt.Sprintf("https://api.opsgenie.com/v2/heartbeats/%s/ping", key.HeartbeatName(cluster, config.Installation)))
```
TheoBrigitte, line 95: Please move this into a key package function

service/key/key.go
```diff
@@ -278,3 +278,7 @@ func IsCAPICluster(obj metav1.Object) bool {
 func IngressClassName() string {
 	return "nginx"
 }
+
+func OpsgenieKey() string {
+	return "opsgenie.key"
```
TheoBrigitte, line 283: Can't we just have ?

```suggestion
	return "opsgenie"
```

TheoBrigitte (changes requested):

service/controller/resource/alerting/heartbeatwebhookconfig/resource.go
QuentinBisson: Solved

service/key/key.go
QuentinBisson: We could, I named that because the value is mounted into a secret and we know what this field is about

TheoBrigitte: It feels redundant to call a key key to me.

PR #867, opened 2022-04-04
Ensure openstack is always detected as CAPI

kopiczko:
Didn't work correctly on `gamma`. Slack: https://gigantic.slack.com/archives/C01BYMF6RN0/p1649095924028909


kopiczko:
Opening a PM instead.

PR #870, opened 2022-04-04
openstack etcd certificates

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/21156

Add support for openstack etcd-certificates, by copying the `giantswarm/etcd-certs` secret.

Requires #869 #868 #871

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add context to generic resources.
+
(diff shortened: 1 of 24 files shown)


QuentinBisson:
We do need to check how to handle the renewal of the certs

TheoBrigitte:
> We do need to check how to handle the renewal of the certs

I checked and it works automagically. I don't have the exact details but I think Prometheus just read those files at scrape time.
When we change the secret it gets propagated into the prometheus container and then Prometheus happily read the new file without even a need for reloading.

QuentinBisson:
Amazing

QuentinBisson (approved):

PR #882, opened 2022-05-04
Improve capi detection

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/21950

currently all AWS WCs are considered capi clusters, so scraping of `etcd` endpoint (and maybe others?) doesn't work.

this PR fixes the CAPI cluster detection function

## Checklist

I have:


service/key/key.go
```diff
@@ -267,15 +269,31 @@ func RemoteWritePasswordKey() string {
 // We do not have a provider agnostic label like "giantswarm.io/version" to define this.
```
QuentinBisson, line 269: I guess this comment is not needed anymore

service/key/key.go
```diff
@@ -267,15 +269,31 @@ func RemoteWritePasswordKey() string {
...
 func IsCAPICluster(obj metav1.Object) bool {
 	// TODO once we have migrated all clusters to CAPI, we can remove this
-	switch v := obj.(type) {
-	case *capiv1alpha3.Cluster:
-		if _, ok := v.Labels["azure-operator.giantswarm.io/version"]; ok {
+
+	checker := func(labels map[string]string) bool {
```
QuentinBisson, line 273: @whites11 is this code anywhere else ? Would it make sense to package that as a giantswarm lib ? Something like capi-utils ?

QuentinBisson (approved):
LGTM if tested on all providers

service/key/key.go
whites11: thanks, fixed

service/key/key.go
whites11: I would not be sure where to put it honestly.
I agree it would be nice though.
Task for a real PE I guess :sweat_smile:

service/key/key.go
```diff
@@ -263,19 +265,34 @@ func RemoteWritePasswordKey() string {
...
+		if _, ok := labels["azure-operator.giantswarm.io/version"]; ok {
+			return false
+		}
+		if _, ok := labels["release.giantswarm.io/version"]; ok {
```
fiunchinho, line 276: Maybe I'm missing something, but we are definitely adding the `release.giantswarm.io/version` label to CAPI clusters. For example, on AWS https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/templates/_helpers.tpl#L29

service/key/key.go
```diff
@@ -263,19 +265,34 @@ func RemoteWritePasswordKey() string {
 	return "password"
 }
 
-// IsCAPICluster returns true if the cluster is in v1alpha3 and does not have the "azure-operator.giantswarm.io/version" label added by the azure operator.
-// We do not have a provider agnostic label like "giantswarm.io/version" to define this.
+// IsCAPICluster returns true if the cluster has any of the legacy labels such as azure-operator.giantswarm.io/version.
```
Gacko, line 268: Are you sure it returns `true` if the cluster has any of the legacy labels? It rather reads like returning `false` below.

service/key/key.go
```diff
@@ -263,19 +265,34 @@ func RemoteWritePasswordKey() string {
...
 	}
+
+	switch v := obj.(type) {
+	case *capiv1alpha3.Cluster:
+		return checker(v.Labels)
+	case *capiv1alpha4.Cluster:
+		return checker(v.Labels)
+	case *capiv1beta1.Cluster:
+		return checker(v.Labels)
+	}
```
fiunchinho, line 292: Why the switch? why not always to call `checker(v.Labels)` ?

whites11: `.(type)` can only be used in a `switch`

service/key/key.go
whites11: ok this is my mistake. I couldn't test with CAPI so I had to go yolo

whites11: https://github.com/giantswarm/prometheus-meta-operator/pull/885

service/key/key.go
whites11: https://github.com/giantswarm/prometheus-meta-operator/pull/885

PR #891, opened 2022-05-06
Use 'ip' node label as target to scrape etcd on MCs.

whites11:
The main node IP as written by the kubelet in the `Node` status field, is unreliable.
In AWS MCs this can be a random POD cidr IP and when that's the case etcd scraping fails.

This PR looks for a label named "ip" in the Node resource and, if present, uses that label for scraping etcd.
The label is set by us on MCs with a well known correct ip address.
This makes scrape success much more likely.

## Checklist

I have:


AverageMarcus (approved):
LGTM

QuentinBisson (approved):

PR #893, opened 2022-05-09
relax prometheus readinessProbe

TheoBrigitte:


TheoBrigitte:
### Before

```
$ k get po prometheus-gauss-0 -oyaml|yq -y -r '(.spec.containers[]|select(.name == "prometheus")|.readinessProbe)'
failureThreshold: 120
httpGet:
  path: /gauss/-/ready
  port: web
  scheme: HTTP
periodSeconds: 5
successThreshold: 1
timeoutSeconds: 3
```

### After

```
$ k get po prometheus-gauss-0 -oyaml|yq -y -r '(.spec.containers[]|select(.name == "prometheus")|.readinessProbe)'
failureThreshold: 120
httpGet:
  path: /gauss/-/ready
  port: web
  scheme: HTTP
initialDelaySeconds: 300
periodSeconds: 5
successThreshold: 1
timeoutSeconds: 3
```

service/controller/resource/monitoring/prometheus/test/case-1-awsconfig.golden
```diff
@@ -32,6 +32,11 @@ spec:
       certFile: /etc/prometheus/secrets/cluster-certificates/crt
       keyFile: /etc/prometheus/secrets/cluster-certificates/key
   arbitraryFSAccessThroughSMs: {}
+  containers:
```
TheoBrigitte, line 35: This fields allow to override any container field, see https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#prometheusspec

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -257,6 +257,15 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 		},
 	}
 
+	prometheus.Spec.Containers = []corev1.Container{
+		{
+			Name: "prometheus",
+			ReadinessProbe: &corev1.Probe{
+				InitialDelaySeconds: 300,
```
QuentinBisson, line 264: Can you explain a bit why you set it to 5 minutes ? I'm not a big fan of magic numbers

TheoBrigitte: Well sorry, but that's a magic number.

Current readiness probe, fail after 10mn, so adding 5mn to it feels correct to me. Ideally this should be proportional to amount of stored data.

QuentinBisson: Yes the number make sense :) My thinking here was that we could add a comment to explain where this is coming from :)

QuentinBisson: Like why you set 300. I know we have to deal with those at times xD

TheoBrigitte: good point, comment added :)

QuentinBisson (approved):

PR #900, opened 2022-05-19
Add boilerplate remotewrite controller

mcbenjemaa:
A new controller needed to reconcile `RemoteWrite` CustomResources 
This is just the boilerplate and the actual implementation.
 
issue: https://github.com/giantswarm/roadmap/issues/1110

## Checklist

I have:


service/controller/remotewrite/controller.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/giantswarm/operatorkit/v7/pkg/controller"
+	promclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+	vpa_clientset "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned"
+
+	"github.com/giantswarm/prometheus-meta-operator/pkg/project"
+)
+
+type ControllerConfig struct {
+	K8sClient        k8sclient.Interface
+	Logger           micrologger.Logger
```
QuentinBisson, line 16: Can you skim this down to just the 3 first ? You will not use most of the rest if any

service/service.go
```diff
@@ -272,6 +274,54 @@ func New(config Config) (*Service, error) {
 		}
 	}
 
+	var remotewriteController *remotewrite.Controller
+	{
+		if shouldCreateCAPIController(provider) {
```
QuentinBisson, line 279: Why is this only for CAPI Managemement cluster ? I do not think this filter is needed

service/controller/remotewrite/controller.go
mcbenjemaa: Sure! I wasn't sure about those attributes.

service/service.go
mcbenjemaa: Copy-paste stuff 😳

service/controller/remotewrite/controller.go
QuentinBisson: Thanks :)

QuentinBisson (approved):

service/service.go
```diff
@@ -272,6 +274,21 @@ func New(config Config) (*Service, error) {
...
+			Logger:           config.Logger,
+			PrometheusClient: prometheusClient,
+		}
+
+		remotewriteController, err = remotewrite.NewController(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+	}
```
TheoBrigitte, line 290: I would rather leave this out for now, and add it at last when we have resources and when everything works. So we keep PMO available for any upcoming and unforeseen fixes needed.

Also we this you need to boot the controller, basically adding a line further down in this file in the `Boot` method.

TheoBrigitte (changes requested):
Looks good, but would leave the final wiring out for now. Keep it in a separate PR so we can use it for testing.

service/service.go
mcbenjemaa: Agree, `remotewriteController` would wired in a separate PR.

TheoBrigitte:
CI is failing because of new nancy update which detect more vulnerabilities know. @hervenicol is working on a fix.

PR #903, opened 2022-05-26
Update dependencies

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/22217

* Updated dependencies I can update without breaking everything
* added ignores in `.nancy-ignore` file
* removed an obsolete ignore from `.nancy-ignore`

## Checklist

I have:


TheoBrigitte (approved):
Thanks, this looks good, even though that's a lot of CVEs. Maybe @giantswarm/sig-security has word to say before we merge which is going to ignore 15 CVEs for the next ~3months.

Just by curiosity why 1st of August as expiration date?

hervenicol:
I tried to find a date:
- not too much in the future because I don't want these ignores to last too long
- but far enough so we have time to update the repos we rely on which have these CVEs.

I believed 2 months should be OK.
Now that I think about it, it means the builds are probably going to break in the middle of the summer when everybody's off, maybe we should rather set it to mid-September (at least after the on-site).

stone-z:
Sorry for late reply, I'm on a North America time zone now. When the expiration is up, some of these can be `replace`d instead of ignored. See https://github.com/giantswarm/architect/pull/736 for reference

PR #907, opened 2022-05-27
Ignore remotewrite field when updating prometheus CR

mcbenjemaa:
This PR changes the Prometheus controller to ignore the diff of `remoteWriteSpec` and ignore the update to this field.

issue: https://github.com/giantswarm/roadmap/issues/1111

## Checklist

I have:

- [x] Updated changelog in `CHANGELOG.md` (No need)

### DIFF
--- a/service/controller/resource/monitoring/prometheus/resource.go
+++ b/service/controller/resource/monitoring/prometheus/resource.go
@@ -389,12 +389,12 @@ func hasChanged(current, desired metav1.Object) bool {
 	c := current.(*promv1.Prometheus)
 	d := desired.(*promv1.Prometheus)
 
-	return !cmp.Equal(c.Spec, d.Spec, cmpopts.EquateEmpty(), cmpopts.IgnoreFields(promv1.PrometheusSpec{}, "RemoteWrite"))
+	return !cmp.Equal(c.Spec, d.Spec, cmpopts.IgnoreFields(promv1.PrometheusSpec{}, "RemoteWrite"))
 }
 
 // Fetch current Prometheus CR and update RemoteWrite field
 func currentRemoteWrite(ctx context.Context, config Config, p *promv1.Prometheus) error {
-	current, err := config.PrometheusClient.MonitoringV1().Prometheuses(p.Namespace).Get(ctx, p.Name, metav1.GetOptions{})
+	current, err := config.PrometheusClient.MonitoringV1().Prometheuses(p.GetNamespace()).Get(ctx, p.GetName(), metav1.GetOptions{})
 	if err != nil {
 		return microerror.Mask(err)
 	}


.gitignore
```diff
@@ -2,3 +2,4 @@
 TODO
 /prometheus-meta-operator
 !vendor/**
+.idea/
```
hervenicol, line 5: I think this should be in your personal global gitignore list.
See https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/ for more info.

mcbenjemaa: Thanks

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -431,5 +383,5 @@ func hasChanged(current, desired metav1.Object) bool {
 	c := current.(*promv1.Prometheus)
 	d := desired.(*promv1.Prometheus)
 
-	return !reflect.DeepEqual(c.Spec, d.Spec)
+	return !cmp.Equal(c.Spec, d.Spec, cmpopts.EquateEmpty(), cmpopts.IgnoreFields(promv1.PrometheusSpec{}, "RemoteWrite"))
```
TheoBrigitte, line 386: Why do we need to add `cmpopts.EquateEmpty()` here ?

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -270,54 +270,6 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 		},
 	}
 
-	if config.RemoteWriteURL != "" {
```
TheoBrigitte, line 273: Along with removing this we can also removing the config and command line flag propagating this value.

service/controller/resource/monitoring/prometheus/resource.go
mcbenjemaa: for computing diff of pointers fields, which means empty fields will be equal.

mcbenjemaa:
@TheoBrigitte ready for review

TheoBrigitte (commented):
Nice job :tada: 

Just few remarks

service/controller/resource/monitoring/prometheus/resource.go
mcbenjemaa:
``` 
// EquateEmpty returns a Comparer option that determines all maps and slices
// with a length of zero to be equal, regardless of whether they are nil.
``` 

for me it's fine, we can remove it. I don't see a strong case!

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -424,12 +375,29 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 		}
 	}
 
+	if config.PrometheusClient != nil {
```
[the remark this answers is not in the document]
mcbenjemaa: But it's not set for Tests :(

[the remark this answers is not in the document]
mcbenjemaa: I think its better then mocking the client

service/controller/resource/monitoring/prometheus/resource.go
TheoBrigitte: Ok, let's please remove this then.

service/controller/resource/monitoring/prometheus/resource.go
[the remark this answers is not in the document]
TheoBrigitte: oh, so test fails then. Fine with me then.

TheoBrigitte (approved):
Looks alright to me.

PR #915, opened 2022-06-10
Run PMO as non root

QuentinBisson:
This PR adds an init container to change the permission of the etcd certs to non root so the PMO container can use them

## Checklist

I have:

- [x] Described why this change is being introduced Towards https://github.com/giantswarm/giantswarm/issues/22291

### DIFF
--- a/helm/prometheus-meta-operator/templates/deployment.yaml
+++ b/helm/prometheus-meta-operator/templates/deployment.yaml
@@ -46,6 +46,21 @@ spec:
       securityContext:
         runAsUser: {{ .Values.pod.user.id }}
         runAsGroup: {{ .Values.pod.group.id }}
+        runAsNonRoot: true
+      initContainers:
+      - name: volume-mount-hack
+        image: busybox
+        command: ["sh", "-c", "chmod -R 644 /etcd-client-certs"]
+        securityContext:
+          runAsUser: 0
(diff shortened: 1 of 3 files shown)


QuentinBisson:
Tested on Gauss

helm/prometheus-meta-operator/templates/deployment.yaml
```diff
@@ -46,6 +46,21 @@ spec:
       securityContext:
         runAsUser: {{ .Values.pod.user.id }}
         runAsGroup: {{ .Values.pod.group.id }}
+        runAsNonRoot: true
+      initContainers:
+      - name: volume-mount-hack
+        image: busybox
+        command: ["sh", "-c", "chmod -R 644 /etcd-client-certs"]
```
TheoBrigitte, line 53: Do we really need this? Can't we set folder/file permissions somewhere else?

QuentinBisson: I do not see any other solution as we are using hostpath volumes

QuentinBisson: Can we discuss about it during refinement ? That looks like an interesting permission topic

QuentinBisson:
Closing for now

PR #905, opened 2022-05-26
Wire remotewrite controller

mcbenjemaa:
Wire remotewrite controller

issue:  https://github.com/giantswarm/roadmap/issues/1110

## Checklist

I have:

- [x] Described why this change is being introduced
- [ ] Separated out refactoring/reformatting in a dedicated PR
- [ ] Updated changelog in `CHANGELOG.md`

PR #916, opened 2022-06-13
Add test for Remotewrite feature

mcbenjemaa:
issue: https://github.com/giantswarm/roadmap/issues/1140

Adding unit tests for remote write functionality. 

## Checklist

I have:

### DIFF
--- a/service/controller/resource/prometheusremotewrite/resource_test.go
+++ b/service/controller/resource/prometheusremotewrite/resource_test.go
@@ -13,10 +13,13 @@ import (
 	"github.com/giantswarm/prometheus-meta-operator/pkg/unittest"
 )
 
-func TestToRemoteWrite(t *testing.T) {
+const (
+	name            = "simple-remotewrite"
+	namespace       = "default"
+	clusterSelector = "giant-cluster"
+)
 
-	name := "simple-remotewrite"
-	namespace := "default"
+func TestToRemoteWrite(t *testing.T) {
 
 	type args struct {
 		obj interface{}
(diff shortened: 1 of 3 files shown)


mcbenjemaa:
HOLD, until #905 gets merged.

service/controller/resource/prometheusremotewrite/resource_test.go
```diff
@@ -0,0 +1,396 @@
...
+	ok bool
+} {
+	rw.Spec.RemoteWrite.Name = rw.GetName()
+
+	r := NewResource()
+	if len(r.HTTPSProxy) > 0 {
+		rw.Spec.RemoteWrite.ProxyURL = r.HTTPSProxy
+	} else if len(r.HTTPProxy) > 0 {
+		rw.Spec.RemoteWrite.ProxyURL = r.HTTPProxy
+	}
```
TheoBrigitte, line 286: Why do we have this everywhere?

I would rather have a different test for the http proxy, instead of having this piece of code duplicated in every test.

TheoBrigitte (approved):
Looks alright.

But please move the http proxy to its own test if possible.

service/controller/resource/prometheusremotewrite/resource_test.go
mcbenjemaa: because `ensurePrometheusRemoteWrite()` have it, so it returns the proxy in the remotewrite config.
If don't use them, tests will fail.

mcbenjemaa: I added `TestEnsurePrometheusRemoteWriteWithPoxy()` and I added an argument `proxy bool` to  `expectedPrometheusAppend()`.

TheoBrigitte (approved):

PR #919, opened 2022-06-15
Supply HTTP proxy in remotewrite config

mcbenjemaa:
Issue: https://github.com/giantswarm/roadmap/issues/1169

Adding `HTTP_PROXY` If needed, once set up in the controller. 

## Checklist

I have:

### DIFF
--- a/service/controller/resource/prometheusremotewrite/create.go
+++ b/service/controller/resource/prometheusremotewrite/create.go
@@ -30,7 +30,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 		for _, current := range prometheusList.Items {
 
-			desired, ok := ensurePrometheusRemoteWrite(r, *remoteWrite, *current)
+			desired, ok := r.ensurePrometheusRemoteWrite(*remoteWrite, *current)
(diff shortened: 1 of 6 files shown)


service/controller/resource/prometheusremotewrite/create.go
```diff
@@ -30,7 +30,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 		for _, current := range prometheusList.Items {
 
-			desired, ok := ensurePrometheusRemoteWrite(*remoteWrite, *current)
+			desired, ok := ensurePrometheusRemoteWrite(r, *remoteWrite, *current)
```
TheoBrigitte, line 33: I would recommend to make this function a method of `Resource` instead of passing `r` as an argument.

So this becomes:

```suggestion
			desired, ok := r.ensurePrometheusRemoteWrite(*remoteWrite, *current)
```

TheoBrigitte (changes requested):

QuentinBisson (approved):
LGTM

TheoBrigitte (approved):
LGTM

PR #928, opened 2022-06-22
Add Field in the remotewrite CRD to sync secrets

mcbenjemaa:
Towards: https://github.com/giantswarm/roadmap/issues/1180
 
Add Secrets field to remotewrite CRD, in order to sync them to the Prometheus namespace

```yaml
secrets:
- name: grafana-cloud-basic-auth
  data:
    username: <Base64>
    password: <Base64> 
```

Ref: https://github.com/giantswarm/roadmap/issues/1180#issuecomment-1162802666

## Checklist

I have:

- [X] Described why this change is being introduced
- [ ] Separated out refactoring/reformatting in a dedicated PR
- [X] Updated changelog in `CHANGELOG.md`

PR #930, opened 2022-06-28
Implement sync RemoteWrite Secrets logic

mcbenjemaa:
Towards https://github.com/giantswarm/roadmap/issues/1180

After adding https://github.com/giantswarm/prometheus-meta-operator/pull/928 
In this PR, I added the logic of the sync.
By adding a new resource `remotewritesecret`.
and implementing `EnsureCreated()` and `EnsureDeleted`

## Checklist

I have:

### DIFF
--- a/service/controller/remotewrite/resource.go
+++ b/service/controller/remotewrite/resource.go
@@ -44,8 +44,8 @@ func newResources(config ControllerConfig) ([]resource.Interface, error) {
 	}
 
 	resources := []resource.Interface{
-		prometheusRemoteWrite,
 		rwSecretResource,
+		prometheusRemoteWrite,
 	}
 
 	return resources, nil
--- a/service/controller/remotewrite/resource.go
+++ b/service/controller/remotewrite/resource.go
@@ -5,6 +5,7 @@ import (
 	"github.com/giantswarm/operatorkit/v7/pkg/resource"
 
 	"github.com/giantswarm/prometheus-meta-operator/service/controller/resource/prometheusremotewrite"
+	"github.com/giantswarm/prometheus-meta-operator/service/controller/resource/remotewritesecret"
 )
 
 func newResources(config ControllerConfig) ([]resource.Interface, error) {
@@ -28,8 +29,23 @@ func newResources(config ControllerConfig) ([]resource.Interface, error) {
 		}
 	}
 
+	var rwSecretResource resource.Interface
+	{
+		c := remotewritesecret.Config{
+			K8sClient:        config.K8sClient,
+			Logger:           config.Logger,
+			PrometheusClient: config.PrometheusClient,
+		}
+
+		rwSecretResource, err = remotewritesecret.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
 	resources := []resource.Interface{
 		prometheusRemoteWrite,
+		rwSecretResource,
 	}
 
 	return resources, nil
--- a/service/controller/resource/remotewritesecret/create.go
+++ b/service/controller/resource/remotewritesecret/create.go
@@ -1,0 +1,59 @@
+package remotewritesecret
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/v7/pkg/controller/context/resourcecanceledcontext"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	r.logger.Debugf(ctx, "ensuring prometheus remotewrite secret")
+	{
+		remoteWrite, err := ToRemoteWrite(obj)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		// fetch current prometheus using the selector provided in remoteWrite resource.
+		prometheusList, err := fetchPrometheusList(ctx, r, remoteWrite)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		if prometheusList == nil || len(prometheusList.Items) == 0 {
+			r.logger.Debugf(ctx, "no prometheus found, cancel reconciliation")
+			resourcecanceledcontext.SetCanceled(ctx)
+			return nil
+		}
+
+		for _, current := range prometheusList.Items {
+
+			// Loop over remote write secrets
+			for _, item := range remoteWrite.Spec.Secrets {
+				desired := r.ensureRemoteWriteSecret(item, remoteWrite.ObjectMeta, current.GetNamespace())
+				secret, err := r.k8sClient.K8sClient().CoreV1().Secrets(current.GetNamespace()).Get(ctx, item.Name, metav1.GetOptions{})
+				if apierrors.IsNotFound(err) {
+					r.logger.Debugf(ctx, fmt.Sprintf("creating Secret %#q in namespace %#q", desired.Name, desired.Namespace))
+
+					//if err := r.k8sClient.K8sClient().SetControllerReference(p, &desiredSa, r.Scheme); err != nil {
+					//	return err
+					//}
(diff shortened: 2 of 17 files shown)


service/controller/remotewrite/resource.go
```diff
@@ -28,8 +29,23 @@ func newResources(config ControllerConfig) ([]resource.Interface, error) {
...
+
+		rwSecretResource, err = remotewritesecret.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
 	resources := []resource.Interface{
 		prometheusRemoteWrite,
+		rwSecretResource,
```
TheoBrigitte, line 48: I think `rwSecretResource` should go first as its better to create secrets before configuring the remote write field.

```suggestion
		rwSecretResource,
		prometheusRemoteWrite,
```

service/controller/resource/remotewritesecret/create.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/v7/pkg/controller/context/resourcecanceledcontext"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	r.logger.Debugf(ctx, "ensuring prometheus remotewrite secret")
+	{
+		remoteWrite, err := ToRemoteWrite(obj)
```
TheoBrigitte, line 16: I think the `ToRemoteWrite` function could be moved into a different packages and be used here and in the `prometheusRemoteWrite` resource also.

service/controller/resource/remotewritesecret/create.go
```diff
@@ -0,0 +1,59 @@
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		// fetch current prometheus using the selector provided in remoteWrite resource.
+		prometheusList, err := fetchPrometheusList(ctx, r, remoteWrite)
```
TheoBrigitte, line 22: Same, we could share the `fetchPrometheusList` func.

service/controller/resource/remotewritesecret/create.go
```diff
@@ -0,0 +1,59 @@
...
+			for _, item := range remoteWrite.Spec.Secrets {
+				desired := r.ensureRemoteWriteSecret(item, remoteWrite.ObjectMeta, current.GetNamespace())
+				secret, err := r.k8sClient.K8sClient().CoreV1().Secrets(current.GetNamespace()).Get(ctx, item.Name, metav1.GetOptions{})
+				if apierrors.IsNotFound(err) {
+					r.logger.Debugf(ctx, fmt.Sprintf("creating Secret %#q in namespace %#q", desired.Name, desired.Namespace))
+
+					//if err := r.k8sClient.K8sClient().SetControllerReference(p, &desiredSa, r.Scheme); err != nil {
+					//	return err
+					//}
+					secret, err = r.k8sClient.K8sClient().CoreV1().Secrets(current.GetNamespace()).Create(ctx, &desired, metav1.CreateOptions{})
```
TheoBrigitte, line 44: I think we need a cleanup mechanism to ensure we do not leave orphaned secrets behind.

When an entry is removed from `remoteWrite.Spec.Secrets` how do we make sure the associated secret is also deleted ?

mcbenjemaa: Has been added here:
https://github.com/giantswarm/prometheus-meta-operator/pull/930/files#diff-e16e10a7a44f3be7e69b6787985c26f000958e235184a68d897c45ee6a684c5cR56

service/controller/resource/remotewritesecret/resource.go
```diff
@@ -0,0 +1,87 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	pmov1alpha1 "github.com/giantswarm/prometheus-meta-operator/api/v1alpha1"
+	"github.com/giantswarm/prometheus-meta-operator/pkg/remotewriteutils"
+)
+
+const (
+	Name  = "remotewrite"
+	label = "monitoring.giantswarm.io/remotewrite"
```
TheoBrigitte, line 18: Using `monitoring.giantswarm.io` feels wrong, how about :

 ```suggestion
	label = "giantswarm.io/resource"
```

**Comment by TheoBrigitte** [comment]
I am now thinking we should make the `clusterSelector` field immutable otherwise we might run into issues.

In case the `clusterSelector` is updated we would left remoteWrite config behind and their associated secrets.

File: service/controller/resource/remotewritesecret/resource.go (hunk 5 above)
Comment (mcbenjemaa) [reply service/controller/resource/remotewritesecret/resource.go:18]: I guess, I need to add another label that contains the name of `RemoteWrite` CR, otherwise we end up deleting other remote write secrets.

File: service/controller/resource/remotewritesecret/create.go [hunk 6]
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		for _, current := range prometheusList.Items {
+
+			installedSecrets := make([]corev1.Secret, 0)
+			// Loop over remote write secrets
+			for _, item := range remoteWrite.Spec.Secrets {
+				desired := r.ensureRemoteWriteSecret(item, remoteWrite.ObjectMeta, current.GetNamespace())
+				secret, err := r.k8sClient.K8sClient().CoreV1().Secrets(current.GetNamespace()).Get(ctx, item.Name, metav1.GetOptions{})
+				if apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
mcbenjemaa: Well, I think we still can support update of the secrets

service/controller/resource/remotewritesecret/create.go
```diff
@@ -0,0 +1,87 @@
...
+				r.logger.Debugf(ctx, fmt.Sprintf("Secret %#q in namespace %#q created", secret.Name, secret.Namespace))
+
+			}
+
+			/*
+			  Cleanup deleted secrets from RemoteWrite CR
+			*/
+			l := labels.SelectorFromSet(labels.Set(map[string]string{
+				label:          Name,
+				labelName:      remoteWrite.GetName(),
```
mcbenjemaa, line 60: I added label `name` & `namespace` to only fetch secrets related to current remotewrite.

service/controller/resource/remotewritesecret/delete.go
```diff
@@ -0,0 +1,57 @@
...
+			*/
+			l := labels.SelectorFromSet(labels.Set(map[string]string{
+				label:          Name,
+				labelName:      remoteWrite.GetName(),
+				labelNamespace: remoteWrite.GetNamespace(),
+			}))
+			listOptions := metav1.ListOptions{
+				LabelSelector: l.String(),
+			}
+			err := r.k8sClient.K8sClient().CoreV1().Secrets(current.GetNamespace()).DeleteCollection(ctx, metav1.DeleteOptions{}, listOptions)
```
mcbenjemaa, line 46: In the `EnsureDeleted()` I fetch related secrets using label and I perform `DeleteCollection()`

service/controller/resource/remotewritesecret/delete.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		// fetch current prometheus using the selector provided in remoteWrite resource.
+		prometheusList, err := remotewriteutils.FetchPrometheusList(ctx, toResourceWrapper(r), remoteWrite)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		if prometheusList == nil && len(prometheusList.Items) == 0 {
+			r.logger.Debugf(ctx, "no prometheus found, cancel reconciliation")
+			resourcecanceledcontext.SetCanceled(ctx)
```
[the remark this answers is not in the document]
TheoBrigitte: We should just return `nil` and not cancel reconciliation here.

Canceling reconciliation will leave the finalizer on the `RemoteWrite` resource and it will therefore not be deleted. Implementation details in [operatorkit](https://github.com/giantswarm/operatorkit/blob/dd879b4b4909d2fefe669ea2fbfb6095c8bff845/pkg/controller/controller.go#L439-L478)

TheoBrigitte (approved):

PR #935, opened 2022-06-30
add remoteWrite.status field

TheoBrigitte:


api/v1alpha1/remoteWrite_types.go
```diff
@@ -32,6 +34,15 @@ type RemoteWriteSecretSpec struct {
 	Data map[string][]byte `json:"data,omitempty"`
 }
 
+type RemoteWriteStatus struct {
+	ConfiguredPrometheuses []RemoteWriteStatusConfiguredPrometheus `json:"configuredPrometheuses"`
```
mcbenjemaa, line 38: it's very long name for type `RemoteWriteStatusConfiguredPrometheus`

I would prefer using an upstream type instead,`ObjectReference`
https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference

in fact, you can choose what you want to supply in this reference, which means the name and namespace is enough in this case.

TheoBrigitte:
patch does not work

```
E 07/01 00:08:26 monitoring/grafana-cloud failed to reconcile | operatorkit/v7/pkg/controller/controller.go:326 | controller=prometheus-meta-operator-remote-write-controller | event=update | loop=5 | version=696171504
        /go/pkg/mod/github.com/giantswarm/operatorkit/v7@v7.0.1/pkg/controller/controller.go:532
        /go/pkg/mod/github.com/giantswarm/operatorkit/v7@v7.0.1/pkg/controller/controller.go:567
        /root/project/service/controller/resource/prometheusremotewrite/create.go:60
        unknown: remotewrites.monitoring.giantswarm.io "grafana-cloud" not found
```

api/v1alpha1/remoteWrite_types.go
TheoBrigitte: Added this, good point. Thanks

TheoBrigitte:
The following annotation is required to make the `status` field a subresources (see https://book-v1.book.kubebuilder.io/basics/status_subresource.html)
```
// +kubebuilder:subresource:status
````

**Comment by TheoBrigitte** [comment]
status is updated :heavy_check_mark: 
```yaml
status:
  configuredPrometheuses:
  - name: ginger
    namespace: ginger-prometheus
```

pkg/remotewriteutils/util.go
```diff
@@ -25,6 +26,12 @@ func ToRemoteWrite(obj interface{}) (*pmov1alpha1.RemoteWrite, error) {
 		return nil, microerror.Maskf(wrongTypeError, "'%T' is not a 'pmov1alpha1.RemoteWrite'", obj)
 	}
 
+	kind := reflect.TypeOf(remotewrite).Elem().Name()
```
mcbenjemaa, line 29: I wonder if this is still needed after adding the subresource marker.

service/controller/resource/prometheusremotewrite/create.go
```diff
@@ -66,3 +55,82 @@ func updateMeta(c, d metav1.Object) {
 	d.SetDeletionGracePeriodSeconds(c.GetDeletionGracePeriodSeconds())
 	d.SetManagedFields(c.GetManagedFields())
 }
+
+func (r *Resource) setRemoteWrite(ctx context.Context, remoteWrite *pmov1alpha1.RemoteWrite, prometheus *promv1.Prometheus) error {
+	err := r.ensureStatusCreated(ctx, remoteWrite, prometheus)
```
mcbenjemaa, line 60: I'm guessing to put this after the update of Prometheus CR, as the last call in this function.
To ensure that the Prometheus has been updated.

mcbenjemaa: We would end up having a status for Prometheus that has failed to configure the remotewrite for some reason.

service/controller/resource/prometheusremotewrite/delete.go
```diff
@@ -31,27 +33,58 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
...
 		}
 
 	}
 	r.logger.Debugf(ctx, "deleted prometheus remoteWrite config")
 
 	return nil
 }
+
+func (r *Resource) unsetRemoteWrite(ctx context.Context, remoteWrite *pmov1alpha1.RemoteWrite, prometheus *promv1.Prometheus) error {
+	err := r.ensureStatusDeleted(ctx, remoteWrite, prometheus)
```
mcbenjemaa, line 54: Same as before, I would suggest putting this call after updating the Prometheus CR

mcbenjemaa:
This PR related to https://github.com/giantswarm/roadmap/issues/1213

QuentinBisson (approved):
That looks pretty good to me considering I'm not in the removewrite codebase

pkg/remotewriteutils/util.go
TheoBrigitte: Good point this might not be needed. But this need to be tested.

service/controller/resource/prometheusremotewrite/delete.go
```diff
@@ -31,27 +33,58 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
 		}
 
 		for _, current := range prometheusList.Items {
```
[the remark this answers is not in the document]
TheoBrigitte: This is bad because the loop won't continue and other Prometheus would not be unconfigured. But without this, errors are ignored which is also bad.
We need a way to catch this error when we remove this check

api/v1alpha1/remoteWrite_types.go
```diff
@@ -32,6 +36,11 @@ type RemoteWriteSecretSpec struct {
 	Data map[string][]byte `json:"data,omitempty"`
 }
 
+type RemoteWriteStatus struct {
+	ConfiguredPrometheuses []corev1.ObjectReference `json:"configuredPrometheuses,omitempty"`
+	SyncedSecrets          []corev1.ObjectReference `json:"syncedSecrets,omitempty"`
```
mcbenjemaa, line 41: I have added this type to distinguish secrets

PR #940, opened 2022-07-13
Monitor using a pod monitor

QuentinBisson:
## Checklist

I have:


QuentinBisson:
Tested on gollum

helm/prometheus-meta-operator/templates/pod-monitor.yaml
```diff
@@ -0,0 +1,16 @@
...
+    name: {{ include "resource.default.name" . }}
+spec:
+  namespaceSelector:
+    matchNames:
+      - {{ include "resource.default.namespace" . }}
+  selector:
+    matchLabels:
+      {{- include "labels.selector" . | nindent 6 }}
+  podMetricsEndpoints:
+    - targetPort: 8000
```
hervenicol, line 16: Do you have to use the port number there, or could you use its name `http` ?

QuentinBisson: We cannot as the port is not exposed in the deployment but that's doable if we expose the port and name it

QuentinBisson:
@giantswarm/team-atlas can I get a quick look?

hervenicol:
The PR could use some more context.

* Why a `podmonitor` and not a `servicemonitor`?
* Wouldn't you want to remove the `service` annotation `giantswarm.io/monitoring`?
* isn't there an issue related to this PR?

QuentinBisson:
> The PR could use some more context.
> 
> * Why a `podmonitor` and not a `servicemonitor`?
Because PMO does not need a Service to run so creating one just for metrics seems useless
> * Wouldn't you want to remove the `service` annotation `giantswarm.io/monitoring`?
Probably yes
> * isn't there an issue related to this PR?
No

hervenicol:
> > * Why a `podmonitor` and not a `servicemonitor`?
>
> Because PMO does not need a Service to run so creating one just for metrics seems useless
>
> > * Wouldn't you want to remove the `service` annotation `giantswarm.io/monitoring`?
>
> Probably yes

Then we can even remove the whole service, if it's useless ?

QuentinBisson:
The service is gone in the PR already :)

PR #942, opened 2022-07-13
push to cloud-director-app-collection

glitchcrab:
push to cloud-director-app-collection


QuentinBisson (approved):
We just released so you might have to put the app in the collection.
Also you need to add the [prometheus-remotewrite](https://github.com/giantswarm/prometheus-remotewrite) app and it's given config also

PR #948, opened 2022-07-18
expose errors total metric for remotewrite

mcbenjemaa:
## Checklist

I have:


go.mod
```diff
@@ -30,13 +31,13 @@ require (
...
 require (
 	github.com/blang/semver v3.5.1+incompatible // indirect
+	github.com/giantswarm/operatorkit v1.2.0 // indirect
 	github.com/go-kit/log v0.2.0 // indirect
 	github.com/gobuffalo/flect v0.2.3 // indirect
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
 	github.com/onsi/gomega v1.16.0 // indirect
 	github.com/pelletier/go-toml/v2 v2.0.1 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
-	github.com/prometheus/client_golang v1.12.1 // indirect
```
hervenicol, line 39: You have to `go mod tidy` after that, otherwise tests fail.

service/controller/resource/prometheusremotewrite/resource_test.go
```diff
@@ -233,7 +233,7 @@ func remoteWrite(name, namespace, clusterSelector string) pmov1alpha1.RemoteWrit
 	}
 }
 
-func prometheus() promv1.Prometheus {
```
hervenicol, line 236: Having a look at this function, is it normal that we have a hardcoded version?

hervenicol (changes requested):

service/controller/resource/prometheusremotewrite/resource_test.go
mcbenjemaa: This function is hardcoded for test,  to just ensure that the underlying function is working as expected.

go.mod
mcbenjemaa: done

PR #952, opened 2022-07-21
node-exporter namespace can be monitoring or kube-system

fiunchinho:
In WCs and the new MCs created for CAPI we deploy `node-exporter` on the `kube-system` namespace. That way WCs and MCs are more similar between them.

But in the vintage product, `node-exporter` is deployed to the `monitoring` namespace for MCs. It's hard to know which namespace would be used depending on the `provider` because `aws` and `azure` could be CAPI and vintage at the same time.

So instead, we relax the regex for `node-exporter` to include both namespaces.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- node-exporter namespace can be `monitoring` orr `kube-system`.
+- node-exporter namespace can be `monitoring` or `kube-system`.
 
 ## [4.1.0] - 2022-07-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- node-exporter namespace can be `monitoring` orr `kube-system`.
+
 ## [4.1.0] - 2022-07-20
 
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -43,10 +47,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Rename `vcd` to `cloud-director`
 
-### Fixed
```
fiunchinho, line 46: It was duplicated

files/templates/scrapeconfigs/_node_exporter_namespace.yaml
```diff
@@ -1,7 +1,3 @@
 [[- define "_node_exporter_namespace" -]]
-[[- if eq .ClusterType "management_cluster" -]]
-monitoring
-[[- else -]]
-kube-system
-[[- end -]]
+monitoring|kube-system
```
kopiczko, line 2: I'd keep the `if` still because customers can deploy their own `node-exporter` to their own `monitoring` namespace.

fiunchinho: wouldn't that be covered already by the regex?

mcbenjemaa:
@fiunchinho Please make sure to fix the test.

files/templates/scrapeconfigs/_node_exporter_namespace.yaml
kopiczko: That's the problem. We don't want to scrape customer's node-exporter.

nprokopic:
@fiunchinho we can close this one is favor of https://github.com/giantswarm/prometheus-meta-operator/pull/956, unless you prefer to have two separate PRs.

nprokopic:
#956 is merged, so closing this one.

nprokopic:
> #956 is merged, so closing this one.

I should have refreshed the page first :smile:

PR #956, opened 2022-07-25
Update node-exporter and cert-exporter namespace

nprokopic:
Closes https://github.com/giantswarm/roadmap/issues/1265

In WCs and the new MCs created for CAPI we deploy `node-exporter` and `cert-exporter` to the `kube-system` namespace. That way WCs and MCs are more similar between them.

But in the vintage product, `node-exporter` and `cert-exporter` are deployed to the `monitoring` namespace for MCs.

This PR changes the logic for `node-exporter` and `cert-exporter` namespaces:
- Use `kube-system` for CAPI MCs and all WC
- Use `monitoring` for vintage MCs
- Assume `aws` and `azure` MCs are vintage (`CAPIManagementCluster` check that is used is checking if the provider is one of cloud-director, gcp, openstack or vsphere, see here https://github.com/giantswarm/prometheus-meta-operator/blob/master/service/key/key.go#L79)

## Checklist

I have:

### DIFF
--- a/files/templates/scrapeconfigs/_cert_exporter_namespace.yaml
+++ b/files/templates/scrapeconfigs/_cert_exporter_namespace.yaml
@@ -1,4 +1,4 @@
-[[- define "_node_exporter_namespace" -]]
+[[- define "_cert_exporter_namespace" -]]
 [[- if or (and (eq .ClusterType "management_cluster") (.CAPIManagementCluster)) (eq .ClusterType "workload_cluster") -]]
 kube-system
 [[- else -]]
--- a/files/templates/scrapeconfigs/_cert_exporter_namespace.yaml
+++ b/files/templates/scrapeconfigs/_cert_exporter_namespace.yaml
@@ -1,7 +1,7 @@
 [[- define "_cert_exporter_namespace" -]]
-[[- if or (and (eq .ClusterType "management_cluster") (.CAPIManagementCluster)) (eq .ClusterType "workload_cluster") -]]
-kube-system
-[[- else -]]
+[[- if and (eq .ClusterType "management_cluster") (.VintageManagementCluster) -]]
 monitoring
+[[- else -]]
+kube-system
 [[- end -]]
 [[- end -]]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Set node-exporter namespace to `kube-system` for CAPI MCs and all WC, and to `monitoring` for vintage MCs.
+- Set cert-exporter namespace to `kube-system` for CAPI MCs and all WC, and to `monitoring` for vintage MCs.
+
 ### Fixed
 
 - Added `pod_name` as a label to distinguish between multiple etcd pods when running in-cluster (e.g. CAPI).
--- a/files/templates/scrapeconfigs/_node_exporter_namespace.yaml
+++ b/files/templates/scrapeconfigs/_node_exporter_namespace.yaml
@@ -1,7 +1,7 @@
 [[- define "_node_exporter_namespace" -]]
-[[- if or (and (eq .ClusterType "management_cluster") (.CAPIManagementCluster)) (eq .ClusterType "workload_cluster") -]]
-kube-system
-[[- else -]]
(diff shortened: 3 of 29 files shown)


files/templates/scrapeconfigs/_cert_exporter_namespace.yaml
```diff
@@ -0,0 +1,7 @@
+[[- define "_cert_exporter_namespace" -]]
+[[- if or (and (eq .ClusterType "management_cluster") (.CAPIManagementCluster)) (eq .ClusterType "workload_cluster") -]]
```
hervenicol, line 2: I would have written this the other way around.
Something like:
```
[[- if and ((not .CAPIManagementCluster) (eq .ClusterType "workload_cluster"))  -]]
    # special case for MC on vintage clusters
    monitoring
[[- else -]] # standard case
    kube-system
[[- end -]]
```

I think this would make it easier to understand what should be the standard case, and remove the "if" the day we have no more vintage clusters.
A comment would be nice, but I'm not sure what's going to happen with comments there. :grimacing: 

Same for the node_exporter namespace test.

hervenicol (changes requested):
Maybe there should be some extra test cases for CAPI clusters?
Depends how much prometheus' config will diverge, don't know.

hervenicol:
Even though I requested changes, that's nothing critical.
If it's urgent for you to merge or if you think my remarks are nonsense, let's discuss, it's not a hard block! ;)

files/templates/scrapeconfigs/_cert_exporter_namespace.yaml
nprokopic: You're right, the other way around looks simpler (just for `ClusterType` it should be `management_cluster`, and not `workload_cluster`). It's like this now:

```
[[- if and (eq .ClusterType "management_cluster") (.VintageManagementCluster) -]]
monitoring
[[- else -]]
kube-system
[[- end -]]
[[- end -]]
```

I added `VintageManagementCluster`, as it looks simpler/clearer than using negation (`not .CAPIManagementCluster`).

hervenicol: Then, now I read it in a simpler form, I'm puzzled by this:
* it has to be `VintageManagementCluster`, AND `management_cluster`.
* But the name `VintageManagementCluster` containing `ManagementCluster` I would assume it's already a `management_cluster`.

However when reading the base function `IsCAPIManagementCluster`
* it only cares for the cloud-provider type (meaning there's no provider for which we could have some CAPI and some vintage clusters?)
* I don't see any check for the `ManagementCluster` part

I guess something's misleading with the naming :thinking:

nprokopic: > it has to be VintageManagementCluster, AND management_cluster

Yeah, it's far from ideal, but I didn't want to refactor code too much, as I'm not very familiar with the codebase.

The issue is that `IsCAPIManagementCluster` just checks for the provider, and I added `VintageManagementCluster` (for easier readability) which is just a negation of `IsCAPIManagementCluster`, while `.ClusterType` calls `IsInCluster` which then checks CRs, so both checks are needed here.

> meaning there's no provider for which we could have some CAPI and some vintage clusters?

Currently we don't have any provider that has both CAPI and vintage, and not sure if that will happen or not. The idea here is to have correct and simple enough solution for the current state of things, which mean that a provider is either CAPI or vintage, without overlap for now.

I thought about renaming `VintageManagementCluster` to `VintageProvider`, but that sounded then even more confusing, since we never called providers vintage.

hervenicol: I fear building on top of unclear functions will make it harder and harder to fix. But I understand you don't want to touch stuff you're unsure about, I wouldn't do better.
Hopefully a someone wiser than us :mage_man: will look at it and fix stuff before we're lost :grin:

hervenicol (approved):
LG enough TM

files/templates/scrapeconfigs/_cert_exporter_namespace.yaml
nprokopic: The whole vintage vs CAPI story, and doing some stuff differently in those two is contributing to confusion across multiple fronts. I'm afraid it's gonna be like that, at least to some extent, until we completely move all MCs and WCs to CAPI world.

fiunchinho (approved):

erkanerol (approved):

PR #980, opened 2022-08-24
Refactor internal Proxy configuration

QuentinBisson:
This is boyscouting but I'm positive this will help when moving out of operatorkit

## Checklist

I have:

### DIFF
--- a/pkg/domain/proxy.go
+++ b/pkg/domain/proxy.go
@@ -1,0 +1,25 @@
+package domain
+
+import "strings"
+
+type ProxyConfiguration struct {
+	HTTPProxy  string
+	HTTPSProxy string
+
+	/* no_proxy is a comma- or space-separated list of machine
+	   or domain names, with optional :port part.  If no :port
+	   part is present, it applies to all ports on that domain.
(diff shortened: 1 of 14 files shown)


mcbenjemaa (approved):
Thanks for the change. 
LGTM

pkg/domain/proxy.go
```diff
@@ -0,0 +1,25 @@
...
+
+	/* no_proxy is a comma- or space-separated list of machine
+	   or domain names, with optional :port part.  If no :port
+	   part is present, it applies to all ports on that domain.
+	*/
+	NoProxy string
+}
+
+func (p ProxyConfiguration) GetURLForEndpoint(endpoint string) string {
+	if !strings.Contains(p.NoProxy, endpoint) {
```
TheoBrigitte, line 17: That's a weird condition, I don't think we are supposed to do any logic in here. Those values are mean to be passed down the stack, its all part of `golang.org/x/net/http/httpproxy` [spec](https://pkg.go.dev/golang.org/x/net/http/httpproxy#Config)

PR #998, opened 2022-09-15
Enable remote write receiver

QuentinBisson:
Enable the remote write receiver so the prometheus agent can send data to it's pair prometheus https://github.com/giantswarm/giantswarm/issues/23130

## Checklist

I have:

- [x] Described why this change is being introduced
- [x] Separated out refactoring/reformatting in a dedicated PR
- [x] Updated changelog in `CHANGELOG.md`

PR #999, opened 2022-09-15
Add remote write ingress boilerplate

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/23130
Add ingress boilerplates so the agent can access prometheus

## Checklist

I have:

### DIFF
--- a/service/controller/resource/monitoring/prometheus/resource.go
+++ b/service/controller/resource/monitoring/prometheus/resource.go
@@ -177,8 +177,11 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 			PodMetadata: &promv1.EmbeddedObjectMetadata{
 				Labels: labels,
 			},
-			LogLevel: config.LogLevel,
-			Image:    &image,
+			// TODO will have to migrate to EnableRemoteWriteReceiver: true but need CRD changes
+			// https://github.com/prometheus-operator/prometheus-operator/pull/4633/files
+			EnableFeatures: []string{"remote-write-receiver"},
+			LogLevel:       config.LogLevel,
+			Image:          &image,
 			Web: &promv1.WebSpec{
 				PageTitle: &pageTitle,
 			},
--- a/service/controller/resource/monitoring/remotewriteingress/v1beta1/client.go
+++ b/service/controller/resource/monitoring/remotewriteingress/v1beta1/client.go
@@ -1,0 +1,29 @@
+package v1beta1
+
+import (
+	"context"
+
+	networkingv1beta1 "k8s.io/api/networking/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/typed/networking/v1beta1"
+)
+
+type wrappedClient struct {
+	client v1beta1.IngressInterface
+}
+
+func (c wrappedClient) Create(ctx context.Context, o metav1.Object, options metav1.CreateOptions) (metav1.Object, error) {
+	return c.client.Create(ctx, o.(*networkingv1beta1.Ingress), options)
+}
+
+func (c wrappedClient) Update(ctx context.Context, o metav1.Object, options metav1.UpdateOptions) (metav1.Object, error) {
+	return c.client.Update(ctx, o.(*networkingv1beta1.Ingress), options)
+}
+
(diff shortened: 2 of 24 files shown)


service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -177,8 +177,11 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 			PodMetadata: &promv1.EmbeddedObjectMetadata{
 				Labels: labels,
 			},
-			LogLevel: config.LogLevel,
-			Image:    &image,
+			// TODO will have to migrate to EnableRemoteWriteReceiver: true but need CRD changes
+			// https://github.com/prometheus-operator/prometheus-operator/pull/4633/files
+			EnableFeatures: []string{"remote-write-receiver"},
```
mcbenjemaa, line 182: those changes are included here: https://github.com/giantswarm/prometheus-meta-operator/pull/998

mcbenjemaa:
can you cleanup the duplication with https://github.com/giantswarm/prometheus-meta-operator/pull/998

service/controller/resource/monitoring/remotewriteingress/v1beta1/resource.go
```diff
@@ -0,0 +1,80 @@
...
+	Name = "monitoringremotewriteingress"
+)
+
+type Config struct {
+	K8sClient  k8sclient.Interface
+	Logger     micrologger.Logger
+	BaseDomain string
+}
+
+func New(config Config) (*generic.Resource, error) {
```
mcbenjemaa, line 28: This file is same for `v1`  and `v1beta1`, can you use a generic one?

service/controller/resource/resource.go
```diff
@@ -123,6 +125,31 @@ func New(config Config) ([]resource.Interface, error) {
 		}
 	}
 
+	var remoteWriteIngressResource resource.Interface
+	if config.IngressAPIVersion == "networking.k8s.io/v1beta1" {
```
mcbenjemaa, line 129: and maybe you can wrap this check in the generic resource file.
If you want to use a generic file

service/controller/resource/monitoring/remotewriteingress/v1beta1/client.go
```diff
@@ -0,0 +1,29 @@
+package v1beta1
```
mcbenjemaa, line 1: you can also use one generic client. and eventually you check the APIVersion

service/controller/resource/monitoring/remotewriteingress/v1beta1/resource.go
QuentinBisson: You mean the config? I'd rather keep the 2 separate especially since I'm not sure we will keep v1beta1

service/controller/resource/monitoring/remotewriteingress/v1beta1/client.go
QuentinBisson: I really do not think so but you can try it yourself and we see how it goes :)

QuentinBisson: It its possible sure :)

QuentinBisson: I do not think focusing on v1beta1 makes much sense here

QuentinBisson:
@mcbenjemaa Creating reusable wrapper clients makes sense though but in a subsequent PR and not this one :)

service/controller/resource/monitoring/remotewriteingress/v1beta1/resource.go
mcbenjemaa: Yes!

mcbenjemaa:
Please Add some TODO comments

mcbenjemaa:
Is this change tested?

QuentinBisson:
Why would I add TODOs?

mcbenjemaa:
> Why would I add TODOs?

For instance, to mention that v1beta1  would be removed

QuentinBisson:
We use ingress v1beta1 outsider if this pr and I added an issue in the board for it :)

mcbenjemaa (approved):

QuentinBisson:
This pr Comes After #998

PR #1000, opened 2022-09-16
Add remote write agent secret

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/23130

This PR create a secret for the remote write agent to access the ingress on the management cluster (using a new package for password management and hashing) and the remote write config for the prometheus agent app itself.

Both those secrets are created in the <cluster_id>-prometheus namespace. It might make sense to move them to another one for the configmap to be added to the observability bundle

## Checklist

I have:


mcbenjemaa (commented):
what does this PR relates to https://github.com/giantswarm/prometheus-meta-operator/pull/1001

QuentinBisson:
#1001 adds the secret to the ingress

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
```diff
@@ -0,0 +1,117 @@
...
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: key.RemoteWriteAgentSecretName(),
+					},
+					Key: "password",
+				},
+			},
+		},
+	}
+
+	yamlValues, err := yaml.Marshal(remoteWrites)
```
mcbenjemaa, line 97: Does this provides yaml values?

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, microerror.Mask(err)
+	}
+
+	remoteWrites := []pov1.RemoteWriteSpec{
+		{
+			URL: fmt.Sprintf("%s/%s/api/v1/write", config.BaseDomain, key.ClusterID(cluster)),
+			BasicAuth: &pov1.BasicAuth{
+				Username: corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: key.RemoteWriteAgentSecretName(),
```
mcbenjemaa, line 83: you can use a constant in this package, instead of function `RemoteWriteAgentSecretName()`

service/controller/resource/monitoring/remotewriteagentconfigsecret/doc.go
```diff
@@ -0,0 +1,4 @@
+package remotewriteagentconfigsecret
+
+// This packages creates the remote write agent secret config
```
mcbenjemaa, line 3: Can you explain more, and be more specific! 
e.g. to be used as `values` for the `prometheus-agent` app

service/controller/resource/resource.go
```diff
@@ -150,6 +154,47 @@ func New(config Config) ([]resource.Interface, error) {
...
+			BaseDomain: config.PrometheusBaseDomain,
+		}
+
+		remoteWriteAgentConfigSecretResource, err = remotewriteagentconfigsecret.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var remoteWriteSecretResource resource.Interface
```
mcbenjemaa, line 185: I wonder, the `remotewritesecret` package for creating secrets from RemoteWrite CR. `spec.secrets`
Why this is needed?

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: I could but we tend to define constants in the key.go file. We could change the function to a constant though. Do you mind doing it?

service/controller/resource/resource.go
QuentinBisson: I think it's not and it's a wrong import

service/controller/resource/monitoring/remotewriteagentconfigsecret/doc.go
QuentinBisson: Done

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: I'm not sure what you expect me to write here. Could you provide more feedback on your review? The function yaml.Marshal marshals a go object into a yaml string so yes those are technically yaml values

QuentinBisson: Renamed to marshalledRemoteWrites

mcbenjemaa: Yeah, but i have used this marshal function which provides json output

mcbenjemaa: I just want to make sure that this generates proper yaml

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
mcbenjemaa: Yes

service/key/key.go
```diff
@@ -246,6 +247,10 @@ func RemoteWriteSecretName() string {
 	return "remote-write"
 }
 
+func RemoteWriteAgentConfigSecretName() string {
```
mcbenjemaa, line 250: Can you change this to constant as well

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: https://alisoftware.github.io/yaml/2021/08/17/yaml-part1-json/#:~:text=One%20thing%20that%20most%20people,represent%20the%20same%20data%20structures.

service/key/key.go
QuentinBisson: Done

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
```diff
@@ -0,0 +1,117 @@
...
+	return r, nil
+}
+
+func getObjectMeta(ctx context.Context, v interface{}) (metav1.ObjectMeta, error) {
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return metav1.ObjectMeta{}, microerror.Mask(err)
+	}
+
+	return metav1.ObjectMeta{
```
mcbenjemaa, line 60: Do we need to add labels like.

```
clusterID
organization
Installation 
etc? 
```

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
```diff
@@ -0,0 +1,101 @@
...
+	return r, nil
+}
+
+func getObjectMeta(ctx context.Context, v interface{}) (metav1.ObjectMeta, error) {
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return metav1.ObjectMeta{}, microerror.Mask(err)
+	}
+
+	return metav1.ObjectMeta{
```
mcbenjemaa, line 58: Same note for labels

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
```diff
@@ -0,0 +1,101 @@
...
+
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	username := key.ClusterID(cluster)
+
+	password, err := config.PasswordManager.GeneratePassword(32)
+	if err != nil {
+		return nil, microerror.Mask(err)
```
mcbenjemaa, line 78: Can you put more context, on what's failing here?

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	hashedPassword, err := config.PasswordManager.Hash(password)
+	if err != nil {
+		return nil, microerror.Mask(err)
```
mcbenjemaa, line 83: More context also

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
```diff
@@ -0,0 +1,101 @@
...
+			"password": []byte(password),
+			"username": []byte(username),
+		},
+		Type: "Opaque",
+	}
+	return secret, nil
+}
+
+func hasChanged(current, desired metav1.Object) bool {
+	return false
```
mcbenjemaa, line 100: Is the secret immutable?

pkg/password/manager.go
```diff
@@ -0,0 +1,33 @@
...
+
+func (m SimpleManager) GeneratePassword(length int) (string, error) {
+	bytes := make([]byte, length)
+	if _, err := rand.Read(bytes); err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(bytes), nil
+}
+
+func (m SimpleManager) Hash(plaintext string) (string, error) {
```
mcbenjemaa, line 27: We have an exported function `GeneratePassword` and `Hash`,
Can we add an exported function `GenerateHashedPassword` that calls them both?

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: What would you configure here?

QuentinBisson: I'd like to have this merged so getting ping pong will not help :p

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
QuentinBisson: What do you mean by context? Can you give an example?

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
QuentinBisson: Yes because the secret would be generated on every reconciliation otherwise

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
mcbenjemaa: If the controller fails in this stage, i don't know whats makes it fails.
If we add something like `error: password generation fails, err`:  
this gives me better view.

pkg/password/manager.go
QuentinBisson: Probably yes, but i'd rather have separate functions (one function one purpose). @TheoBrigitte any thoughts?

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
QuentinBisson: Sure I can log something. My thinking is the error should be self-sufficient but I can add logs

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
mcbenjemaa: Maybe, we need it to distinguish the created secrets.
Or we need it to implement cleanup mechanism (same we did in remotewrite secret).
But, i don't know, this was my suggestion.

Apart from that, what does `ping pong` means in this context 😄

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
mcbenjemaa: you can added it in the underlying function. if you want.

pkg/password/manager.go
QuentinBisson: Also, we cannot because we need both the secret and the hash for different purposes (hash for basic auth, password for the agent to be able to access prometheus remote write endpoint)

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: There is no copy of the secret so we do not need a copy mechanism as the secrets are created in the clsuter_id-prometheus namespace and they are in different namespaces so ther eis no need to differenciate them :)

Ping pong means I have to ask for more info when you could have added it on the initial review so we avoid back and forth to get to what you think is best in the end :) 

For example, you could have said something along the lines of:
Do we need to add some labels to this secret if we need to distinguish it from the other secrets (e.g. if they are in the same namespace) ?
That would have helped me get a better understanding of why you might think those labels could be needed :)

mcbenjemaa: thanks

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
QuentinBisson: It's added

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: Done

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
QuentinBisson: Done

PR #1001, opened 2022-09-16
Configure remote wrote agent ingress

QuentinBisson:
Needed so the remote write agent can send data to prometheus securely

## Checklist

I have:


QuentinBisson:
Tested on Gauss

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Enable remote write receiver.
+- Add prometheus remote write ingress boilerplate.
+- Add prometheus agent remote write secret boilerplate.
+- Generate prometheus remote write agent secret and config.
```
mcbenjemaa, line 15: Those must be rebased after merging #1000

mcbenjemaa: Please consider merging #1000  first.

mcbenjemaa: it makes me a little bit confused, as it already includes changes of #1000

service/controller/resource/monitoring/remotewriteingress/v1/resource.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return metav1.ObjectMeta{}, microerror.Mask(err)
+	}
+
+	return metav1.ObjectMeta{
+		Name:      fmt.Sprintf("prometheus-%s-remote-write", key.ClusterID(cluster)),
+		Namespace: key.Namespace(cluster),
+		Labels:    key.PrometheusLabels(cluster),
+		Annotations: map[string]string{
+			"nginx.ingress.kubernetes.io/auth-type":   "basic",
```
mcbenjemaa, line 65: Can you have a function that generates those annotation for basic auth

service/controller/resource/monitoring/remotewriteingress/v1/resource.go
```diff
@@ -0,0 +1,146 @@
...
+				{
+					Host: config.BaseDomain,
+					IngressRuleValue: networkingv1.IngressRuleValue{
+						HTTP: &networkingv1.HTTPIngressRuleValue{
+							Paths: []networkingv1.HTTPIngressPath{
+								{
+									Path: fmt.Sprintf("/%s/api/v1/write", key.ClusterID(cluster)),
+									Backend: networkingv1.IngressBackend{
+										Service: &networkingv1.IngressServiceBackend{
+											Name: "prometheus-operated",
```
mcbenjemaa, line 122: Can you put the name of service as constant.

service/controller/resource/monitoring/remotewriteingress/v1beta1/resource.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return metav1.ObjectMeta{}, microerror.Mask(err)
+	}
+
+	return metav1.ObjectMeta{
+		Name:      fmt.Sprintf("prometheus-%s-remote-write", key.ClusterID(cluster)),
+		Namespace: key.Namespace(cluster),
+		Labels:    key.PrometheusLabels(cluster),
+		Annotations: map[string]string{
+			"nginx.ingress.kubernetes.io/auth-type":   "basic",
```
mcbenjemaa, line 66: Same, please put those annotation in a function.

service/controller/resource/monitoring/remotewriteingress/v1beta1/resource.go
```diff
@@ -0,0 +1,139 @@
...
+			Rules: []networkingv1beta1.IngressRule{
+				{
+					Host: config.BaseDomain,
+					IngressRuleValue: networkingv1beta1.IngressRuleValue{
+						HTTP: &networkingv1beta1.HTTPIngressRuleValue{
+							Paths: []networkingv1beta1.HTTPIngressPath{
+								{
+									Path: fmt.Sprintf("/%s/api/v1/write", key.ClusterID(cluster)),
+									Backend: networkingv1beta1.IngressBackend{
+										ServiceName: "prometheus-operated",
```
mcbenjemaa, line 119: Same remark for name of the service!

CHANGELOG.md
QuentinBisson: Because it's not possible to have them separate and this saves me time to wait for reviews 😁

service/controller/resource/monitoring/remotewriteingress/v1/resource.go
QuentinBisson: Can you explain why? Considering I'm planning on removing v1beta1, I'd rather not :)

service/controller/resource/monitoring/remotewriteingress/v1/resource.go
QuentinBisson: Sure

service/controller/resource/monitoring/remotewriteingress/v1/resource.go
mcbenjemaa: I don’t like hardcoded things like this!

QuentinBisson: TBH, hardcoding it here or in a function is the same to me :D

QuentinBisson: Done

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
```diff
@@ -0,0 +1,117 @@
...
+	return r, nil
+}
+
+func getObjectMeta(ctx context.Context, v interface{}) (metav1.ObjectMeta, error) {
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return metav1.ObjectMeta{}, microerror.Mask(err)
+	}
+
+	return metav1.ObjectMeta{
```
mcbenjemaa, line 60: Don't we need to add some labels?

service/controller/resource/monitoring/remotewritesecret/resource.go
```diff
@@ -0,0 +1,81 @@
+package remotewritesecret
```
mcbenjemaa, line 1: I'm a bit confused 😕 
Why this package? `remotewritesecret` while the logic is added in `remotewriteagentsecret`

QuentinBisson: I think this one needs to go 😓 
Too many refactoring between the first and the last iteration

service/controller/resource/monitoring/remotewriteagentconfigsecret/doc.go
```diff
@@ -0,0 +1,4 @@
+package remotewriteagentconfigsecret
+
+// This packages creates the remote write agent secret config
+// to be used as values by the prometheus agent app to configure the remote write feature.
```
TheoBrigitte, line 4: If I got this correctly in this package we are wrapping the `RemoteWrite` config in a secret which will then be used by prometheus-operator to configure the agent. Is this right ?

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
mcbenjemaa: Do we really need to add labels like. 
```
clusterID
organization
Installation 
etc? 
```

mcbenjemaa:
@QuentinBisson some changes are included here https://github.com/giantswarm/prometheus-meta-operator/pull/1000
?

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: I don't see which other labels to set tbh

QuentinBisson: So i'd rather have suggestions :)

QuentinBisson:
Yes I already explained that in the previous PRs :D

QuentinBisson:
#1001 is reliant on #1000 so #1000 needs to be merged first

mcbenjemaa:
I will move my reviews notes to #1000

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	cluster, err := key.ToCluster(v)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	remoteWrites := []pov1.RemoteWriteSpec{
```
TheoBrigitte, line 78: Why is this an array?

service/controller/resource/monitoring/remotewriteagentconfigsecret/test/case-1-awsconfig.golden
```diff
@@ -0,0 +1,12 @@
+data:
+  values: LSB1cmw6IGh0dHBzOi8vaHR0cHM6Ly9wcm9tZXRoZXVzL2FsaWNlL2FwaS92MS93cml0ZQogIG5hbWU6ICIiCiAgc2VuZGV4ZW1wbGFyczogbnVsbAogIHJlbW90ZXRpbWVvdXQ6ICIiCiAgaGVhZGVyczoge30KICB3cml0ZXJlbGFiZWxjb25maWdzOiBbXQogIG9hdXRoMjogbnVsbAogIGJhc2ljYXV0aDoKICAgIHVzZXJuYW1lOgogICAgICBsb2NhbG9iamVjdHJlZmVyZW5jZToKICAgICAgICBuYW1lOiBhZ2VudC1yZW1vdGUtd3JpdGUKICAgICAga2V5OiB1c2VybmFtZQogICAgICBvcHRpb25hbDogbnVsbAogICAgcGFzc3dvcmQ6CiAgICAgIGxvY2Fsb2JqZWN0cmVmZXJlbmNlOgogICAgICAgIG5hbWU6IGFnZW50LXJlbW90ZS13cml0ZQogICAgICBrZXk6IHBhc3N3b3JkCiAgICAgIG9wdGlvbmFsOiBudWxsCiAgYmVhcmVydG9rZW46ICIiCiAgYmVhcmVydG9rZW5maWxlOiAiIgogIGF1dGhvcml6YXRpb246IG51bGwKICBzaWd2NDogbnVsbAogIHRsc2NvbmZpZzogbnVsbAogIHByb3h5dXJsOiAiIgogIHF1ZXVlY29uZmlnOiBudWxsCiAgbWV0YWRhdGFjb25maWc6IG51bGwK
```
TheoBrigitte, line 2: We do not really see anything here. This does not bring much value.

pkg/password/manager.go
```diff
@@ -0,0 +1,33 @@
...
+func (m SimpleManager) GeneratePassword(length int) (string, error) {
+	bytes := make([]byte, length)
+	if _, err := rand.Read(bytes); err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(bytes), nil
+}
+
+func (m SimpleManager) Hash(plaintext string) (string, error) {
+	bytes, err := bcrypt.GenerateFromPassword([]byte(plaintext), 14)
```
TheoBrigitte, line 28: Why are we hashing the password through bcrypt here ?

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
```diff
@@ -0,0 +1,102 @@
...
+	hashedPassword, err := config.PasswordManager.Hash(password)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	secret := &corev1.Secret{
+		ObjectMeta: objectMeta,
+		Data: map[string][]byte{
+			// To be used by the ingress basic auth
+			"auth":     []byte(fmt.Sprintf("%s:%s", username, hashedPassword)),
```
TheoBrigitte, line 91: Can we leave this magic to the client/prometheus-agent ? I would rather that we provider username and password and the client use them as he wants to.

Also why are we sending a hashed password here?

service/controller/resource/monitoring/remotewriteagentsecret/test/case-1-awsconfig.golden
```diff
@@ -0,0 +1,14 @@
+data:
+  auth: YWxpY2U6ZW5jcnlwdGVk
+  password: cGFzc3dvcmQ=
+  username: YWxpY2U=
```
TheoBrigitte, line 4: again we do not see much what is being produced here, this does not bring much value.

service/key/key.go
```diff
@@ -17,6 +17,9 @@ const (
 	monitoring = "monitoring"
 
 	PrometheusMemoryLimitCoefficient float64 = 1.2
+	PrometheusServiceName                    = "prometheus-operated"
+	RemoteWriteAgentConfigSecretName string  = "agent-remote-write-config"
+	RemoteWriteAgentSecretName       string  = "agent-remote-write"
```
TheoBrigitte, line 22: In general I would prefer to avoid being too specific. We are just configuring a remote write endpoint here. We do not need to mention the `agent` thing here. It could be consumed by a standard Prometheus also, and this might changes. Let's avoid being too specific please. This is also valid for package names.

TheoBrigitte (changes requested):
Good job overall :+1:

service/controller/resource/monitoring/remotewriteagentconfigsecret/doc.go
QuentinBisson: We could eventually fill this with the remotewrite CR as well

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: Because the prometheus agent app expects an array as we can use it to send remote write data to customers in the end

service/controller/resource/monitoring/remotewriteagentconfigsecret/test/case-1-awsconfig.golden
QuentinBisson: What would you like to have here? No tests?

pkg/password/manager.go
QuentinBisson: 1: Because it needs to be hashed
2: because nginx does not support a lot of hashes
3: because bcrypt is one of the best security hashes out there

service/controller/resource/monitoring/remotewriteagentsecret/resource.go
QuentinBisson: No we cannot because this auth parameter is used by the ingress on the management cluster and the password needs to be hashed for nginx to work

service/controller/resource/monitoring/remotewriteagentsecret/test/case-1-awsconfig.golden
QuentinBisson: I'd rather have minimal tests than no tests at all

service/key/key.go
QuentinBisson: Except we are not just configuring a remote write endpoint, we are configuring it for the agent. We already have remotewritesecret and remotewriteconfig as resources anyway so we need to avoid collision between the 2

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
QuentinBisson: I would rather we hare too broad here instead of too narrow

QuentinBisson:
@TheoBrigitte can you check my answer to your comments?

service/controller/resource/monitoring/remotewriteagentconfigsecret/resource.go
TheoBrigitte: Ok good.

service/controller/resource/monitoring/remotewriteagentconfigsecret/test/case-1-awsconfig.golden
TheoBrigitte: Having the content of the values would be enough I guess.

pkg/password/manager.go
TheoBrigitte: I though basic auth where just base64 encoded, maybe I am wrong.

service/key/key.go
TheoBrigitte: Maybe something like `incoming-remote-write` ? `agent` feels wrong here, this is just a mode used to run Prometheus.

service/controller/resource/monitoring/remotewriteagentsecret/test/case-1-awsconfig.golden
TheoBrigitte: But the test should produce output that we can compare against input in code.

pkg/password/manager.go
QuentinBisson: Sadly no: https://docs.giantswarm.io/advanced/ingress/configuration/#authentication It requires a hash of the password. In the simplest example, it uses apache `htpasswd` but there is no go implementation to cypher the password so I picked the best in go that nginx supports

service/controller/resource/monitoring/remotewriteagentconfigsecret/test/case-1-awsconfig.golden
QuentinBisson: That's encoded in base 64 by the secret creation mechanism. The code will not allow us to generate non base64 secret just for the tests

QuentinBisson:
@TheoBrigitte apart from the secret encoded in base64, the rest should be addressed :)

service/controller/resource/monitoring/remotewriteagentconfigsecret/test/case-1-awsconfig.golden
TheoBrigitte: Maybe we can have a totally different test that just test the part that generate the values before the secret then.

pkg/password/manager.go
TheoBrigitte: Ok fine

TheoBrigitte (approved):
I would like to see better tests output but I won't block this longer.

QuentinBisson:
Let's improve the tests once we have a working feature :)

PR #1007, opened 2022-09-30
Define custom Prometheus Storage size based on annotation value

mcbenjemaa:
Signed-off-by: Mohamed Chiheb <mohamed@giantswarm.io>

Towards: https://github.com/giantswarm/giantswarm/issues/23670

Define custom Prometheus Storage size based on annotation `monitoring.giantswarm.io/prometheus-volume-size`

## Checklist

I have:


mcbenjemaa:
TODO: 

- [ ] When this value changed: reconcile the prometheus CR

service/key/key.go
```diff
@@ -298,3 +299,18 @@ func IngressClassName() string {
...
 }
+
+// PrometheusDiskSize returns the desired disk size based on the
+// value of annotation monitoring.giantswarm.io/prometheus-disk-size
+func PrometheusDiskSize(value, defaulVal string) string {
+	switch value {
+	case "small":
+		return "30Gi"
+	case "medium":
+		return defaulVal
```
hervenicol, line 310: I don't much like that `small` and `large` values are defined there, but not `medium`.
Couldn't we have all `label=size` definitions in one place?

QuentinBisson: I agree. I would create a struct passed to the config instead of adding another function in key.
Maybe something like PrometheusSize.

It would have a New method that would set medium to config.PrometheusSize

mcbenjemaa: I didn't' fully understand

pkg/pvresizing/pv.go
```diff
@@ -0,0 +1,27 @@
+package pvresizing
+
+const (
+	Name = "prometheus"
```
QuentinBisson, line 4: Why is that here? :D

pkg/pvresizing/pv.go
```diff
@@ -0,0 +1,27 @@
...
+type PrometheusStorageSizeType string
+
+// PrometheusDiskSize returns the desired disk size based on the
+// value of annotation monitoring.giantswarm.io/prometheus-disk-size
+func PrometheusDiskSize(v string) string {
+	value := PrometheusStorageSizeType(v)
+	switch value {
+	case PrometheusStorageSizeSmall:
+		return "30Gi"
+	case PrometheusStorageSizeMedium:
```
QuentinBisson, line 20: I think you can remove the medium case here, it's redundant

pkg/pvresizing/pv.go
```diff
@@ -0,0 +1,27 @@
...
+
+	PrometheusStorageSizeSmall  PrometheusStorageSizeType = "small"
+	PrometheusStorageSizeMedium PrometheusStorageSizeType = "medium"
+	PrometheusStorageSizeLarge  PrometheusStorageSizeType = "large"
+)
+
+type PrometheusStorageSizeType string
+
+// PrometheusDiskSize returns the desired disk size based on the
+// value of annotation monitoring.giantswarm.io/prometheus-disk-size
```
QuentinBisson, line 14: Can you not reference the annotation here? This is infrastructure level information that has no reason to be in the domain

pkg/pvresizing/pv.go
```diff
@@ -0,0 +1,27 @@
...
+func PrometheusDiskSize(v string) string {
+	value := PrometheusStorageSizeType(v)
```
QuentinBisson, line 16: Can you name v something more meaningful? Also, documenting what this parameter is makes sense :)

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -118,7 +113,7 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 
 	var storage promv1.StorageSpec
 	diskSizeAnnotationValue := cluster.GetAnnotations()[key.PrometheusDiskSizeAnnotation]
-	diskSize := prometheusDiskSize(prometheusStorageSizeType(diskSizeAnnotationValue))
+	diskSize := pvresizing.PrometheusDiskSize(diskSizeAnnotationValue)
```
QuentinBisson, line 116: I think we should rename disk to volume though as volume is the kubernetes term

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	r.logger.Debugf(ctx, "ensuring pv resizing")
+	{
+		cluster, err := key.ToCluster(obj)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		namespace := fmt.Sprintf("%s-%s", cluster.GetName()+"-prometheus")
```
QuentinBisson, line 26: I'm pretty sure we have key.Namespace(cluster) or similar :)

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+		pvc, err := r.fetchPVC(ctx, cluster.GetName(), namespace)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		currentValue := pvc.Spec.Resources.Requests.Storage().String()
+		anoVal := cluster.GetAnnotations()[key.PrometheusDiskSizeAnnotation]
```
QuentinBisson, line 33:
```suggestion
		desiredPVSize := cluster.GetAnnotations()[key.PrometheusDiskSizeAnnotation]
```

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+			return microerror.Mask(err)
+		}
+
+		namespace := fmt.Sprintf("%s-%s", cluster.GetName()+"-prometheus")
+		pvc, err := r.fetchPVC(ctx, cluster.GetName(), namespace)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		currentValue := pvc.Spec.Resources.Requests.Storage().String()
```
QuentinBisson, line 32:
```suggestion
		currentPVSize := pvc.Spec.Resources.Requests.Storage().String()
```

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+		anoVal := cluster.GetAnnotations()[key.PrometheusDiskSizeAnnotation]
+		desireDiskSize := pvresizing.PrometheusDiskSize(anoVal)
```
QuentinBisson, line 34:
```suggestion
		desiredDiskSize := pvresizing.PrometheusDiskSize(anoVal)
```

service/controller/resource/monitoring/pvresizingresource/create.go
QuentinBisson, line 33: configuredPVSizeAnnotation maybe

service/controller/resource/monitoring/pvresizingresource/create.go
QuentinBisson, line 34:
```suggestion
		desiredVolumeSize := pvresizing.PrometheusDiskSize(anoVal)
```

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+
+		// Check the value of annotation with the current value in PVC.
+		if currentValue != desireDiskSize {
+			// if different we can trigger an update
```
QuentinBisson, line 38: I would comment something more along the lines of:
Resizing requested. Following the procedude described here https:// until stateful set resizing made it into kubernetes (link the KEP for it)

service/controller/resource/monitoring/pvresizingresource/delete.go
```diff
@@ -0,0 +1,11 @@
+package pvresizingresource
+
+import (
+	"context"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	// TODO Check if we need delete
```
QuentinBisson, line 8: We do not really need to delete anything as the PV get's deleted by the prometheus operator, you can add a comment saying it and remove the TODO

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+		}
+
+	}
+	r.logger.Debugf(ctx, "ensured pv resizing")
+
+	return nil
+}
+
+func (r *Resource) fetchPVC(ctx context.Context, clusterID, namespace string) (*corev1.PersistentVolumeClaim, error) {
```
QuentinBisson, line 52: PVC are named prometheus-<cluster_id>-db-prometheus-<cluster_id>-0. I'm wondering if we should not use a Get instead.
If you'd rather use a list, I would actually return a list of PVC (in case we move to sharding later) than return the first object without checking if there are more than 1 in the slice

QuentinBisson, line 52:
```suggestion
func (r *Resource) listPVCs(ctx context.Context, clusterID, namespace string) (*corev1.PersistentVolumeClaim, error) {
```

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+	l := labels.SelectorFromSet(labels.Set(map[string]string{
```
QuentinBisson, line 53: Can this map be a const at the top?

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+		LabelSelector: l.String(),
+	}
+	list, err := r.k8sClient.K8sClient().CoreV1().PersistentVolumeClaims(namespace).List(ctx, options)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	if len(list.Items) == 0 {
+		return nil, fmt.Errorf("couldn't retrieve PVC")
+	}
+	return &list.Items[0], err
```
QuentinBisson, line 70: As I stated earlier, either return the whole slice of check that there is only one

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (r *Resource) updatePrometheusDiskSize(ctx context.Context, clusterID, namespace, desiredValue string, pvc *corev1.PersistentVolumeClaim) error {
```
QuentinBisson, line 73:
```suggestion
func (r *Resource) resize(ctx context.Context, clusterID, namespace, desiredValue string, pvc *corev1.PersistentVolumeClaim) error {
```

QuentinBisson: desiredValue => desiredVolumeSize

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+
+	prometheus, err := r.prometheusClient.MonitoringV1().Prometheuses(namespace).Get(ctx, clusterID, metav1.GetOptions{})
```
QuentinBisson, line 75: I think you can get the namespace and clusterID from the PVC instead of passing them in the function though

QuentinBisson: You can probaly go with context, pvc, desiredVolumeSize as the only function arguments

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	newSize := resource.MustParse(desiredValue)
```
QuentinBisson, line 80:
```suggestion
	newVolumeSize := resource.MustParse(desiredValue)
```

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+	// Updating Prometheus CR Storage size if it doesn't match with the desiredValue
+	if !(prometheus.Spec.Storage.VolumeClaimTemplate.Spec.Resources.Requests.Storage().String() == desiredValue) {
```
QuentinBisson, line 82: You are doing this check a bit earlier, why do it once again?

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+	// Patch PVC with the new size
+	pvc.Spec.Resources.Requests["Storage"] = newSize
+	patch := []byte(fmt.Sprintf(`{"spec": { "resources": { "requests": { "storage": "%v" } } } }`, newSize))
+	_, err = r.k8sClient.K8sClient().CoreV1().PersistentVolumeClaims(namespace).
+		Patch(ctx, pvc.GetName(), types.StrategicMergePatchType, patch, metav1.PatchOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// Delete the STS to update its definition, the recreation is done right away by the Prometheus operator
```
QuentinBisson, line 99: Delete the sts without the PVC (using orphan)

service/controller/resource/monitoring/pvresizingresource/create.go
QuentinBisson: link of the kep https://github.com/kubernetes/enhancements/issues/661

service/controller/resource/monitoring/pvresizingresource/create.go
QuentinBisson: Actually, reading point 2, https://github.com/prometheus-operator/prometheus-operator/issues/4079#issuecomment-1211989005 I'd return a slice

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+	prometheus, err := r.prometheusClient.MonitoringV1().Prometheuses(namespace).Get(ctx, clusterID, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	newSize := resource.MustParse(desiredValue)
+	// Updating Prometheus CR Storage size if it doesn't match with the desiredValue
+	if !(prometheus.Spec.Storage.VolumeClaimTemplate.Spec.Resources.Requests.Storage().String() == desiredValue) {
+		prometheus.Spec.Storage.VolumeClaimTemplate.Spec.Resources.Requests["Storage"] = newSize
+		_, err = r.prometheusClient.MonitoringV1().Prometheuses(namespace).Update(ctx, prometheus, metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
QuentinBisson: No ref to Prometheus here would make it simpler to use :)
Also, the prometheus resource would override this value

service/controller/resource/monitoring/pvresizingresource/create.go
```diff
@@ -0,0 +1,108 @@
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	// Patch PVC with the new size
+	pvc.Spec.Resources.Requests["Storage"] = newSize
```
QuentinBisson, line 91: We might want to ensure the PVC is not resizing already (checking it's status) but we can do it if this become an issue

pkg/pvresizing/pv.go
```diff
@@ -0,0 +1,27 @@
...
+	PrometheusStorageSizeSmall  PrometheusStorageSizeType = "small"
+	PrometheusStorageSizeMedium PrometheusStorageSizeType = "medium"
+	PrometheusStorageSizeLarge  PrometheusStorageSizeType = "large"
+)
+
+type PrometheusStorageSizeType string
+
+// PrometheusDiskSize returns the desired disk size based on the
+// value of annotation monitoring.giantswarm.io/prometheus-disk-size
+func PrometheusDiskSize(v string) string {
```
QuentinBisson, line 15: I know it's trivial but can you add a test file?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Define custom Prometheus storage size based on an annotation value
```
QuentinBisson, line 12: Can you define the ann
```suggestion
- Customize Prometheus volume size based via the ... annotation (cf. documentation somewhere probably public doc)
```

service/controller/resource/monitoring/pvresizingresource/error.go
```diff
@@ -0,0 +1 @@
+package pvresizingresource
```
QuentinBisson, line 1: It does not look like it's needed. Can you make it a doc.go file and explain what this package is for instead?

service/controller/resource/monitoring/pvresizingresource/resource.go
```diff
@@ -0,0 +1,37 @@
+package pvresizingresource
```
QuentinBisson, line 1: Are we resizing the PV or the PVC 😅 I know I gave you that naming idea, my apologies

service/controller/resource/monitoring/pvresizingresource/resource.go
```diff
@@ -0,0 +1,37 @@
...
+	k8sClient        k8sclient.Interface
+	logger           micrologger.Logger
+	prometheusClient promclient.Interface
+}
+
+func New(config Config) (*Resource, error) {
+	r := &Resource{
+		k8sClient:        config.K8sClient,
+		logger:           config.Logger,
+		prometheusClient: config.PrometheusClient,
```
QuentinBisson, line 29: Unneeded

service/service.go
```diff
@@ -197,7 +196,6 @@ func New(config Config) (*Service, error) {
 				PrometheusAddress:           config.Viper.GetString(config.Flag.Service.Prometheus.Address),
 				PrometheusBaseDomain:        config.Viper.GetString(config.Flag.Service.Prometheus.BaseDomain),
 				PrometheusCreatePVC:         config.Viper.GetBool(config.Flag.Service.Prometheus.Storage.CreatePVC),
-				PrometheusStorageSize:       config.Viper.GetString(config.Flag.Service.Prometheus.Storage.Size),
```
QuentinBisson, line 200: You can remove it from the configmap https://github.com/giantswarm/prometheus-meta-operator/blob/144cc87bb8a47efec40d7ab3a575b0c1d3eacd9e/helm/prometheus-meta-operator/templates/configmap.yaml#L51 and from the config repo (apart for KVM installations)

pkg/pvresizing/pv.go
QuentinBisson, line 4:
```suggestion
```

service/controller/resource/monitoring/pvresizingresource/create.go
mcbenjemaa: I prefer to list PVC with the selector,

mcbenjemaa: I'm actually returning the first object in the list. thats why i named `fetchPVC`

service/controller/resource/monitoring/pvresizingresource/create.go
mcbenjemaa: it can but? i prefer to have as function instead. because we need to supply the clusterID

QuentinBisson: sure

service/controller/resource/monitoring/pvresizingresource/create.go
mcbenjemaa: No this check is for Prometheus CR, as it is changed in the prometheus resource.
But, I'm afraid if the Value is still not applied to prometheus CR when this reconciliation loop has triggered.

service/controller/resource/monitoring/pvresizingresource/create.go
QuentinBisson: Yes but you should actually return the list as we can have shard hence multiple pvc

QuentinBisson: Also fetch is used in Javascript mostly, Get is the term used in go

service/controller/resource/monitoring/pvresizingresource/create.go
QuentinBisson: I think we should ensure the prometheus resource runs before this one and we are fine
I'd rather not add an extra dependency here

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,150 @@
...
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	r.logger.Debugf(ctx, "ensuring pvc resizing")
+	{
+		cluster, err := key.ToCluster(obj)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		namespace := key.Namespace(cluster)
+		pvcList, err := r.listPVC(ctx, cluster.GetName(), namespace)
```
QuentinBisson, line 27:
```suggestion
		pvcList, err := r.listPVCs(ctx, cluster.GetName(), namespace)
```

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,150 @@
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		for _, pvc := range pvcList {
+			fmt.Println("PVC Name", pvc.GetName())
```
QuentinBisson, line 33:
```suggestion
```
A logger or nothing you will use :D

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,150 @@
...
+				// https://github.com/prometheus-operator/prometheus-operator/issues/4079#issuecomment-1211989005
+				// until stateful set resizing made it into kubernetes:
+				// https://github.com/kubernetes/enhancements/issues/661
+				fmt.Println("resize..........", pvc.GetName())
+				err = r.resize(ctx, desiredVolumeSize, pvc)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+			} else if currentPVCSize > desiredPVCSize {
+				// Since Resizing to lower storage is forbidden
```
QuentinBisson, line 53:
```suggestion
				// Since downsizing a volume is forbidden, we have to replace the PVC and the STS, causing a data loss
```

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,150 @@
...
+	return labels.SelectorFromSet(labels.Set(map[string]string{
+		"app.kubernetes.io/instance":   clusterID,
+		"app.kubernetes.io/managed-by": "prometheus-operator",
+		"app.kubernetes.io/name":       "prometheus",
+		"operator.prometheus.io/name":  clusterID,
+		"prometheus":                   clusterID,
+	}))
+}
+
+func (r *Resource) listPVC(ctx context.Context, clusterID, namespace string) ([]corev1.PersistentVolumeClaim, error) {
```
QuentinBisson, line 79:
```suggestion
func (r *Resource) listPVCs(ctx context.Context, clusterID, namespace string) ([]corev1.PersistentVolumeClaim, error) {
```

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,150 @@
...
+		Patch(ctx, pvc.GetName(), types.StrategicMergePatchType, patch, metav1.PatchOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	err = r.k8sClient.K8sClient().CoreV1().PersistentVolumeClaims(namespace).
+		Delete(ctx, pvc.GetName(), metav1.DeleteOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	fmt.Println("PVC deleted.....")
```
QuentinBisson, line 138: You might have to delete the PV?

service/key/key.go
```diff
@@ -20,6 +20,7 @@ const (
 	PrometheusServiceName                    = "prometheus-operated"
 	RemoteWriteAgentConfigSecretName string  = "agent-remote-write-config"
 	RemoteWriteAgentSecretName       string  = "agent-remote-write"
+	PrometheusDiskSizeAnnotation     string  = "monitoring.giantswarm.io/prometheus-volume-size"
```
QuentinBisson, line 23:
```suggestion
	PrometheusVolumeSizeAnnotation     string  = "monitoring.giantswarm.io/prometheus-volume-size"
```

service/controller/resource/monitoring/pvcresizingresource/create.go
marieroque: PVC has StorageClass gp2 defined, so the  PV provisioning and deleting is dynamic

service/controller/resource/monitoring/pvcresizingresource/create.go
QuentinBisson: Can you remove the printlns?

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -30,21 +30,16 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		}
 
 		for _, pvc := range pvcList {
-			fmt.Println("PVC Name", pvc.GetName())
 			currentPVCSize := resource.MustParse(pvc.Spec.Resources.Requests.Storage().String())
 			desiredPVCSize := cluster.GetAnnotations()[key.PrometheusVolumeSizeAnnotation]
```
QuentinBisson, line 34: Looking really good but can you add some logs in this resource?

mcbenjemaa: which one?

QuentinBisson: A bit of debug mostly :) something like resizing, changing pvc and so on

QuentinBisson: I'd go with the following on every code branch

// resizinv PVC xxx
resizing code...
// Resizing succeeded/failed

mcbenjemaa: can you take a look

QuentinBisson: Nice

QuentinBisson (approved):
LGTM when testing is done

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,142 @@
...
+				r.logger.Debugf(ctx, "resizing PVC %v", pvc.GetName())
+				// Resizing requested. Following the procedure described here:
+				// https://github.com/prometheus-operator/prometheus-operator/issues/4079#issuecomment-1211989005
+				// until stateful set resizing made it into kubernetes:
+				// https://github.com/kubernetes/enhancements/issues/661
+				err = r.resize(ctx, desiredVolumeSize, pvc)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+				r.logger.Debugf(ctx, "resizing PVC %v succeeded", pvc.GetName())
```
QuentinBisson, line 48:
```suggestion
				r.logger.Debugf(ctx, "resized PVC %v", pvc.GetName())
```

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,142 @@
...
+			// Check the value of annotation with the current value in PVC.
+			if currentPVCSize.Value() < desiredVolumeSize.Value() {
+				r.logger.Debugf(ctx, "resizing PVC %v", pvc.GetName())
+				// Resizing requested. Following the procedure described here:
+				// https://github.com/prometheus-operator/prometheus-operator/issues/4079#issuecomment-1211989005
+				// until stateful set resizing made it into kubernetes:
+				// https://github.com/kubernetes/enhancements/issues/661
+				err = r.resize(ctx, desiredVolumeSize, pvc)
+				if err != nil {
+					return microerror.Mask(err)
```
QuentinBisson, line 46: Can you add an error log here?

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,142 @@
...
+			} else if currentPVCSize.Value() > desiredVolumeSize.Value() {
+				r.logger.Debugf(ctx, "replacing PVC %v", pvc.GetName())
+				// Since downsizing a volume is forbidden, we have to replace the PVC and the STS, causing a data loss
+				// Therefore, we replace the PVC and STS
+				// But this will cause data loss
+				err = r.replacePVC(ctx, pvc)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+				r.logger.Debugf(ctx, "replacing PVC %v succeeded", pvc.GetName())
```
QuentinBisson, line 58:
```suggestion
				r.logger.Debugf(ctx, "replaced PVC %v", pvc.GetName())
```

service/controller/resource/monitoring/pvcresizingresource/create.go
```diff
@@ -0,0 +1,142 @@
...
+				}
+				r.logger.Debugf(ctx, "resizing PVC %v succeeded", pvc.GetName())
+			} else if currentPVCSize.Value() > desiredVolumeSize.Value() {
+				r.logger.Debugf(ctx, "replacing PVC %v", pvc.GetName())
+				// Since downsizing a volume is forbidden, we have to replace the PVC and the STS, causing a data loss
+				// Therefore, we replace the PVC and STS
+				// But this will cause data loss
+				err = r.replacePVC(ctx, pvc)
+				if err != nil {
+					return microerror.Mask(err)
```
QuentinBisson, line 56: Same can you add an error log?

QuentinBisson (approved):

PR #1022, opened 2022-10-19
Configure working queueconfig

QuentinBisson:
After some experimentation, it seems that either nginx or the prometheus receiver are not able to accept too many requests and setting the number of maxShards to 10 plus an increased capacity (more samples stored per shards and more samples sent per batch) does the trick.

Remote write is not falling behind and the situation appears stable:

image

image

For information:
- [MaxShards](https://prometheus.io/docs/practices/remote_write/#max_shards): We currently use only 1 shards (we have only 5 targets so ...) and it's not increasing. I set it to 10 because it can take a bit more when replaying the WAL but it's mostly fine
- [Capacity](https://prometheus.io/docs/practices/remote_write/#capacity): Set to 30000 as it should be 3-10 times the max number of sent samples
- [MaxSamplesPerSecond](https://prometheus.io/docs/practices/remote_write/#max_samples_per_send): Sending 10K works. maybe we can reduce that value but remote write is not falling behind so far

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add scraping of Cilium on Management Clusters.
 
+### Changed
+
+- Configure working queue config for the remote write receiver (reducing max number of shards but increasing sample capacity).
+
 ## [4.6.4] - 2022-10-17
 
 ### Changed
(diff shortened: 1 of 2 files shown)


mcbenjemaa (approved):
LGTM 
If tested

QuentinBisson:
I'm pretty sure the graphs show that it is working tested and working 😅

PR #1040, opened 2022-10-31
Cleanup rw old secrets

mcbenjemaa:
Towards https://github.com/giantswarm/giantswarm/issues/24222

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


TheoBrigitte (approved):
`Logger` is actually never used and could be removed to simplify things. But other than that fine with me.

QuentinBisson (commented):
Does that mean we can remove this resource entirely now if it was deployed everywhere?

mcbenjemaa:
PMO was released,  which means next we can merge #1040 that will cleanup the actual code.

QuentinBisson:
LOVE IT

PR #1048, opened 2022-11-09
Fix CAPI MC reconciliation

QuentinBisson:
Fix reconciliation bug to ensure the clusterapi controller only reconciles cluster api workload clusters as it currently reconciles both MC and WC cluster CR.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix a reconciliation bug on CAPI MC that were reconciled twice.
+- Ensure the remote write endpoint configuration is enabled for MCs as well
+
 ### Removed
 
 - Old remotewrite secret
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


QuentinBisson:
@giantswarm/team-atlas can I get a review?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix a reconciliation bug on CAPI MC that were reconciled twice.
+- Ensure the remote write endpoint configuration is enabled for MCs as well
```
TheoBrigitte, line 13: I see we are fixing 2 different problems here I would prefer to have those in 2 separate PRs. It ease up reviews and makes reverting changes in case of problem easier.

service/controller/clusterapi/controller.go
```diff
@@ -65,6 +66,12 @@ func NewController(config ControllerConfig) (*Controller, error) {
 		}
 	}
 
+	// We ensure the cluster-api controller is not checking reconciling the MC cluster CR to avoid conflicts.
```
TheoBrigitte, line 69: We could use a more elaborated comment here

```suggestion
	// Ensure cluster-api controller is not reconciling the MC cluster CR to avoid duplicate reconciliation with the management-cluster controller.
	// This due to CAPI installation providing a cluster CR for the MC.
```

TheoBrigitte (changes requested):

TheoBrigitte (approved):
Fine with me when tested against CAPI and non-capi MC

TheoBrigitte (approved):
Nice, thanks for splitting

PR #1054, opened 2022-11-15
Enable remote write receiver for MC prometheus

QuentinBisson:
## Checklist

This change is being introduced because CAPI MC clusters will be running the agent as default apps are installing the observability bundle everywhere.

Enabling the remote write receiver on the MC prometheus will ensure we can use service monitors on both CAPI MC and WC and remove the additional scrape config section.

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


QuentinBisson:
Tested multiple times. Spliitted from https://github.com/giantswarm/prometheus-meta-operator/pull/1048

TheoBrigitte (approved):
Nice, thanks for splitting

PR #1055, opened 2022-11-16
ADD Inhibition rule for prometheus-agent

mcbenjemaa:
towards https://github.com/giantswarm/giantswarm/issues/24121

Ref: https://github.com/giantswarm/prometheus-rules/pull/497


files/templates/alertmanager/alertmanager.yaml
```diff
@@ -425,3 +425,9 @@ inhibit_rules:
   target_matchers:
   - cancel_if_cluster_has_no_workers=true
   equal: [cluster_id]
+
+- source_matchers:
+    - release_before_v19=false
```
QuentinBisson, line 430:
```suggestion
    - vintage_release_before_v19=false
```

QuentinBisson, line 430:
```suggestion
    - vintage_cluster_release_before_v19=false
```

QuentinBisson:
Does this only takes into account vintage? Is it not set for CAPI clusters?

QuentinBisson (approved):

PR #1057, opened 2022-11-17
Drop job in favor of SM

ljakimczuk:
## Description

Towards: https://github.com/giantswarm/giantswarm/issues/24541

App Exporter provides Service Monitor that configures job that is redundant to the `CLUSTER-prometheus/app-exporter-CLUSTER/0`. The target can be  dropped from this job with the `giantswarm.io/monitoring` label set, but setting this label at the same time adds (keeps) target to the `CLUSTER-prometheus/workload-CLUSTER/0` job, still making the App Exporter to be scrapped twice. 

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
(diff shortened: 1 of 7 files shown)


hervenicol (approved):
Looking at the parent issue, it seems like you already ensured all MCs have a servicemonitor, so LGTM.

PR #1075, opened 2022-11-30
Clean up unused property

QuentinBisson:
This property is only used on KVM and the latest PMO will not work there due to the differences in Kubernetes versions. Cleaning up will help make this easier to maintain

## Checklist

I have:


helm/prometheus-meta-operator/templates/configmap.yaml
```diff
@@ -25,9 +25,6 @@ data:
         address: {{ .Values.alertmanager.address }}
         baseDomain: {{ .Values.alertmanager.host }}
         logLevel: {{ .Values.alertmanager.logLevel }}
-        storage:
-          createPVC: {{ and .Values.alertmanager.storage.enabled .Values.alertmanager.storage.createPVC }}
-          size: {{ .Values.alertmanager.storage.pvcSize }}
```
marieroque, line 30: are you sure you want to remove size field ?

QuentinBisson: Actually no I'm adding it back

TheoBrigitte (approved):
This is a breaking change but anyway KVM is far behind already.

PR #1082, opened 2022-11-30
Use prometheus-agent on CAPA MCs

mcbenjemaa:
towards: https://github.com/giantswarm/giantswarm/issues/24788

This will make PMO ignore static targets on CAPA MCs, and add namespaceSelector for Prometheus to ignore kube-system.

### DIFF
--- a/service/controller/resource/monitoring/scrapeconfigs/resource.go
+++ b/service/controller/resource/monitoring/scrapeconfigs/resource.go
@@ -237,6 +237,12 @@ func hasPrometheusAgent(ctx context.Context, ctrlClient client.Client, cluster m
 				return false, microerror.Mask(err)
 			}
 			return version.GTE(gcpAgentVersion), nil
+		case "capa":
+			gcpAgentVersion, err := semver.Parse("0.11.0")
+			if err != nil {
+				return false, microerror.Mask(err)
+			}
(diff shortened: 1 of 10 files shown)


service/controller/resource/monitoring/scrapeconfigs/resource.go
```diff
@@ -237,6 +237,12 @@ func hasPrometheusAgent(ctx context.Context, ctrlClient client.Client, cluster m
 				return false, microerror.Mask(err)
 			}
 			return version.GTE(gcpAgentVersion), nil
+		case "capa":
+			gcpAgentVersion, err := semver.Parse("0.11.0")
```
QuentinBisson, line 241: Can you rename that to capaVersion?

Let's check with @erkanerol if we have a working release for capv/vcd so we Can support all providers starting now

QuantumEnigmaa: I updated the variable's name to capaAgentVersion. Is that ok for you ?

QuentinBisson: That's good ;)

QuentinBisson (approved):

QuantumEnigmaa (approved):

PR #1085, opened 2022-12-02
Opsgenie messages: revert to markdown

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/20661

Revert to markdown so it's looking OK on slack

## Checklist

I have:


hervenicol:
Looking rather OK on opsgenie:
image

now, will it look good on slack also? Let's test!

hervenicol:
At least it's the same okay-ish on both :shrug: 
Definitely an improvement over current html version.

QuentinBisson (approved):

PR #1102, opened 2022-12-19
Ignore remotewrite feature in kube-system namespace.

mcbenjemaa:
Ignore remotewrite feature in kube-system namespace. #1102


pkg/remotewriteutils/util.go
```diff
@@ -41,5 +41,12 @@ func FetchPrometheusList(ctx context.Context, r *ResourceWrapper, rw *pmov1alpha
 		return nil, microerror.Maskf(errorFetchingPrometheus, "Could not fetch Prometheus with label selector %#q", rw.Spec.ClusterSelector.String())
 	}
 
+	for i, p := range prometheusList.Items {
```
QuentinBisson, line 44: Would it not be easier to change the label selector to skip the agent using the not operator and this label app.kubernetes.io/name: prometheus-agent ?

mcbenjemaa: it's clusterSelector and not prometheusSelector

QuentinBisson: Well that's how w ecall it but it's used to select prometheus is it not?

QuentinBisson:
```go
prometheusList, err := r.PrometheusClient.
		MonitoringV1().
		Prometheuses(metav1.NamespaceAll).
		List(ctx, metav1.ListOptions{LabelSelector: selector.String()})
```

mcbenjemaa: Hmmm, let me check!

QuentinBisson: LabelSelecto contains a MatchLabels and MatchExpressions :)

QuentinBisson: So we can add a MatchExpressions Not Equal app.kubernetes.io/name: prometheus-agent

mcbenjemaa: let's try

mcbenjemaa: works ✅

mcbenjemaa: can you take a look?

QuentinBisson (approved):
Looks way better than the loop to me :)

pkg/remotewriteutils/util.go
QuentinBisson: Nice

PR #1107, opened 2022-12-21
Add template for harbor slack alerts

William-Young-97:
Added the `harbor_implementation` receiver in order to send slack alerts to the `#harbor-implementation` channel based on harbor alerts.

Signed-off-by: William Young <will.young@engineerbetter.com>
Co-authored-by: Ajayi Dipo <ajayidipo@ymail.com>

### DIFF
--- a/files/templates/alertmanager/alertmanager.yaml
+++ b/files/templates/alertmanager/alertmanager.yaml
@@ -20,7 +20,7 @@ route:
 
   - receiver: harbor_implementation
     matchers:
-    - team="harbor"
+    - app="harbor"
     continue: false
 
   # Falco noise Slack
--- a/files/templates/alertmanager/alertmanager.yaml
+++ b/files/templates/alertmanager/alertmanager.yaml
@@ -17,6 +17,12 @@ route:
   receiver: root
 
   routes:
+
+  - receiver: harbor_implementation
+    matchers:
+    - team="harbor"
+    continue: false
+
   # Falco noise Slack
   - receiver: falco_noise_slack
     matchers:
@@ -114,6 +120,10 @@ route:
 
 receivers:
 - name: root
+
+- name: 'harbor_implementation'
+  slack_configs:
+  - channel: '#harbor-implementation'
 
 - name: falco_noise_slack
   slack_configs:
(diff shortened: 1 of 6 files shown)


files/templates/alertmanager/alertmanager.yaml
```diff
@@ -17,6 +17,12 @@ route:
   receiver: root
 
   routes:
+
+  - receiver: harbor_implementation
+    matchers:
+    - team="harbor"
```
mcbenjemaa, line 23: Why it's `team="harbor"` ?

files/templates/alertmanager/alertmanager.yaml
```diff
@@ -115,6 +121,10 @@ route:
 receivers:
 - name: root
 
+- name: 'harbor_implementation'
+  slack_configs:
```
mcbenjemaa, line 125: Do you want to customize the actions?

mcbenjemaa:
Execute 

```
go test ./service/controller/resource/alerting/alertmanagerconfig -update=true

```

To resolve the failing test...

files/templates/alertmanager/alertmanager.yaml
MarcelMue: This is simply for testing alerts which are being implemented by ContainerSolutions. No other internal team should get / own these currently. Does this work to just set harbor?

mcbenjemaa: it's kind of misleading configuration, We forward alerts based on Giant Swarm Teams.
So, I prefer if you can use another matcher.. instead of `team="harbor"`

William-Young-97: Will change it.

files/templates/alertmanager/alertmanager.yaml
William-Young-97: Unsure what to customise them with at this time.

mcbenjemaa: Like buttons, and so on,,

mcbenjemaa: if you don't needed, that's fine

mcbenjemaa (approved):
LGTM

files/templates/alertmanager/alertmanager.yaml
William-Young-97: Yeah it's ok for now I think in the future this would be good though

PR #1108, opened 2023-01-02
Remotewrite ingress allows bigger requests

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/25049

For prometheus-agent remotewrite to work fine, it has to send data in rather big batches.
We have observed that proper performance requires batches at minimum between 1 to 2 MB big.
And default nginx config limits requests to 1MB.

So let's increase this limit for remotewrite ingress.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- remotewrite ingress allows bigger requests
+
 ## [4.18.0] - 2022-12-19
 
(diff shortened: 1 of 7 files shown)


TheoBrigitte (approved):
LGTM

Also I know debugging was hard to come to those values. I think we mostly used some graphs around the http requests being made (status code, body size, delay, etc ...). I would suggest to add what we used either in a dashboard, an opsrecipe, or whatever feels right.

QuantumEnigmaa (approved):

hervenicol:
Good idea @TheoBrigitte.
Done here: https://github.com/giantswarm/giantswarm/pull/25187

PR #1109, opened 2023-01-02
prometheus-agent: increase max samples per send

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/25049

For prometheus-agent remotewrite to work fine, it has to send data in rather big batches.
The previous batch size was too small, let's increase it.

I have to understand the deployment procedure before merging it. I'll probably add stuff in the CHANGELOG when I test it.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - remotewrite ingress allows bigger requests
+- prometheus-agent: increase max samples per send. ⚠️ Warning: updates an immutable secret, will require manual actions at deployment.
 
 ## [4.18.0] - 2022-12-19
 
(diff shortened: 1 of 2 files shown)


hervenicol:
App gets deployed correctly, with no error.

Then, secrets named `[cluster]-remote-write-api-endpoint-config` are not updated, because immutable, but it looks like this does not generate errors. Actually, I think PMO does not even try to update these resources when already existing.
But if I delete one, it is then re-created (after a few minutes) with the new settings.

So, I think there's nothing preventing from releasing this change, then look for the best way to recycle these secrets.
At least, that would give a way to change and persist settings on installations that have problems with prometheus-agent.

TheoBrigitte (approved):
Sounds legit. I think we should document why this secret is immutable, but not necessarily in this PR.

hervenicol:
> Sounds legit. I think we should document why this secret is immutable, but not necessarily in this PR.

I guess @QuentinBisson or @mcbenjemaa have more context on this?
But I agree: not this PR's problem :grin:

QuentinBisson:
FYI it is immutable because we decided that the api key for the remote write endpoint secret should not change on every reconciliation so making it immutable made sense (because we should technically not change it) but we can definitly make it an Opaque secret so we can change the queueConfig

PR #1112, opened 2023-01-06
fix node-exporter discovery

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/25216

This PR fixes the `node-exporter` job to correctly ignore targets when they have the `giantswarm.io/monitoring` annotation or label.

This was broken due to the fact that `node-exporter` job is targeting `pods` but our relabeling config was using service metadata.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add proxy support to remote write endpoint consumers.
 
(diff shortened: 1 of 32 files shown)


TheoBrigitte:
This fixed the duplicated node-exporter targets

### After

node-exporter targets pods are only present in the `workload-guppy` job

image
image

hervenicol (approved):

QuentinBisson (approved):

PR #1115, opened 2023-01-12
Fix getDefaultAppVersion org namespace

mcbenjemaa:
Sometimes the org label is wrong, 

which leads to:
```
E 01/12 09:02:32 org-multi-project/etcd1 failed to reconcile | micrologger@v1.0.0/logger.go:88 | controller=prometheus-meta-operator-cluster-api-controller | event=update | loop=568 | version=58107919
	/go/pkg/mod/github.com/giantswarm/operatorkit/v7@v7.1.0/pkg/controller/controller.go:532
	/go/pkg/mod/github.com/giantswarm/operatorkit/v7@v7.1.0/pkg/controller/controller.go:567
	/go/pkg/mod/github.com/giantswarm/operatorkit/v7@v7.1.0/pkg/resource/wrapper/metricsresource/basic_resource.go:43
	/go/pkg/mod/github.com/giantswarm/operatorkit/v7@v7.1.0/pkg/resource/wrapper/retryresource/basic_resource.go:64
	/go/pkg/mod/github.com/giantswarm/backoff@v1.0.0/retry.go:23
	/go/pkg/mod/github.com/giantswarm/operatorkit/v7@v7.1.0/pkg/resource/wrapper/retryresource/basic_resource.go:52
	/root/project/service/controller/resource/generic/create.go:14
	/root/project/service/controller/resource/monitoring/scrapeconfigs/resource.go:133
	/root/project/service/controller/resource/monitoring/scrapeconfigs/resource.go:160
	/root/project/service/controller/resource/monitoring/scrapeconfigs/resource.go:174
	/root/project/service/controller/resource/monitoring/scrapeconfigs/resource.go:221
	unknown: apps.application.giantswarm.io "etcd1-default-apps" not found
```

@QuentinBisson  what do you think?


service/controller/resource/monitoring/scrapeconfigs/resource.go
```diff
@@ -201,7 +201,7 @@ func getTemplateData(ctx context.Context, ctrlClient client.Client, cluster meta
 
 func getDefaultAppVersion(ctx context.Context, ctrlClient client.Client, cluster metav1.Object, config Config) (string, error) {
 	appName := fmt.Sprintf("%s-default-apps", key.ClusterID(cluster))
-	appNamespace := fmt.Sprintf("org-%s", key.GetOrganization(cluster))
+	appNamespace := cluster.GetNamespace()
```
QuentinBisson, line 204: If this is true, sure, but I would rather kaas teams validate the org label

QuentinBisson (approved):

service/controller/resource/monitoring/scrapeconfigs/resource.go
mcbenjemaa: @giantswarm/area-kaas  Do we have an admission that validates the org label?

fiunchinho: If sometimes the label is wrong, then I don't think we do :sweat_smile: 

When it's wrong, what's its value? what does wrong mean here?

mcbenjemaa: Sometimes users created clusters with a wrong org label.
For example, they changed the namespace of the app.
But the cluster CR still points to org-giantswarm..

fiunchinho: If I remember correctly, we decided to remove the organization label during a KaaS sync last year. The reasoning being that the namespace is the source of truth for the organization. So I think it's good that you move away from using it.

QuentinBisson: I really wonder if those decisions should not be shared in news channel or in adrs because this is a pain to track all those changes that are impacting for consumers 😅 Thanks @fiunchinho

service/controller/resource/monitoring/scrapeconfigs/resource_test.go
```diff
@@ -163,44 +163,39 @@ func TestOpenStackScrapeconfigs(t *testing.T) {
 	var apps = []runtime.Object{
 		&appsv1alpha1.App{
 			ObjectMeta: v1.ObjectMeta{
-				Name:      "alice-default-apps",
-				Namespace: "org-my-organization",
```
QuentinBisson, line 167: How is removing the namespace fixing it?

mcbenjemaa: the fake client, i guess fetching apps from default

QuentinBisson: Except this is not fixing the test but making it less precise. What you have to do is change the namespace on the input values https://github.com/giantswarm/prometheus-meta-operator/tree/master/pkg/unittest/input instead

QuentinBisson: Beware this might break other stuff but this is what needs to be done :)

QuentinBisson: Some might need to stay in the default ns btw

mcbenjemaa: I fixed it

pkg/unittest/input/case-3-kvmconfig.golden
```diff
@@ -6,3 +6,4 @@ metadata:
     giantswarm.io/organization: my-organization
     "release.giantswarm.io/version": 18.0.0
   name: bar
+  namespace: org-my-organization
```
QuentinBisson, line 9:
```suggestion
  namespace: org-my-organization
  
```

pkg/unittest/input/case-2-azureconfig.golden
```diff
@@ -6,3 +6,4 @@ metadata:
     giantswarm.io/organization: my-organization
     "release.giantswarm.io/version": 18.0.0
   name: foo
+  namespace: org-my-organization
```
QuentinBisson, line 9:
```suggestion
  namespace: org-my-organization

```

QuentinBisson (approved):
Fine with me if you add the eol back :)

mcbenjemaa:
I will release a patch version, since there's a couple of issues in different clusters...

PR #1120, opened 2023-01-16
Bump prometheus and alertmanager versions

QuentinBisson:
Bump prometheus and alertmanager to match prometheus-operator default versions

## Checklist

I have:


main.go
```diff
@@ -135,7 +135,7 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Retention.Duration, "2w", "Retention duration for prometheus.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Retention.Size, "90Gi", "Retention size for prometheus.")
 	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Storage.Size, "100Gi", "Storage size for prometheus.")
-	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Version, "v2.39.1", "Prometheus container image version.")
+	daemonCommand.PersistentFlags().String(f.Service.Prometheus.Version, "v2.41.0", "Prometheus container image version.")
```
hervenicol, line 138: Why 2.41.0 when changelog says 2.40.1 ?
BTW, 2.41.0 is only RC for the moment, latest stable is 2.40.7.

hervenicol (commented):
Could you explain why upgrading prometheus and alertmanager requires that much changes?
~~And why we don' upgrade to current latest prometheus version?~~ to match prometheus-operator versions :+1:

main.go
QuentinBisson: Becaue of this https://github.com/prometheus-community/helm-charts/blob/f4555244480525b46a6fad320960955d655ca376/charts/kube-prometheus-stack/values.yaml#L2666

QuentinBisson:
Sure thing Herve, I was planning on doing it today :)


So basically, this bump prometheus and alertmanager version was made to use the versions configured in prometheus-operator app but when upgrading the app on all MCs yesterday, it paged a lot because the PMO dependencies are old and prometheus-operator changed a lot of it's internal and CRDs default values (without bumping the major version of course) which cause a lot of issues (prometheus keep restarting).

Those changes bump the prometheus-operator dependencies to avoid that. Prometheus will still restart and some heartbeat will fail (because it's using a statefulset with only 1 replicas so rolling is not really rolling when only 1 replicas) but heartbeats are less failures.

In the end, those changes basically make prometheus-meta-operator compatible with the latest prometheus-operator

files/templates/alertmanager/alertmanager.yaml
```diff
@@ -7,7 +7,7 @@ global:
   slack_api_url: [[ .SlackApiURL ]]
 
 templates:
-- '*.tmpl'
+- '/etc/alertmanager/config/*.tmpl'
```
QuentinBisson, line 10: In latest prometheus-operator, the template location needs to be explicit otherwise they are not found

main.go
QuentinBisson: I updated the changelog

QuentinBisson: And actually it's not RC now :D https://github.com/prometheus/prometheus/releases/tag/v2.41.0

service/controller/resource/alerting/alertmanager/test/case-1-awsconfig.golden
```diff
@@ -16,6 +16,7 @@ spec:
             operator: NotIn
             values:
             - master
+  alertmanagerConfigMatcherStrategy: {}
```
QuentinBisson, line 19: Coming from latest prom operator

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -164,100 +155,110 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
...
-				},
-			},
-			Retention:      config.RetentionDuration,
-			RetentionSize:  config.RetentionSize,
-			WALCompression: &walCompression,
-			AdditionalScrapeConfigs: &corev1.SecretKeySelector{
-				LocalObjectReference: corev1.LocalObjectReference{
-					Name: key.PrometheusAdditionalScrapeConfigsSecretName(),
+
+			CommonPrometheusFields: promv1.CommonPrometheusFields{
```
QuentinBisson, line 159: In latest prometheus-operator, most of the common fields are shared in a CommonPrometheusFields.

This is prework so they can add PrometheusAgentCR

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -164,100 +155,110 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
...
 			},
-			PriorityClassName: "prometheus",
-		},
-	}
-
-	// Add 5mn initial delay to the readinessProbe to give Prometheus more
-	// time to load data and start. This fixes the current 10mn delay set
-	// by default in prometheus-operator which results in crash looping Prometheus
-	// when there are too much data to load.
-	prometheus.Spec.Containers = []corev1.Container{
```
QuentinBisson, line 255: Moved this to the common section

service/controller/resource/monitoring/prometheus/test/case-1-awsconfig.golden
```diff
@@ -37,8 +37,7 @@ spec:
     readinessProbe:
       initialDelaySeconds: 300
     resources: {}
-  enableFeatures:
-  - remote-write-receiver
+  enableRemoteWriteReceiver: true
```
QuentinBisson, line 40: Other syntax is deprecated for remote write

main.go
hervenicol: erf, changelog on their master branch is not up to date :disappointed:

marieroque (approved):

PR #1137, opened 2023-01-24
Move Alertmanager to helm

mcbenjemaa:
Towards https://github.com/giantswarm/giantswarm/issues/25137

This PR:

- Moves the Alertmanager CR to the Helm chart


QuentinBisson:
All the static ones. I think alertmanager and
alertmanagerconfig are the only ones but let's check :)

helm/prometheus-meta-operator/templates/alertmanager/servicemonitor.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.alertmanager.enabled }}
```
primeroz, line 1: should we also remove alertmanager scraping from https://github.com/giantswarm/prometheus-meta-operator/blob/master/files/templates/scrapeconfigs/additional-scrape-configs.template.yaml#L660 ? 

this target is not matching any pods in discovery anyway , at least in CAPZ , because the label selector is not like this one in the SM

mcbenjemaa: However, i think we have to remove this target

mcbenjemaa: since, we have added a ServiceMonitor to scrape Alertmanager

primeroz: that's what i meant. since we are adding the servicemonitor then should we remove the hardcoded one ... did not have any better place to put that comment :)

mcbenjemaa: Thanks a lot

helm/prometheus-meta-operator/values.yaml
```diff
@@ -14,6 +14,7 @@ pod:
     id: 1000
 
 alertmanager:
+  enabled: false
```
QuentinBisson, line 17: Why is that not true?

files/templates/alertmanager/alertmanager.yaml
```diff
@@ -6,487 +6,487 @@ global:
 [[- end ]]
   slack_api_url: [[ .SlackApiURL ]]
```
QuentinBisson, line 7: Should this file not be gone?

helm/prometheus-meta-operator/templates/alertmanager/alertmanager.yaml
```diff
@@ -1,8 +1,9 @@
+{{- if .Values.alertmanager.enabled }}
```
QuentinBisson, line 1: I thinks there's an issue here right?

helm/prometheus-meta-operator/values.yaml
mcbenjemaa: We need 2 releases: 
- a release to delete alertmanager ownership from PMO
- another release that inlcudes alertmanager in Helm

service/controller/resource/alerting/alertmanagerconfig/resource.go
```diff
@@ -93,10 +93,10 @@ func toSecret(v interface{}, config Config) (*corev1.Secret, error) {
 		return nil, microerror.Mask(err)
 	}
 
-	notificationTemplate, err := renderNotificationTemplate(templateDirectory, config)
```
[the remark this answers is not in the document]
mcbenjemaa: Yes, I guess, 
there's some properties That I could not set them in Helm, essentially e.g. the Proxy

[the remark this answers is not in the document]
mcbenjemaa: but almost the config was re-written to the AlertmanagerConfig CR.

mcbenjemaa:
another issue: 
```
ts=2023-03-02T10:50:32.888Z caller=coordinator.go:118 level=error component=configuration msg="Loading configuration file failed" file=/etc/alertmanager/config_out/alertmanager.env.yaml err="no global Slack API URL set either inline or in a file"
```

files/templates/alertmanager/alertmanager.yaml
mcbenjemaa: Well, I had conflicts in the usage of configSecret and the `alertmanagerConfiguration`. 
they couldn't be loaded without minimum 1 route.

My idea is to keep some global config in the `configSecret` for now, 
Later, once `alertmanagerConfiguration` has more features so we can fully remove the configSecret.

so, configSecret defines global config.

and the `AlertmanagerConfig` gets merged into the config and overwrite the `root` route.

mcbenjemaa:
configSecret will:

```
## TODO: To be deleted and moved
# to Alertmanager.spec.alertmanagerConfiguration
# once the field has more support for our uses cases: e.g. (slack_api_url)
global:
  resolve_timeout: 5m
[[- if .ProxyURL ]]
  http_config:
    proxy_url:  [[ .ProxyURL ]]
[[- end ]]
  slack_api_url: [[ .SlackApiURL ]]

templates:
  - '/etc/alertmanager/config/*.tmpl'

route:
  group_by: [alertname, cluster_id, installation, status]
  group_interval: 15m
  group_wait: 30s
  repeat_interval: 4h
  receiver: root

receivers:
  - name: root

```

mcbenjemaa:
@QuantumEnigmaa we have to talk about inhibitions script test.

QuantumEnigmaa:
> we have to talk about inhibitions script test.

Sure !

mcbenjemaa (commented):
This PR is ready:
please read this [comment](https://github.com/giantswarm/giantswarm/issues/25137#issuecomment-1465846179)

helm/prometheus-meta-operator/templates/_helpers.tpl
```diff
@@ -32,3 +32,18 @@ Selector labels
...
+
+{{- define "prometheus-meta-operator.alertmanager-ownership-annotations" -}}
+"helm.sh/hook": "pre-install,pre-upgrade"
+"helm.sh/hook-delete-policy": "before-hook-creation,hook-succeeded"
+{{- end -}}
+
+{{/* Create a label which can be used to select any orphaned alertmanager-ownership hook resources */}}
+{{- define "prometheus-meta-operator.alertmanager-ownership-selector" -}}
+{{- printf "%s" "alertmanager-ownership-hook" -}}
+{{- end -}}
```
QuentinBisson, line 49: Line break

QuentinBisson (approved):
LGTM

mcbenjemaa:
This has been tested few times, 👍 
everything works ✅

TheoBrigitte (approved):

PR #1148, opened 2023-01-27
remove prometheus readinessProbe delay

TheoBrigitte:
This changed was introduced back in #893 and is no longer needed, as there is now a 15mn startupProbe allowing for slow Prometheus with a big WAL data to load.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Remove Prometheus readinessProbe 5mn delay; since there is already a 15mn startupProbe
+
 ### Changed
 
 - Use resource.Quantity.AsApproximateFloat64() instead of AsInt64(), in order to avoid conversion issue when multiply cpu, e.g. 3880m
--- a/service/controller/resource/monitoring/prometheus/resource.go
+++ b/service/controller/resource/monitoring/prometheus/resource.go
@@ -182,18 +182,6 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 						},
 					},
 				},
-				// Add 5mn initial delay to the readinessProbe to give Prometheus more
-				// time to load data and start. This fixes the current 10mn delay set
-				// by default in prometheus-operator which results in crash looping Prometheus
-				// when there are too much data to load.
-				Containers: []corev1.Container{
-					{
-						Name: "prometheus",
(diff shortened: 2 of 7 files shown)


hervenicol:
I investigated with Theo, and I pre-approve this PR :grin:

TheoBrigitte:
But this need to be tested first.

This need to be tested in the scenario where Prometheus takes more than 5mn; suggestion is to create a large test cluster (~100 nodes) to try reproduce this.

Also the goal here is to improve the prometheus-agent reliability, because of this delay the agent always has to wait another 5mn before sending its data upon each Prometheus restart. Without this we hope this gets slightly better.

hervenicol:
Created a 100-node WC on `gaia`.
Let prometheus run for 3h so its WAL is full.

## Killing prometheus with current config

* first message in logs: `ts=2023-02-23T13:37:59.888Z`
* `ready` message: `ts=2023-02-23T13:39:13.294Z caller=main.go:978 level=info msg="Server is ready to receive web requests."`
That makes less than 2 minutes of replaying WAL, not much.

Concerning init time (polled every 5 seconds):
* init
```
Thu 23 Feb 14:37:54 CET 2023                                                                                                                                                                  
NAME                 READY   STATUS     RESTARTS   AGE                                                                                                                                        
prometheus-t3q7m-0   0/2     Init:0/1   0          2s
```
* Running at 12s
```
Thu 23 Feb 14:38:04 CET 2023                                                                                                                                                                  
NAME                 READY   STATUS    RESTARTS   AGE                                                                                                                                         
prometheus-t3q7m-0   1/2     Running   0          12s 
```

* 2/2 Ready at 5m8s, so it dit not wait for the end of the WAL replay before counting.
```
Thu 23 Feb 14:43:00 CET 2023
NAME                 READY   STATUS    RESTARTS   AGE
prometheus-t3q7m-0   2/2     Running   0          5m8s
```

QuentinBisson (approved):
Thank, that's great :)

hervenicol:
Effect on the agent, looking at remote-writes graphs:

We can see that the agent "stacked" less pending metrics for the second restart, and did less retries.
image
image

hervenicol:
So here, it takes 2min before prometheus is ready vs 5min before.
That's a 3 minutes improvement in start time. The gap is not huge, but it's a big cluster with a lot of nodes, and a big WAL.
When the WAL takes 30s to load, the improvement will be 4m30.

TheoBrigitte (commented):
:heavy_check_mark: approved (I cannot approve as I am the author here)

PR #1149, opened 2023-01-31
switch to Aliyun Container Registry to Enterprise

kubasobon:


hervenicol:
This means new images are pushed to new registry, but not available any more on old registry.
It means next release of this app has to be synchronized with a change of CN registry for it?

It seems quite touchy to me. Could we keep pushing to both registries for a while, to ease switch?

kubasobon:
The old registry is still active and contains every tag pushed there. New tags/repositories should be pushed to new registry instead.
New registry has all the old repositories/tags imported though, so we're safe.

hervenicol:
So, once this PR is merged, we can't upgrade until registry is switched.
I guess all is ready for the registry switch, and this is a very temporary situation?

hervenicol (approved):
Good luck with that migration :crossed_fingers: :four_leaf_clover:

kubasobon:
Next release will just pull images from new registry. You're free to upgrade. Pulling does not require credentials.

PR #1155, opened 2023-02-07
Fix certificates created as directories rather than files

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/25713

=> to avoid certificates created as directories when `pmo` is started before systemd unit runs into the node.

## Checklist

I have:


QuentinBisson (approved):
<3

PR #1158, opened 2023-02-08
Fix etcd-certs

mcbenjemaa:
Towards https://github.com/giantswarm/giantswarm/issues/25713

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix PMO etcd-certs volumes binding
+- Make PMO etcd-certs volumes binding only on vintage, so that CAPI doesn't fail deployment because the certs are binded via a secret.  
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Make PMO etcd-certs volumes binding only on vintage, so that CAPI doesn't fail deployment because the certs are binded via a secret.  
+- Manage etcd certificates differently between CAPI/Vintage. On Vintage, etcd certificates are binding via a volume. On CAPI, certificates are binding via a secret.
 
 ### Added
 
-- Add `.provider.flavor` property 
+- Add `.provider.flavor` property in Helm chart. 
 
 ## [4.20.4] - 2023-02-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Manage etcd certificates differently between CAPI/Vintage. On Vintage, etcd certificates are binding via a volume. On CAPI, certificates are binding via a secret.
+- Manage etcd certificates differently between CAPI/Vintage. On Vintage, etcd certificates are binded via a volume. On CAPI, certificates are binded via a secret.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix PMO etcd-certs volumes binding
+
+### Added
+
+- Add `.provider.flavor` property 
+
 ## [4.20.4] - 2023-02-07
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


helm/prometheus-meta-operator/templates/deployment.yaml
```diff
@@ -33,6 +33,7 @@ spec:
           items:
           - key: secret.yml
             path: secret.yml
+      {{- if eq .Values.managementCluster.provider.flavor "vintage" }}
```
QuentinBisson, line 36: How does that fix the issue ? Because if the secret does not exist, sure PMO will run but there will be no CA for etcd :)

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix PMO etcd-certs volumes binding
```
QuentinBisson, line 12: Can you explain the issue a bit more?

helm/prometheus-meta-operator/templates/deployment.yaml
mcbenjemaa: The secret is available in CAPI, but doesn't exists in Vintage

QuentinBisson: Oh ok then maybe we need this if in the go code also where the etcd secret is read from the disk

marieroque: `pmo` will raise errors if secret not found:
https://github.com/giantswarm/prometheus-meta-operator/blob/master/service/controller/resource/etcd-certificates/source.go#L38

QuentinBisson (approved):

helm/prometheus-meta-operator/templates/deployment.yaml
mcbenjemaa: if not, pmo will get the certs from Disk

QuentinBisson: Yes but then why would PMO go through  sc.getSourceFromDisk()  on CAPI :D

marieroque: On CAPI there are no etcd certificates on disk according Pau

mcbenjemaa: What do you suggest?

mcbenjemaa: I think the code, is good. 
We just added a condition in helm, so we make sure that the etcd-certs binding are only for vintage.

QuentinBisson: So my suggestion is that we add this check also in the go code because if  the etcd secret source namespace is not found on capi clusters, what will happen to the code? Will sc.getSourceFromDisk() error out or panic and crash PMO?

marieroque: If the secret is not found,  pmo will look at certificates on disk and if there are none, it will throw an error like then one we have today when certificates are seen as directories rather than files.

I'm adding a test on CAPI vs Vintage to throw the proper error.

QuentinBisson: Sure I agree with you here ;)

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Manage etcd certificates differently between CAPI/Vintage. On Vintage, etcd certificates are binding via a volume. On CAPI, certificates are binding via a secret.
```
QuentinBisson, line 12:
```suggestion
- Manage etcd certificates differently between CAPI/Vintage. On Vintage, etcd certificates are binded via a volume. On CAPI, certificates are binded via a secret.
```

helm/prometheus-meta-operator/templates/deployment.yaml
QuentinBisson: BTW I approved yesterday :)

marieroque: I added some code this morning with the check you suggested... Can you review it ?

service/controller/resource/etcd-certificates/source.go
```diff
@@ -11,17 +11,19 @@ import (
...
 // getSource retrieves data for the desired Secret.
-func (sc *secretCopier) getSource(ctx context.Context, v interface{}) (map[string]string, error) {
-	data, err := sc.getSourceFromSecret(ctx, key.EtcdSecretSourceName(), key.EtcdSecretSourceNamespace())
-	if err != nil {
-		sc.logger.Debugf(ctx, "could not get certificates from secret : %v", err)
-
+func (sc *secretCopier) getSource(ctx context.Context, v interface{}, config Config) (map[string]string, error) {
+	var data map[string]string
+	var err error
+	if key.IsCAPIManagementCluster(config.Provider) {
```
QuentinBisson, line 17: <3

QuentinBisson (approved):

QuentinBisson:
Lovely :)

PR #1165, opened 2023-02-15
Improve VPA maxAllowedCpu

mcbenjemaa:
Towards https://github.com/giantswarm/giantswarm/issues/25877

This is a Proposal

Explanation: https://github.com/giantswarm/giantswarm/issues/25877#issuecomment-1437140886

Testing: Tested on testing clusters: gauss, gohan, golem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use resource.Quantity.AsApproximateFloat64() instead of AsInt64(), in order to avoid conversion issue when multiply cpu, e.g. 3880m
 - Use label selector to selects only worker nodes for vpa resource to get maxCPU and maxMemory
 - List nodes from API only once in VPA resource
-- Improve VPA maxAllowedCPU depending on the node allocatable CPU
+- Improve VPA maxAllowedCPU, use 70% of the node allocatable CPU.
 
 ## [4.20.6] - 2023-02-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use resource.Quantity.AsApproximateFloat64() instead of AsInt64(), in order to avoid conversion issue when multiply cpu, e.g. 3880m
 - Use label selector to selects only worker nodes for vpa resource to get maxCPU and maxMemory
 - List nodes from API only once in VPA resource
+- Improve VPA maxAllowedCPU depending on the node allocatable CPU
 
 ## [4.20.6] - 2023-02-13
 
--- a/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
@@ -175,13 +175,10 @@ func (r *Resource) getMaxCPU(nodes *v1.NodeList) (*resource.Quantity, error) {
 		return nil, microerror.Mask(nodeCpuNotFoundError)
 	}
 
-	n, ok := nodeCpu.AsInt64()
-	if !ok {
-		return nil, microerror.Maskf(quantityConversionError, nodeCpu.String())
-	}
-
+	// We cast nodeCpu.AsApproximateFloat64()
+	// to avoid failure in clusters that have millis cpu set.
+	n := int64(nodeCpu.AsApproximateFloat64())
 	var q *resource.Quantity
-
 	if n <= 8 {
 		// If allocatable cpu is equal or less than 8
 		// maxAllowedCPU is 70% of total
--- a/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
@@ -175,20 +175,7 @@ func (r *Resource) getMaxCPU(nodes *v1.NodeList) (*resource.Quantity, error) {
 		return nil, microerror.Mask(nodeCpuNotFoundError)
 	}
 
-	// We cast nodeCpu.AsApproximateFloat64()
-	// to avoid failure in clusters that have millis cpu set.
-	n := int64(nodeCpu.AsApproximateFloat64())
-	var q *resource.Quantity
-	if n <= 8 {
-		// If allocatable cpu is equal or less than 8
-		// maxAllowedCPU is 70% of total
-		q, err = quantityMultiply(nodeCpu, 0.7)
-	} else {
-		// More than 8 CPUs
-		// maxAllowedCPU is 50% of total
-		q, err = quantityMultiply(nodeCpu, 0.5)
-	}
-
+	q, err := quantityMultiply(nodeCpu, 0.7)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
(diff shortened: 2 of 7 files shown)


mcbenjemaa:
Tested on gohan,

I need to tested on grizzly as well,

CHANGELOG.md
```diff
@@ -7,13 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+<<<<<<< Updated upstream
```
hervenicol, line 10: oops, merge conflicts?

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -165,7 +165,25 @@ func (r *Resource) getMaxCPU(ctx context.Context) (*resource.Quantity, error) {
...
+	if n <= 6 {
+		q, err = quantityMultiply(nodeCpu, 0.7)
+		// If allocatable cpu is more then 12
+		// maxAllowedCPU is 60%
+	} else if n >= 12 {
+		q, err = quantityMultiply(nodeCpu, 0.6)
+		// Default 90%
+	} else {
+		q, err = quantityMultiply(nodeCpu, 0.9)
+	}
```
hervenicol, line 185: 1. let's set the comments inside each "if" so it's easier to spot what it talks about.
2. set the conditions in an ascending order, so it's easier to understand the logic

```suggestion
	if n <= 6 {
		// If allocatable cpu is less than 6
		// maxAllowedCPU is 70% of total
		q, err = quantityMultiply(nodeCpu, 0.7)
	} else if n < 12 {
		// Between 6 and 12 CPUs
		// maxAllowedCPU is 90% of total
		q, err = quantityMultiply(nodeCpu, 0.9)
	} else {
		// More than 12 CPUs
		// maxAllowedCPU is 60% of total
		q, err = quantityMultiply(nodeCpu, 0.6)
	}
```

I don't understand why the bump in %age between 6 and 12 CPUs.
Maybe this code will require a link to the page with your findings regarding CPU usage.

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -165,7 +165,25 @@ func (r *Resource) getMaxCPU(ctx context.Context) (*resource.Quantity, error) {
...
-	q, err := quantityMultiply(nodeCpu, 0.9)
+	n, ok := nodeCpu.AsInt64()
+	if !ok {
+		return nil, microerror.Maskf(quantityConversionError, nodeCpu.String())
+	}
+
+	var q *resource.Quantity
+	// If allocatable cpu is less then 6
+	// maxAllowedCPU is 70%
+	if n <= 6 {
```
QuentinBisson, line 176: As @hervenicol said, I'd like an explanation on those magical numbers. Ideally a link to some documentation maybe inside PMO

QuentinBisson:
@mcbenjemaa can you add more context in the PR description so we can understand what you want to fix and how without having to look at the code?

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -165,7 +165,23 @@ func (r *Resource) getMaxCPU(ctx context.Context) (*resource.Quantity, error) {
 		return nil, microerror.Mask(nodeCpuNotFoundError)
 	}
 
-	q, err := quantityMultiply(nodeCpu, 0.9)
+	n, ok := nodeCpu.AsInt64()
```
mcbenjemaa, line 168: This will fail in capz ⚠️

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -165,7 +165,23 @@ func (r *Resource) getMaxCPU(ctx context.Context) (*resource.Quantity, error) {
...
+	if !ok {
+		return nil, microerror.Maskf(quantityConversionError, nodeCpu.String())
+	}
+
+	var q *resource.Quantity
+
+	if n <= 8 {
```
QuentinBisson, line 175: Still, what are those magic numbers? :D

mcbenjemaa: https://github.com/giantswarm/giantswarm/issues/25877#issuecomment-1437140886

mcbenjemaa: I will create a table later.

QuentinBisson: Sure because I think this should be documented here TBH.

Also, I might be wrong on my hypothesis but I think that not allowing more than 50% of node CPU should work everywhere. I'd like to know why deu01 requires so much but I guess this is due to rule evaluation again?

QuentinBisson: There might be a correlation about rule evaluation * nodes = CPU usage but we should be fine with 50% top

QuentinBisson: I.e. I do not think the number of allocatable CPUs on the node should be a good way to know if we need to limit to 50% or more of CPU usage but it should be around the number of worker nodes.

We can revisit this later tbh. If 50 works, let's go with 50

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
mcbenjemaa: Fixed

mcbenjemaa:
Added a table: https://github.com/giantswarm/giantswarm/issues/25877#issuecomment-1437219497

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
mcbenjemaa: So, you mean 50% for all?

mcbenjemaa: But We have grizzly failing! It had to requests 2 CPU at least

mcbenjemaa: grizzly nodes have 4 cpu. so if we have 70% (2.8)

mcbenjemaa: I see 0.5 is enough, but for the nodes that has only 4 cpus and needs to evaluate something. 
I really want to avoid the eviction of Prometheus Pod circumstances, due to high cpu usage.

QuentinBisson: Then maybe we set it to 70 everywhere 😅

mcbenjemaa: Yes, If you wish)

mcbenjemaa: done)

hervenicol: Can nodes get evicted due to high CPU usage?
It's not like RAM where you can OOM. Your app just gets slow, has problems catching back, and maybe can't do its job anymore. But it won't crash (except unknown reasons like strange race conditions that only happen when under a lot of stress).

hervenicol:
So, we just reduce CPU% from 90% to 70%.

That can seem quite a lot, but as this change goes from the switch from integers to floats, for some 4-cores servers this means we go from 3 (3.6 computed, but truncated) to 2.8.

Won't change much. If we want to have a real impact on CPU usage on 4-cores nodes, I advocate for an even lower threshold.
If we go to 50%: worst case is promtheus will behave strangely on small nodes, data may not be very reliable. But at least it won't prevent other pods from starting.

So I advocate for 50%, drastic cut! :fire:

hervenicol (approved):
Even though I think the 70% threshold is a timid reduction and I don't think it will change much on low-end nodes, this is enough to ensure we can release PMO and at least not allow more CPU tu prometheus than before. :+1:

QuantumEnigmaa (approved):

PR #1166, opened 2023-02-16
Use resource.Quantity.AsApproximateFloat64() instead of AsInt64()

mcbenjemaa:
Towards https://github.com/giantswarm/giantswarm/issues/25909

Explantation: https://github.com/giantswarm/giantswarm/issues/25909#issuecomment-1433320441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Use resource.Quantity.AsApproximateFloat64() instead of AsInt64(), in order to avoid conversion issue when multiply cpu, e.g. 3880m
+
 ## [4.20.6] - 2023-02-13
 
 ### Changed
--- a/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
@@ -215,7 +215,7 @@ func quantityMultiply(q *resource.Quantity, multiplier float64) (*resource.Quant
 
 	i := q.AsApproximateFloat64()
 	n := i * multiplier
-	q.Set(int64(n))
+	q.SetMilli(int64(n * 1000))
 
 	return q, nil
 }
--- a/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
@@ -212,13 +212,9 @@ func (r *Resource) getMaxMemory(ctx context.Context) (*resource.Quantity, error)
 }
 
 func quantityMultiply(q *resource.Quantity, multiplier float64) (*resource.Quantity, error) {
-	i, ok := q.AsInt64()
-	if !ok {
-		return nil, microerror.Maskf(quantityConversionError, q.String())
-	}
 
-	n := float64(i) * multiplier
-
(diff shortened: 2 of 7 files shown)


primeroz:
this did work on `glippy` and addressed both issues we had - https://github.com/giantswarm/roadmap/issues/2021#top

* VPA for prometheus is now created
```
➜ k get vpa -n glippy-prometheus
NAME         MODE   CPU   MEM        PROVIDED   AGE
prometheus   Auto   11m   52428800   True       2m2s
```
* HB is created ( so for some reason the failing VPA was breaking the reconcile loop entirely ) 
Screenshot from 2023-02-16 17-07-03


kudos @mcbenjemaa for the super quick turn-around on this

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Use resource.Quantity.AsApproximateFloat64() instead of AsInt64(), in order to avoid conversion issue when multiply cpu, e.g. 3880m
+
```
hervenicol, line 13: I'm not sure I understand the problematic case.
* It fails at multiplying `3880*0.9`? That's a round number - 3492
* It fails at multiplying `3.880x0.9`? Then, why it didn't fail when multiplying `4x0.9`?

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -212,13 +212,9 @@ func (r *Resource) getMaxMemory(ctx context.Context) (*resource.Quantity, error)
...
-	i, ok := q.AsInt64()
-	if !ok {
-		return nil, microerror.Maskf(quantityConversionError, q.String())
-	}
-
-	n := float64(i) * multiplier
 
+	i := q.AsApproximateFloat64()
+	n := i * multiplier
 	q.Set(int64(n))
```
hervenicol, line 218: No need to `round()` to get an int from a float? (sorry, golang noob).

hervenicol (commented):
I totally agree with the idea that we should use floats when multiplying by 0.x numbers.
I just don't understand how the example value fails. Personal curiosity.
And go noob questions.

hervenicol (approved):
LGTM if tested.

mcbenjemaa:
**Tested:** 

- [x] glippy
- [x] gauss
- [x] gohan
- [x] golem

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -212,14 +212,10 @@ func (r *Resource) getMaxMemory(ctx context.Context) (*resource.Quantity, error)
...
-	if !ok {
-		return nil, microerror.Maskf(quantityConversionError, q.String())
-	}
-
-	n := float64(i) * multiplier
 
-	q.Set(int64(n))
+	i := q.AsApproximateFloat64()
+	n := i * multiplier
+	q.SetMilli(int64(n * 1000))
```
QuentinBisson, line 218: Why use `SetMilli` to then multiply it by 1000 ?

hervenicol: We have a number like `2.5`
`Set` requires cast to int => becomes `2`
and `SetMilli` uses milli-cpus, so expects `2500`.

QuentinBisson: :ack:

PR #1167, opened 2023-02-17
Optimize the request to the API for VPA resource

mcbenjemaa:
This will list nodes only once, and will only selects worker nodes

## Checklist

I have:


mcbenjemaa:
@JosephSalisbury  😅

hervenicol (approved):
Looks good.
BTW: if we have different nodepools with different node sizes, our computation may be completely whrong, right?
I guess that wouldn't be our only problem then, though... :sweat_smile:

QuantumEnigmaa (approved):

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -140,12 +143,20 @@ func (r *Resource) getObject(ctx context.Context, v interface{}) (*vpa_types.Ver
 	return vpa, nil
 }
 
-func (r *Resource) getMaxCPU(ctx context.Context) (*resource.Quantity, error) {
-	nodes, err := r.k8sClient.K8sClient().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+func (r *Resource) listWorkerNodes(ctx context.Context) (*v1.NodeList, error) {
+
```
QuentinBisson, line 147: What's your thing with extra lines :D

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -140,12 +143,19 @@ func (r *Resource) getObject(ctx context.Context, v interface{}) (*vpa_types.Ver
 	return vpa, nil
 }
 
-func (r *Resource) getMaxCPU(ctx context.Context) (*resource.Quantity, error) {
-	nodes, err := r.k8sClient.K8sClient().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+func (r *Resource) listWorkerNodes(ctx context.Context) (*v1.NodeList, error) {
```
QuentinBisson, line 146: I would have moved this to a shared place because I see this being reusable but I'm fine for now

QuentinBisson:
Can you remove the binary?

This VPA CR is created on MCs only so we have only 1 node pool

mcbenjemaa:
@QuentinBisson 

> Can you remove the binary?

Which binary?

QuentinBisson:
image

QuentinBisson (approved):

PR #1177, opened 2023-02-23
Increase Scrape and Evaluation intervals from 30s to 60s

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/25938

Increase ScrapeInterval and EvaluationInterval from 30s to 60s.

## Checklist

I have:


hervenicol:
Hmm... value from helm chart does not seem to be used. Something's missing in my code :thinking: 
Tested on gaia: PMO configmap looks good, but prometheus resources still have "30s" intervals.

hervenicol:
Tested on `gaia`: looks good!

CPU usage decreases. Twice less, because twice less rules evaluations.
image

* Rate of samples appended decreased a bit.
* Total series dropped, but is unrelated: recurrent prometheus cleanup.
* Looks like rule evaluation and notification latency decreased, probably because less contention on CPU.
image

No visible impact on RAM usage.
Disk usage to be seen on long term.

Drawbacks:
- less precision when looking at graphs: a really short peak will be smoothed over 1 minute instead of smoothed over 30 seconds.
- up to 30s more delay on alerts.

QuentinBisson (approved):
LGTM, let's revisit those settings when we move to mimir :)

QuantumEnigmaa (approved):

PR #1181, opened 2023-02-27
Find Prometheus Rules in all namespaces

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/25940 and https://github.com/giantswarm/giantswarm/issues/18281

## Checklist

I have:

### DIFF
--- a/service/controller/resource/monitoring/prometheus/resource.go
+++ b/service/controller/resource/monitoring/prometheus/resource.go
@@ -246,9 +246,14 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 			EvaluationInterval: promv1.Duration(config.EvaluationInterval),
 			Retention:          promv1.Duration(config.RetentionDuration),
 			RetentionSize:      promv1.ByteSize(config.RetentionSize),
+			// Fetches Prometheus rules from any namespace on the Management Cluster
+			// using https://v1-22.docs.kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name
 			RuleNamespaceSelector: &metav1.LabelSelector{
-				MatchLabels: map[string]string{
-					"name": key.NamespaceMonitoring(),
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{
+						Key:      "kubernetes.io/metadata.name",
+						Operator: metav1.LabelSelectorOpExists,
+					},
 				},
 			},
 		},
--- a/service/controller/resource/monitoring/prometheus/resource.go
+++ b/service/controller/resource/monitoring/prometheus/resource.go
@@ -281,6 +281,10 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 					Operator: metav1.LabelSelectorOpNotIn,
 					Values:   []string{"management_cluster"},
 				},
+				{
+					Key:      key.TeamLabel,
+					Operator: metav1.LabelSelectorOpExists,
+				},
 			},
 		}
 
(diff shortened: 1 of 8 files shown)


service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -246,9 +246,14 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 			EvaluationInterval: promv1.Duration(config.EvaluationInterval),
 			Retention:          promv1.Duration(config.RetentionDuration),
 			RetentionSize:      promv1.ByteSize(config.RetentionSize),
+			// Fetches Prometheus rules from any namespace on the Management Cluster
+			// using https://v1-22.docs.kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name
 			RuleNamespaceSelector: &metav1.LabelSelector{
-				MatchLabels: map[string]string{
-					"name": key.NamespaceMonitoring(),
+				MatchExpressions: []metav1.LabelSelectorRequirement{
```
hervenicol, line 252: I don't understand the point in just checking if this label exists.
As far as I understand, this label should exist on all namespaces, right?

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -246,9 +246,14 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 			EvaluationInterval: promv1.Duration(config.EvaluationInterval),
 			Retention:          promv1.Duration(config.RetentionDuration),
 			RetentionSize:      promv1.ByteSize(config.RetentionSize),
+			// Fetches Prometheus rules from any namespace on the Management Cluster
+			// using https://v1-22.docs.kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name
 			RuleNamespaceSelector: &metav1.LabelSelector{
-				MatchLabels: map[string]string{
-					"name": key.NamespaceMonitoring(),
```
hervenicol, line 251: We're removing this constraint, but do we really want to fetch ALL rules?
I mean, it's on MCs so chances are probably very low someone deploys some unknown rules, but I thought aligning on servicemonitors, using the "giantswarm.io/team" label, could have been nice?
Not defined yet on our rules though.

QuentinBisson: The issue here is that this is a `rule namespace selector` and we need to set it to something otherwise the prometheus will only have the rules present in it's namespace.

If we wanted to add the `giantswarm.io/team` label on the rules, we would need to change it in the `rule selector` instead.

Considering customers can now deploy things on MCs, maybe we want to reduce the scope to our apps.

That means I will create the PR on prometheusrules first then release PMO

QuentinBisson: The new commit addresses that :) I'll update the prometheus-rules app when I have time :D

QuentinBisson: Testing will happen tomorrow

service/controller/resource/monitoring/prometheus/resource.go
QuentinBisson: Yes the idea is that we check for rules in all namespaces as this label is set by Kubernetes, because otherwise prometheus operator will only check for rules in the namespace where prometheus is located and not the whole cluster

service/controller/resource/monitoring/prometheus/resource.go
hervenicol: Yes, looking good :smile:

QuantumEnigmaa (approved):
LGTM :)

QuentinBisson:
Tested on Gremlin

PR #1185, opened 2023-03-06
Change GetOrganization to rely on the namespace

mcbenjemaa:
Towards https://github.com/giantswarm/giantswarm/issues/26105

Both vintage and capi clusters reside on the org namespace. 
Since, org label is deprecated, we need to rely on the namespace...

### DIFF
--- a/service/key/key.go
+++ b/service/key/key.go
@@ -3,7 +3,6 @@ package key
 import (
 	"fmt"
 	"math"
-	"regexp"
 
 	"github.com/giantswarm/microerror"
 	v1 "k8s.io/api/core/v1"
@@ -20,11 +19,9 @@ const (
 	MonitoringNamespace = "monitoring"
 
 	DefaultServicePriority string = "highest"
-	DefaultOrganization    string = "giantswarm"
 
 	ClusterLabel                 string = "giantswarm.io/cluster"
 	MonitoringLabel              string = "giantswarm.io/monitoring"
-	OrganizationLabel            string = "giantswarm.io/organization"
 	ServicePriorityLabel         string = "giantswarm.io/service-priority"
 	TeamLabel                    string = "application.giantswarm.io/team"
 	OpsGenieApiKey               string = "opsGenieApiKey" // #nosec G101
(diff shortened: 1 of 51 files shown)


service/key/key.go
```diff
@@ -88,8 +88,10 @@ func GetServicePriority(cluster metav1.Object) string {
 }
 
 func GetOrganization(cluster metav1.Object) string {
-	if organization, ok := cluster.GetLabels()[OrganizationLabel]; ok && organization != "" {
-		return organization
+	re := regexp.MustCompile(`org-(.*)`)
+	match := re.FindStringSubmatch(cluster.GetNamespace())
+	if len(match) >= 2 {
+		return match[1]
```
fiunchinho, line 94: Could we fetch the organization from the namespace instead? In that case, if we ever change the name format used for our organizations (currently `org-$orgName`) we wouldn't need to change it here

There should be a label in the namespace with the organization name called `"giantswarm.io/organization"`.

QuentinBisson: We there should bé a label, are you sure it's set everywhere ? Because if it is, we can use thé label on thé namespace for sure.

QuentinBisson:
I think we need to keep thé default=giantwarm at least for vintage MCs

service/key/key.go
mcbenjemaa: fetching it from namespace, require a request to the api, 
You think that's okay, from a performance perspective...

QuentinBisson: I'm fine with 1 call to the api for each cluster every 5 minutes yes :)

QuentinBisson: But are we sure the label is set?

mcbenjemaa: it should be set, by the organization-operator

QuentinBisson: Can you confirm this is the case? :)

mcbenjemaa: https://github.com/giantswarm/organization-operator/blob/main/service/controller/resource/organization/resource.go#L78

T-Kukawka:
@giantswarm/team-atlas any progress here?

QuentinBisson:
Tested on Gauss. testing on a CAPI installation

QuentinBisson:
Tested on cuckoo

marieroque (approved):
LGTM

PR #1187, opened 2023-03-06
adjust VPA settings to prevent prometheus from using all node RAM

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/25767

Memory limit could go up to 108% of node's memory. And memory exhaustion on node leads to killing node.
This change in codes makes:
* max requests reduced from 90% to 80% node memory
* max limit reduced from 108% to 96% node memory

Maybe we should restrict even more to avoid nodes being killed, but I don't want to go too fast. Step by step.

Also, removed VPA config for `prometheus-config-reloader` and `rules-configmap-reloader`: these containers don't exist, so this config has no effect.
On top of that, we actually have a `config-reloader` container, which is scaled down with default VPA config, and it works well like that.

BTW, found and fixed a bug when our computed maxmemory would give a non-round number of bytes, and VPA did not like!
=> CPU can (should) be float
=> RAM must be int

## Checklist

I have:

### DIFF
--- a/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
@@ -204,7 +204,11 @@ func (r *Resource) getMaxMemory(nodes *v1.NodeList) (*resource.Quantity, error) 
 		return nil, microerror.Mask(nodeMemoryNotFoundError)
 	}
 
-	q, err := quantityMultiply(nodeMemory, 0.9)
+	// set max `requests` RAM to 80% node RAM.
+	// When setting default limit, make sure max VPA limit won't go higher than available RAM!
+	// because limit grows proportionnaly to requests, and here we compute max requests
+	// So check that PrometheusMemoryLimitCoefficient*MaxMemory < node memory
+	q, err := quantityMultiply(nodeMemory, 0.8)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
--- a/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
@@ -99,7 +99,6 @@ func (r *Resource) getObject(ctx context.Context, v interface{}) (*vpa_types.Ver
 
 	updateModeAuto := vpa_types.UpdateModeAuto
 	containerScalingModeAuto := vpa_types.ContainerScalingModeAuto
-	containerScalingModeOff := vpa_types.ContainerScalingModeOff
 	containerControlledValuesRequestsAndLimits := vpa_types.ContainerControlledValuesRequestsAndLimits
 	vpa := &vpa_types.VerticalPodAutoscaler{
 		ObjectMeta: objectMeta,
--- a/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
+++ b/service/controller/resource/monitoring/verticalpodautoscaler/resource.go
@@ -209,6 +209,12 @@ func (r *Resource) getMaxMemory(nodes *v1.NodeList) (*resource.Quantity, error) 
 		return nil, microerror.Mask(err)
 	}
 
+	// Memory must be a whole number of bytes
+	q.Set(int64(q.AsApproximateFloat64()))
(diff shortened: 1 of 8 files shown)


QuentinBisson:
I'll check tomorrow because i'm confused. I thought Mohamed reduced it already

hervenicol:
> I'll check tomorrow because i'm confused. I thought Mohamed reduced it already

AFAIK he worked on CPU, not RAM.

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -127,14 +126,6 @@ func (r *Resource) getObject(ctx context.Context, v interface{}) (*vpa_types.Ver
 							v1.ResourceMemory: *maxMemory,
 						},
 					},
-					{
-						ContainerName: "prometheus-config-reloader",
```
QuentinBisson, line 131: Good to know

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -214,6 +209,12 @@ func (r *Resource) getMaxMemory(nodes *v1.NodeList) (*resource.Quantity, error)
 		return nil, microerror.Mask(err)
 	}
 
+	// Memory must be a whole number of bytes
+	q.Set(int64(q.AsApproximateFloat64()))
```
QuentinBisson, line 213: Can we not use https://github.com/kubernetes/apimachinery/blob/f357b1fa74b71ca0cffc090e447983950c2bafc5/pkg/api/resource/quantity.go#L483 ?

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
```diff
@@ -204,7 +195,11 @@ func (r *Resource) getMaxMemory(nodes *v1.NodeList) (*resource.Quantity, error)
 		return nil, microerror.Mask(nodeMemoryNotFoundError)
 	}
 
-	q, err := quantityMultiply(nodeMemory, 0.9)
+	// set max `requests` RAM to 80% node RAM.
+	// When setting default limit, make sure max VPA limit won't go higher than available RAM!
+	// because limit grows proportionnaly to requests, and here we compute max requests
+	// So check that PrometheusMemoryLimitCoefficient*MaxMemory < node memory
+	q, err := quantityMultiply(nodeMemory, 0.8)
```
QuentinBisson, line 202: Are we sure 0.8 is correct everywhere?

QuentinBisson: I read the end :)

hervenicol: Makes the max `requests` to 80% of total node capacity,
but the `limit` will be 96%.
Do you think there's places where we would like to allow it to use more than 96% RAM?
Do you think we should **reserve** more than 80% for prometheus?

This setting seems reasonable to me, and I even think it's a rather small step: we could even be more extreme.
But if I did not convince you, let's talk about it! :grinning:

QuentinBisson: I was also thinking about being a bit more drastic and reserving less on bigger nodes tbh because 96% will prevent other pods from running

QuentinBisson: We can merge this, monitor and reduce if necessary

service/controller/resource/monitoring/verticalpodautoscaler/resource.go
hervenicol: I don't know how, and I guess it would be more complex.

It says "AsInt64 returns a representation of the current value as an int64 if a fast conversion is possible. If false is returned, callers must use the inf.Dec form of this quantity."
Long story short: as far as I understand these functions, they don't want to lose data/precision.
So I have to do it myself, with this conversion.

QuentinBisson: Sure then let's go

QuentinBisson (approved):

PR #1193, opened 2023-03-14
Prometheus resources: set requests=limits.

hervenicol:
Changed prometheus resources: set requests=limits.
Still allowing prometheus up to 90% of node capacity.

This means,
- Before:
  - initial values: request=1G, limit=1.2G
  - Maximum value: request=75% node capacity, limit=90% node capacity.
- After:
  - initial values: request=limit=1G
  - Maximum value: request=limit=90% node capacity.

Towards https://github.com/giantswarm/giantswarm/issues/26064

## Checklist

I have:


QuentinBisson:
Makes sense but should we not set a bigger value at the beginning? It would be nice to know across all our installations what is the smallest Prometheus memory we have and maybe use this as a baseline to reduce VPA steps?

QuentinBisson (approved):

hervenicol:
> Makes sense but should we not set a bigger value at the beginning? It would be nice to know across all our installations what is the smallest Prometheus memory we have and maybe use this as a baseline to reduce VPA steps?

Your question makes sense, but requires more work and should be addressed in https://github.com/giantswarm/giantswarm/issues/25767

QuentinBisson:
Sure thing that's why i still approved

PR #1196, opened 2023-03-14
implements prometheus.getClusterNodeCount

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/25767

This PR _intend_ to changes Prometheus request memory to be dynamically computed according the target cluster size.

Cluster size is defined by the amount of nodes in the cluster Prometheus will monitor.
This information should be available in MachineDeployment or MachinePool depending of provider (see [slack)](https://gigantic.slack.com/archives/C01F7T2MNRL/p1678460016257829)

This PR tries to read those resources and to find the highest node count and otherwise fallback to 1.

TODO: actually implement the calculation of Prometheus request memory


service/controller/resource/monitoring/prometheus/node_count.go
```diff
@@ -0,0 +1,86 @@
...
+
+	var nodeCount int32 = 1
+	if mdCount > nodeCount {
+		nodeCount = mdCount
+	}
+	if mpCount > nodeCount {
+		nodeCount = mpCount
+	}
+
+	// TODO: Is this ok ? Are we loosing precision here going from int32 to int ?
```
QuentinBisson, line 34: I don't think we lose any kind of precision with integers :)

service/controller/resource/monitoring/prometheus/node_count.go
```diff
@@ -0,0 +1,86 @@
...
+		return 0, microerror.Mask(err)
+	}
+
+	var nodeCount int32 = 1
+	if mdCount > nodeCount {
+		nodeCount = mdCount
+	}
+	if mpCount > nodeCount {
+		nodeCount = mpCount
+	}
```
QuentinBisson, line 32: I would think that you would be getting the number of nodes of the bigger machine deployment or machinepool here when you probably want to add them together

service/controller/resource/monitoring/prometheus/node_count.go
```diff
@@ -0,0 +1,86 @@
...
+		key.ClusterLabel: key.ClusterID(cluster),
+	})
+	opts = append(opts, client.InNamespace(key.OrganizationNamespace(cluster)))
+
+	err := k8sClient.CtrlClient().List(ctx, &mds, opts...)
+	if err != nil {
+		return 0, microerror.Mask(err)
+	}
+
+	var highestNodeCount int32 = 0
```
QuentinBisson, line 52: Same I think we do not want to know the highest node count but sum of all the nodes

service/controller/resource/monitoring/prometheus/node_count.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return 0, microerror.Mask(err)
+	}
+
+	mpCount, err := getClusterNodeCountFromMachinePool(ctx, k8sClient, cluster)
+	if err != nil {
+		return 0, microerror.Mask(err)
+	}
+
+	var nodeCount int32 = 1
```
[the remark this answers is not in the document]
QuentinBisson: I would go with 4 at a minimum here :)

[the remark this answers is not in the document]
QuentinBisson: wait, customers can have 0 worker nodes if they want, so 1 might be good 🤔

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -102,6 +104,12 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 		return nil, microerror.Mask(err)
 	}
 
+	// TODO: use the return value to compute request memory for Prometheus.
+	_, err = getClusterNodeCount(ctx, config.K8sClient, cluster)
```
QuentinBisson, line 108: Could we not name this countClusterNodes?

service/controller/resource/monitoring/prometheus/node_count.go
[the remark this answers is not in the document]
TheoBrigitte: Yes we are only counting worker nodes here. There's no way to get master nodes count on vintage AFAIK. And I did not get how to get this info on CAPI either.

I agree 1 is probably no a good default, but that's what we currently have more or less. But let's test and find a better value here.

service/controller/resource/monitoring/prometheus/node_count.go
TheoBrigitte: Just a self reminder to check int32 vs int precision

service/controller/resource/monitoring/prometheus/node_count.go
TheoBrigitte: Here I am getting the highest number found between MachineDeployment and MachinePool. Adding them together feels wrong.

service/controller/resource/monitoring/prometheus/node_count.go
TheoBrigitte: This is a best effort at trying to find the node count, since we do not really know where the information is I am trying to find the highest value. I think some provider might have both MachineDeployment and MachinePool, I am not sure what adding those values would mean. Needs to be tested.

service/controller/resource/monitoring/prometheus/resource.go
TheoBrigitte: Yes sure

service/controller/resource/monitoring/prometheus/node_count.go
QuentinBisson: The thing here is that I do not understand what you want to achieve tbh. If you want to count all the nodes, you have to sum the number of nodes for all machine pools of a cluster (because customers can have multiple node pools for different reasons) + the number of nodes for all the machine deployments so we can correlate the memory per node.

What you are getting right now is the highest value of the bigger machine pool or machine deployment but if customers have 5 machine pools of 20 nodes each, using 20 as a base value feels wrong.

For CAPI, you will have to ensure you do not fetch the control plane node machine pool/deployment.

service/controller/resource/monitoring/prometheus/node_count.go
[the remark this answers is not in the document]
hervenicol: > I agree 1 is probably no a good default, but that's what we currently have more or less.

I don't think it can relate to "what we currently have".
Today we start with 1G, and that's it.
But maybe the right sizing is like: 500MB / node, so "what we currently have" would be 2?

service/controller/resource/monitoring/prometheus/resource.go
```diff
@@ -141,6 +144,22 @@ func toPrometheus(ctx context.Context, v interface{}, config Config) (metav1.Obj
 
 	labels[key.MonitoringLabel] = "true"
 
+	nodeCount, err := countClusterNodes(ctx, config.K8sClient, cluster)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	prometheusMemory := key.PrometheusDefaultMemory()
+	if nodeCount > 2 {
```
QuentinBisson, line 153: We set a minimum of 1Gi or RAM if node count is less than 2 nodes but then we multiply 500MB per node

QuentinBisson:
Default configured memory for prometheus on gauss:
```
k get prometheus -A -oyaml -w | grep memory
      memory: "1073741824"
      memory: "1073741824" <- 1 node cluster, we default to 1Gi
      memory: "1610612736"
      memory: "1610612736" <- 3 node cluster
      memory: "1073741824"
      memory: "1073741824" <- 1 node cluster, we default to 1Gi
```

service/controller/resource/monitoring/prometheus/node_count.go
TheoBrigitte: That's alright
image
https://pkg.go.dev/builtin#int

TheoBrigitte (commented):
Sounds alright to me.

Can we test this on larger cluster (e.g. 10 nodes) and make sure Prometheus does not restart too much ( less than 3 restarts) due to vpa adjusting memory ?

QuentinBisson:
This has no effect when a cluster is created because it takes too long so Prometheus new memory is scaled down to VPA's min allowed value. As we are planning on moving to mimir, i'd not investigate further. @TheoBrigitte what do you think?

TheoBrigitte:
> This has no effect when a cluster is created because it takes too long so Prometheus new memory is scaled down to VPA's min allowed value. As we are planning on moving to mimir, i'd not investigate further. @TheoBrigitte what do you think?

Alright let's move on.

PR #1213, opened 2023-03-21
Drop uid label from Kube-state-metrics metrics

QuentinBisson:

In an effort to reduce prometheus resource usage, this PR removes the uid label from kube-state-metrics metrics as it is unused

## Checklist

I have:


QuentinBisson:
Also removed the container_id label as it looks to be a high contender for memory usage on big clusters and it's really not used

QuentinBisson:
image

mcbenjemaa (approved):
LGTM

Perhaps, we can monitor this in our gorilla cluster.

QuentinBisson:
Sure, but how? Gorilla is production so we cannot deploy there

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -517,6 +517,20 @@
   - source_labels: [__name__]
     regex: (nginx_ingress_controller_bytes_sent_bucket|nginx_ingress_controller_request_size_bucket|nginx_ingress_controller_response_duration_seconds_bucket|nginx_ingress_controller_response_size_bucket)
     action: drop
+  # drop uid label from kube-state-metrics
+  - source_labels: [app,uid]
+    separator: ;
+    regex: kube-state-metrics;(.+)
+    replacement: ""
+    action: replace
```
marieroque, line 525: You prefer using `replace` than `drop` in case those labels are used somewhere ?

QuentinBisson: Well i want to remove those labels from ksm only, Hence thé 2 source labels. If i were to use drop, it would also drop the app label 😅

hervenicol (approved):
And the most beautiful: as it's in PMO it's gonna be effective fast! :heart_eyes_cat:

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
QuentinBisson: Sure but I think labeldrop would also drop the app label right? Maybe you Can fix it tomorrow, but it's useful to add it to the prometheus-operator pr to enable ksm

QuentinBisson: Added here but needs to be tested

marieroque: I love that :-)

PR #1214, opened 2023-03-21
Update default image for prometheus to latest

mcbenjemaa:


mcbenjemaa:
The intention is to test this change on various installations. 
i will test this change on MC, and WCs as well as the agent.

QuentinBisson:
Did you test on an installation? If yes let's go :)

QuentinBisson (approved):

PR #1219, opened 2023-03-23
remove id and name labels from cAdvisor metrics

QuentinBisson:

This PR removes 2 of the most memory hungry labels in big clusters because they are unique per cgroup so this causes a lot of cardinality.

The id label is only used  in the private dashboard https://github.com/giantswarm/dashboards/blob/master/helm/dashboards/dashboards/shared/private/kubernetes.json

There is no reason to keep the id label because:
- It creates a lot of churn in prometheus
- The value is not really interesting because we are more interested in finding out the container that is causing issues as opposed to the cgroup
- The system services panel in the dashboard are currently broken (system slice does not exist since we migrated to containerd instead of docker, and maybe it did not even exist prior to that):
image
- The All Process panels are broken on most installations and not useful in anyway where it works:
image

@giantswarm/team-honeybadger  as you own this dashboard, you might want to create an issue to revisit it or deprecate it if no one is using it.

## Checklist

I have:

### DIFF
--- a/.nancy-ignore
+++ b/.nancy-ignore
@@ -16,3 +16,6 @@ sonatype-2021-1485 until=2023-04-01
 
 # pkg:golang/k8s.io/apiserver@v0.25.4
 sonatype-2022-6522 until=2023-04-01
+
+# pkg:golang/k8s.io/apiserver@v0.25.4
+CVE-2020-8561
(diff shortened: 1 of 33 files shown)


TheoBrigitte:
I saw you had some interesting result with this, can you post them here also please ?

QuentinBisson:
The results were for the previous PR removing other labels from KSM :D

QuentinBisson:
Monitoring on Gauss but the effect will probably be small as we do not have a lot of nodes

QuentinBisson:
Not that visible on gauss, lets check on anteater

PR #1231, opened 2023-03-27
Clean up unused labels

QuentinBisson:

Drops UID label from kubelet
Drops image_id label from ksm

## Checklist

I have:


hervenicol (commented):
For dropping labels, what is the difference between `labeldrop` and the `replace` magic?

QuentinBisson:
It's done différently for ksm because it's in the workload job so we do not drop the app label

marieroque (approved):

PR #1243, opened 2023-03-29
Drop another bunch of unused metrics

QuentinBisson:
Let's drop another bunch of unused metrics to reduce prometheus stability issues

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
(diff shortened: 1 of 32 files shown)


QuantumEnigmaa (approved):
Maybe the nginx ones will be useful in the future but if they're not currently used let's go on with removing them

QuentinBisson:
@QuantumEnigmaa I'm keeping the one with 1M series that is the only one used :D
image

QuentinBisson:
Hopefully, we can add them back once we have mimir

Gacko:
tbh: I do not remind using them. Maybe some other @giantswarm/team-cabbage does. But yeah, let's drop and re-add when required.

Gacko (approved):

PR #1246, opened 2023-04-03
Allow configuration of prometheus properties

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/26238

This PR adds flexibility to PMO so we can test a new image without releasing a new version of Prometheus.

This PR also removes some unused flags

## Checklist

I have:

### DIFF
--- a/.github/workflows/zz_generated.check_values_schema.yaml
+++ b/.github/workflows/zz_generated.check_values_schema.yaml
@@ -1,6 +1,6 @@
 # DO NOT EDIT. Generated with:
 #
-#    devctl@5.19.0
+#    devctl@5.21.0
 #
 name: 'Values and schema'
 on:
--- a/.github/workflows/zz_generated.create_release.yaml
+++ b/.github/workflows/zz_generated.create_release.yaml
@@ -1,6 +1,6 @@
 # DO NOT EDIT. Generated with:
 #
-#    devctl@5.19.0
+#    devctl@5.21.0
 #
 name: Create Release
 on:
@@ -122,8 +122,8 @@ jobs:
           fi
       - name: Set up git identity
         run: |
-          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          git config --local user.name "github-actions[bot]"
+          git config --local user.email "dev@giantswarm.io"
+          git config --local user.name "taylorbot"
       - name: Commit changes
         run: |
           file="${{ needs.gather_facts.outputs.project_go_path }}"
@@ -131,13 +131,13 @@ jobs:
           git commit -m "Bump version to ${{ steps.update_project_go.outputs.new_version }}"
       - name: Push changes
         env:
-          REMOTE_REPO: "https://${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}.git"
+          REMOTE_REPO: "https://${{ github.actor }}:${{ secrets.TAYLORBOT_GITHUB_ACTION }}@github.com/${{ github.repository }}.git"
           branch: "${{ github.ref }}-version-bump"
         run: |
           git push "${REMOTE_REPO}" HEAD:${{ env.branch }}
       - name: Create PR
         env:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          GITHUB_TOKEN: "${{ secrets.TAYLORBOT_GITHUB_ACTION }}"
           base: "${{ github.ref }}"
           branch: "${{ github.ref }}-version-bump"
           version: "${{ needs.gather_facts.outputs.version }}"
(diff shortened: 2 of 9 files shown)


helm/prometheus-meta-operator/templates/configmap.yaml
```diff
@@ -38,15 +38,17 @@ data:
         {{- range .Values.prometheus.bastions }}
         - {{ . }}
         {{- end }}
-        evaluationInterval: 60s
+        evaluationInterval: {{ .Values.prometheus.evaluationInterval }}
```
hervenicol, line 41: `scrapeInterval` and `evaluationInterval` were already configurable by command-line arguments, right?
What is the difference now?

hervenicol (commented):
I didn't see the changes concerning the image version.

But I think currently we can give a specific image tag, but it would be even better to give a specific full image name. So we can test upstream images without retagging, quite useful for tests.

helm/prometheus-meta-operator/templates/configmap.yaml
QuentinBisson: Now we can set them via giantswarm/config

QuentinBisson: I can check how we Can overrode thé image but i'm not sure we can

QuentinBisson: So I did not explain properly, sorry I was managing the kids. So the values are overridable from cli-arguments but we configured it in a way that the configmap actually fills out the cli args so values we set here are picked up by cobra (the go cli library) as cli args.

So what we set as values here will be set in main.go.

Regarding the image change, we cannot currently override the image repository but only the registry as per https://github.com/giantswarm/prometheus-meta-operator/blob/41bc5acf00a1500b0e7df4299733dc823f996144/service/controller/resource/monitoring/prometheus/resource.go#L144

The values from Config are injected by the configmap

helm/prometheus-meta-operator/values.yaml
```diff
@@ -54,13 +54,24 @@ prometheus:
...
 
+  retention:
+    ## Retention duration for prometheus.
+    duration: "2w"
+
+  ## Default scrape interval for prometheus jobs.
+  scrapeInterval: "60s"
+
+  ## Prometheus container image version.
+  # version: ""
```
QuentinBisson, line 73: @hervenicol if we set prometheus.version here it will be used everywhere. I can add the repository also

.nancy-ignore
```diff
@@ -6,16 +6,16 @@
 #  - github.com/giantswarm/microendpoint@v1.0.0
 #  - github.com/giantswarm/microkit@v1.0.0
 #  - sigs.k8s.io/cluster-api@v1.0.5
-CVE-2022-29153 until=2023-04-01
-CVE-2021-41803 until=2023-04-01
-CVE-2022-24687 until=2023-04-01
+CVE-2022-29153 until=2023-06-01
+CVE-2021-41803 until=2023-06-01
```
QuentinBisson, line 10: Had to extend CVE ignorance

QuentinBisson, line 10: Had to extend CVE ignorance

flag/service/alertmanager/alertmanager.go
```diff
@@ -1,13 +0,0 @@
-package alertmanager
-
-type Alertmanager struct {
-	Address    string
```
QuentinBisson, line 4: Not needed since we moved to the helm chart

flag/service/prometheus/prometheus.go
```diff
@@ -6,24 +6,14 @@ type Prometheus struct {
 	BaseDomain              string
 	Bastions                string
 	EvaluationInterval      string
+	ImageRepository         string
```
QuentinBisson, line 9: Adding support for repository

helm/prometheus-meta-operator/templates/alertmanager/alertmanager.yaml
```diff
@@ -23,7 +23,8 @@ spec:
       app.kubernetes.io/name: alertmanager
   configSecret: alertmanager-config
   externalUrl: {{ .Values.alertmanager.address }}
-  logLevel: info
+  image: {{ printf "%s/%s:%s" .Values.registry.domain .Values.alertmanager.imageRepository .Values.alertmanager.version }}
```
QuentinBisson, line 26: Needed to work in China

QuentinBisson:
I'll stop here because this is getting out of hand. I keep finding issues. If you prefer me to split this into smaller PRs let me know

helm/prometheus-meta-operator/values.yaml
QuentinBisson: imageRepository added

service/controller/resource/monitoring/prometheus/test/case-1-awsconfig.golden
```diff
@@ -113,6 +113,7 @@ spec:
     topologyKey: kubernetes.io/hostname
     whenUnsatisfiable: ScheduleAnyway
   tsdb: {}
+  version: v2.28.1
```
QuentinBisson, line 116: That way we get the version displayed also when using kubectl

marieroque (approved):

PR #1265, opened 2023-04-14
Only route page and/or notify alert to Atlas slack

TheoBrigitte:
This PR updates the Atlas slack alert routing to only route alert to slack when :

- severity=page on stable installations
- severity~=notify|page on other installations

This way we can keep some alerts in Alertmanager and not send them anywhere.

## Checklist

I have:

### DIFF
--- a/files/templates/alertmanager/alertmanager.yaml
+++ b/files/templates/alertmanager/alertmanager.yaml
@@ -63,7 +63,9 @@ route:
   - receiver: team_atlas_slack
     matchers:
     [[- if eq .Pipeline "stable" ]]
-    - severity!="page"
+    - severity="notify"
+    [[- else ]]
+    - severity=~"page|notify"
     [[- end ]]
     - team="atlas"
     - type!="heartbeat"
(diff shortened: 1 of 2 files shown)


files/templates/alertmanager/alertmanager.yaml
```diff
@@ -63,7 +63,9 @@ route:
   - receiver: team_atlas_slack
     matchers:
     [[- if eq .Pipeline "stable" ]]
-    - severity!="page"
+    - severity="notify"
```
marieroque, line 66: You describe the opposite in the PR: severity=page

TheoBrigitte: I agree this deserve a comment as the logic is twisted here.

We sent only `notify` alert to slack for **stable** installation, `page` alerts goes into opsgenie.
For other installations (testing) we send notify and page alert to slack, as we do not see them in opsgenie.

marieroque (approved):

hervenicol (approved):

hervenicol:
Fine with me if it builds :sweat_smile:

PR #1267, opened 2023-04-18
Fix Observability Bundle detection for vintage WCs

QuentinBisson:
We currently have duplicate targets because the observability bundle app for vintage WC is not where pmo is looking for it

## Checklist

I have:


hervenicol (approved):
LGTM

QuentinBisson:
Tested on Gauss

PR #1271, opened 2023-04-21
Add Inhibition rule when prometheus-agent is down

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/26678

To avoid several types of alerts to fire because metrics are missing, we create an inhibition when prometheus-agent is down.

## Checklist

I have:

### DIFF
--- a/files/templates/alertmanager/alertmanager.yaml
+++ b/files/templates/alertmanager/alertmanager.yaml
@@ -508,9 +508,9 @@ inhibit_rules:
   equal: [cluster_id]
 
 - source_matchers:
-    - cluster_is_prometheus_agent_down=true
+    - prometheus_agent_down=true
   target_matchers:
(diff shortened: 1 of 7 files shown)


hervenicol (approved):
LGTM

files/templates/alertmanager/alertmanager.yaml
```diff
@@ -507,6 +507,12 @@ inhibit_rules:
     - cancel_if_cluster_is_not_running_prometheus_agent=true
   equal: [cluster_id]
 
+- source_matchers:
+    - inhibition_prometheus_agent_down=true
```
QuentinBisson, line 511: I'm confused, the new doc says it should be inhibit_prometheus_agent_down https://github.com/giantswarm/prometheus-rules/pull/709/files

QuentinBisson (approved):

PR #1280, opened 2023-05-01
heartbeat tuning

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/26770

- alertmanager heartbeats config updated for faster response when receiving heartbeats, and less stress on opsgenie
- opsgenie now waits for 1h with no heartbeat before sending an alert, instead of 25min previously
- runbook link style is consistent with other alerts

## Checklist

I have:


QuentinBisson (approved):

hervenicol:
Tested on gauss, here is the new config on opsgenie:
image

PR #1286, opened 2023-05-02
Missing kubelet node label

QuentinBisson:
## Checklist

I have:

### DIFF
--- a/files/templates/alertmanager/alertmanager.yaml
+++ b/files/templates/alertmanager/alertmanager.yaml
@@ -345,7 +345,7 @@ receivers:
 
 inhibit_rules:
 - source_matchers:
-  - kube_state_metrics_down=true
+  - inhibit_kube_state_metrics_down=true
   target_matchers:
   - cancel_if_kube_state_metrics_down=true
   equal: [cluster_id]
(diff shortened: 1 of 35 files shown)


whites11 (approved):

files/templates/alertmanager/alertmanager.yaml
```diff
@@ -345,7 +345,7 @@ receivers:
 
 inhibit_rules:
 - source_matchers:
-  - kube_state_metrics_down=true
+  - inhibit_kube_state_metrics_down=true
```
marieroque, line 348: Do you have the related PR on prometheus-rules for this change ?

QuentinBisson: https://github.com/giantswarm/prometheus-rules/pull/733#discussion_r1182352088

QuantumEnigmaa (approved):

files/templates/alertmanager/alertmanager.yaml
marieroque: Why doesn't the prometheus-rules PR set `cancel_if_inhibit_kube_state_metrics_down` everywhere `cancel_if_kube_state_metrics_down` is set ?

QuentinBisson: Because we decided to use cancel_if_kube_state_metrics_down and not cancel_if_inhibit_kube_state_metrics_down right?

marieroque: I'm loosing my mind...

PR #1311, opened 2023-06-07
Remove cpu requests=limit ratio

QuentinBisson:
Towards https://github.com/giantswarm/babymarkt/issues/1213

This PR removes Prometheus requests/limits cpu ratio so VPA does not overprovision CPU on machines

## Checklist

I have:


QuentinBisson:
This is not a working solution because Prometheus gets throttled and VPA does not do anything because it does not really detects CPU spikes ........

PR #1321, opened 2023-06-19
Remove image pull secret

QuentinBisson:
Prometheus-meta-operator app is failing to deploy since the imagepullsecret was removed from giantswarm/config - https://github.com/giantswarm/giantswarm/issues/27267.

As we should now rely on the registry configuration present on the nodes, this PR removes this secret to fix PMO deployments

## Checklist

I have:


QuentinBisson:
cc @fiunchinho as its related to your change

fiunchinho:
Ooops, I missed this one, sorry.

fiunchinho (changes requested):
You also need to remove this https://github.com/giantswarm/prometheus-meta-operator/blob/master/helm/prometheus-meta-operator/templates/_resource.tpl#L22-L24

PR #1334, opened 2023-06-28
Change shard computation based on the number of head series

QuentinBisson:
## Checklist

I have:


QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/27270 This PR actually adds a shard for each 1M time series

hervenicol (approved):
Wow, that was fast :clap: 

LGTM overall.

I guess prometheus URL computation will soon have to be able to select between cluster prometheus or Mimir, and we'll have to write a function for that.
But for a first version, this should do the job!

PR #1340, opened 2023-07-03
Fix race condition between Prometheus and Shard configuation

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/27461

## Checklist

I have:

### DIFF
--- a/service/controller/resource/monitoring/remotewriteconfig/resource.go
+++ b/service/controller/resource/monitoring/remotewriteconfig/resource.go
@@ -2,6 +2,8 @@ package remotewriteconfig
 
 import (
 	"context"
+	"errors"
+	"net"
 
 	"github.com/giantswarm/k8sclient/v7/pkg/k8sclient"
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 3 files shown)


service/controller/resource/monitoring/remotewriteconfig/resource.go
```diff
@@ -133,6 +135,12 @@ func (r *Resource) desiredConfigMap(cluster metav1.Object, name string, namespac
 func (r *Resource) getShardsCountForCluster(ctx context.Context, cluster metav1.Object, currentShardCount int) (int, error) {
 	headSeries, err := prometheusquerier.QueryTSDBHeadSeries(key.ClusterID(cluster))
 	if err != nil {
+		// If prometheus is not accessible (for instance, not running because this is a new cluster, we check if prometheus is accessible)
+		var dnsError *net.DNSError
```
QuentinBisson, line 139: Coming from https://stackoverflow.com/questions/41669475/golang-idiomatic-way-to-detect-no-such-host-error

PR #1348, opened 2023-07-10
Stable-testing pipeline only route MC alerts to opsgenie

ericgraf:
This change routes only MC alerts to opsgenie when the pipeline is set to stable-testing for the MC.

towards: https://github.com/giantswarm/giantswarm/issues/26718
Related other changes:
- https://github.com/giantswarm/prometheus-rules/pull/825
- https://github.com/giantswarm/installations/pull/2208
- https://github.com/giantswarm/config/pull/1877

## Checklist

I have:

- [x] Described why this change is being introduced
- [ ] Separated out refactoring/reformatting in a dedicated PR
- [x] Updated changelog in `CHANGELOG.md`

PR #1358, opened 2023-08-03
fix null receiver name

TheoBrigitte:
Replace the null receiver introduced in #1348.

The receiver name cannot be null it has to be a string, so I replaced this with `blackhole`.
When the received is null the following error occur when prometheus-operator tries to configure alertmanager:
```
level=error ts=2023-08-03T18:54:53.770986178Z caller=klog.go:116 component=k8s_client_runtime func=ErrorDepth msg="sync \"monitoring/alertmanager\" failed: provision alertmanager configuration: failed to initialize from secret: missing name in receiver"
```

This lead to many failing heartbeat.


marieroque (approved):

hervenicol (approved):
LGTM

PR #1364, opened 2023-08-30
Support flux managed clusters

QuentinBisson:
Towards: https://github.com/giantswarm/roadmap/issues/2809

To connect to the api server of CAPI Clusters,  prometheus-meta-operator is copying the kubeconfig for the cluster (certificate and private key) to be able to establish a TLS connection.

For Flux managed clusters described in this issue https://github.com/giantswarm/roadmap/issues/2670, the kubeconfig contains a token and not a certificate/private key pair so we need to adjust prometheus-meta-operator to be able to configure the Prometheus to use a token to connect to the API server instead of the certificates.

This PR does just that

## Checklist

I have:

### DIFF
--- a/files/templates/scrapeconfigs/_apiserver.yaml
+++ b/files/templates/scrapeconfigs/_apiserver.yaml
@@ -1,6 +1,8 @@
 [[- define "_apiserver" -]]
 [[- if ne .ClusterType "management_cluster" ]]
     api_server: https://[[ .APIServerURL ]]
+    authorization:
+      credentials_file: /etc/prometheus/secrets/[[ .SecretName ]]/token
     tls_config:
       ca_file: /etc/prometheus/secrets/[[ .SecretName ]]/ca
       cert_file: /etc/prometheus/secrets/[[ .SecretName ]]/crt
--- a/files/templates/scrapeconfigs/_apiserver.yaml
+++ b/files/templates/scrapeconfigs/_apiserver.yaml
@@ -1,12 +1,15 @@
 [[- define "_apiserver" -]]
 [[- if ne .ClusterType "management_cluster" ]]
     api_server: https://[[ .APIServerURL ]]
-    authorization:
-      credentials_file: /etc/prometheus/secrets/[[ .SecretName ]]/token
+[[- if eq .AuthenticationType "token" ]]
+    bearer_token_file: /etc/prometheus/secrets/[[ .SecretName ]]/token
+[[- end ]]
     tls_config:
       ca_file: /etc/prometheus/secrets/[[ .SecretName ]]/ca
+[[- if eq .AuthenticationType "certificates" ]]
       cert_file: /etc/prometheus/secrets/[[ .SecretName ]]/crt
       key_file: /etc/prometheus/secrets/[[ .SecretName ]]/key
+[[- end ]]
       insecure_skip_verify: false
 [[- end -]]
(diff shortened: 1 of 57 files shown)


QuentinBisson:
Tested on Snail, working fine :)

files/templates/scrapeconfigs/_apiserver.yaml
```diff
@@ -1,10 +1,15 @@
 [[- define "_apiserver" -]]
 [[- if ne .ClusterType "management_cluster" ]]
     api_server: https://[[ .APIServerURL ]]
+[[- if eq .AuthenticationType "token" ]]
```
QuentinBisson, line 4: Add a new field named authentication type that explains if the tls config needs to use a token or a cert/key pair

files/templates/scrapeconfigs/additional-scrape-configs.template.yaml
```diff
@@ -177,16 +177,10 @@
...
-    tls_config:
-      ca_file: /etc/prometheus/secrets/[[ .SecretName ]]/ca
-      cert_file: /etc/prometheus/secrets/[[ .SecretName ]]/crt
-      key_file: /etc/prometheus/secrets/[[ .SecretName ]]/key
-      insecure_skip_verify: false
-[[ else ]]
+[[ include "_apiserver" . ]]
+[[- if eq .ClusterType "management_cluster" ]]
   bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
-[[ end ]]
```
marieroque, line 189: why these 3 above lines are not embedded in the _apiserver section ?

QuentinBisson: That's because etcd is a bit special, especially on vintage

QuentinBisson: Also the apiserver secion is only for workload clusters because prometheus in the MC have access to the api server "locally"

marieroque: On MC there is always a file with the token ?

service/key/key.go
```diff
@@ -298,3 +300,17 @@ func ClusterType(installation string, obj interface{}) string {
...
+	if err != nil {
+		return "", err
+	}
+
+	if secret.Data["token"] != nil && len(secret.Data["token"]) > 0 {
+		return "token", nil
+	} else if (secret.Data["crt"] != nil && len(secret.Data["crt"]) > 0) && (secret.Data["key"] != nil && len(secret.Data["key"]) > 0) {
+		return "certificates", nil
+	}
+	return "vintage", nil
```
marieroque, line 315: What does "vintage" mean ?
I see no use case with this authentication type ?!

QuentinBisson: Well, it means the vintage authentication mechanism (for KVM). It's not used but i wanted to avoid setting an empty string here

marieroque: Can you add a comment please :-)

QuentinBisson: I made it into an error :)

QuentinBisson:
@giantswarm/team-atlas The changes should be easier to review :)

QuentinBisson:
Tested on `gauss`, `snail` and `grizzly`

marieroque (approved):

PR #1365, opened 2023-08-31
Enable Shield opsgenie alerts

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/27693, we have combined with Honeybadger for oncall, so we can now let Shield alerts through to the opsgenie receiver.

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Enable Opsgenie alerts for Shield.
+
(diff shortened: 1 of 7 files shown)


fhielpos (approved):
We will continue receiving them to the alerts channel too, right?

stone-z:
> We will continue receiving them to the alerts channel too, right?

Yes. Before this the `continue: false` just meant that the alerts would stop at the slack receiver, which was chronologically before the opsgenie receiver. Now they will first go to the opsgenie receiver and then to our slack receiver.

stone-z:
Ok repo says we gotta wait for @giantswarm/team-atlas

hervenicol (approved):
LGTM

PR #1366, opened 2023-09-04
Add app tag to OpsGenie alert

marians:
Towards https://github.com/giantswarm/giantswarm/issues/28076

This PR adds a tag to enable linking from a developer portal component page to alerts regarding the component. Example:

`app:app-exporter`

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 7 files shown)


marians:
So in my previous iteration I just emitted the tag without condition. This resulted in an alert with the tag `app:`.

image

So I made the tag output conditional.

marians:
It turns out that this PR is unnecessary. Querying for `app:NAME` via the API works already, obviously using the extra properties. ☺️

PR #1373, opened 2023-09-19
improve agent shards calculation

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/28174

Improve agent shards calculation:
we base our calculations on the max number of series for the last 6 hours, so it changes less often.

## Checklist

I have:


hervenicol:
Tested on `gauss` with small WCs: no error in PMO, but as the clusters are rather small there's always only 1 shard.

marieroque (approved):

QuentinBisson (approved):
LGTM

PR #1379, opened 2023-09-21
Add kube-proxy to the list of ignored targets

QuentinBisson:
Towards monitoring of EKS clusters, we noticed that the imported EKS cluster (on snail) has kube-proxy running and we would prefer to move all EKS scraping to use ServiceMonitors instead of going through the api server proxy (related to https://github.com/giantswarm/roadmap/issues/2832)

We figured with @TheoBrigitte that it is safer to have a ServiceMonitor that is not always needed than to have it being scraped from outside the cluster.

Hence this PR in relation with https://github.com/giantswarm/observability-bundle/pull/113 will move the scraping of a possible kube-proxy to be done by the agent instead of externally to the cluster.

This PR is also in close relation with https://github.com/giantswarm/roadmap/issues/2670

⚠️ Do not merge until https://github.com/giantswarm/observability-bundle/pull/113 is released in 0.8.3

## Checklist

I have:


service/controller/resource/monitoring/scrapeconfigs/resource.go
```diff
@@ -296,6 +297,10 @@ func listTargetsToIgnore(ctx context.Context, ctrlClient client.Client, cluster
 		if err != nil {
 			return nil, microerror.Mask(err)
 		}
+		bundleWithKubeProxyExporterVersion, err := semver.Parse("0.8.3")
```
QuentinBisson, line 300: Note that this will be released once https://github.com/giantswarm/observability-bundle/pull/113 is merged

marieroque (approved):
👍🏽

PR #1387, opened 2023-09-25
Ignore PMO reconciliation issues on stable-testing installations

QuentinBisson:
Grizzly clusters are causing a lot of PMO reconcilation errors because clusters are created a lot and some with issues. Let's not get paged everytime there is an issue on grizzly

## Checklist

I have:


QuantumEnigmaa (approved):
Good idea !

AverageMarcus (approved):

PR #1402, opened 2023-09-27
Add ClusterIssuerName configuration option

ubergesundheit:
This PR introduces a new configuration option to specify the name of the cert-manager ClusterIssuer to be used on Ingress resources installed by the chart in this repository.

Towards https://github.com/giantswarm/roadmap/issues/2844

## Checklist

I have:

### DIFF
--- a/helm/prometheus-meta-operator/templates/alertmanager/alertmanager-ingress.yaml
+++ b/helm/prometheus-meta-operator/templates/alertmanager/alertmanager-ingress.yaml
@@ -2,7 +2,12 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:
-    kubernetes.io/tls-acme: "{{ .Values.prometheus.letsencrypt }}"
+    {{- if .Values.prometheus.letsencrypt }}
+    kubernetes.io/tls-acme: "true"
+    {{- with .Values.prometheus.clusterIssuerName }}
+    cert-manager.io/cluster-issuer: "{{ . }}"
+    {{- end }}
+    {{- end }}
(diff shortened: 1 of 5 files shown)


QuentinBisson:
Is there a reason why we might need this?

ubergesundheit:
> Is there a reason why we might need this?

cabbage recently updated cert-manager templates and configuration to follow upstream more closely. One of the changes is that cert-manager does not have a default issuer selected. This makes sure Ingresses can still obtain TLS certificates through cert-manager + let's encrypt by explicitly specifying which ClusterIssuer to use.

helm/prometheus-meta-operator/templates/alertmanager/alertmanager-ingress.yaml
```diff
@@ -2,7 +2,12 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:
-    kubernetes.io/tls-acme: "{{ .Values.prometheus.letsencrypt }}"
+    {{- if .Values.prometheus.letsencrypt }}
```
QuentinBisson, line 5: We have a few ingresses that are generated in PMO code. Could you also add it to them?

QuentinBisson:
Thanks I just saw that in the roadmap issue because I did not check it prior to me asking the question :(

helm/prometheus-meta-operator/templates/alertmanager/alertmanager-ingress.yaml
QuentinBisson: Oh right because they rely on the main ones as they are subpaths

QuentinBisson (approved):

PR #1408, opened 2023-09-28
Route ClusterUnhealthyPhase for test clusters to blackhole

AverageMarcus:
The `ClusterUnhealthyPhase` alert should only page for Workload Clusters that aren't E2E test clusters. The best we have to work this out is the cluster name starting with `t-` as is the convention in the test suites.

## Checklist

I have:

### DIFF
--- a/files/templates/alertmanager/alertmanager.yaml
+++ b/files/templates/alertmanager/alertmanager.yaml
@@ -30,6 +30,11 @@ route:
     matchers:
     - alertname="PrometheusMetaOperatorReconcileErrors"
     continue: false
+  - receiver: blackhole
+    matchers:
+    - alertname="ClusterUnhealthyPhase"
+    - name=~"t-.*"
+    continue: false
   [[- end ]]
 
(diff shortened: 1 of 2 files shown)


fiunchinho (approved):

files/templates/alertmanager/alertmanager.yaml
```diff
@@ -30,6 +30,11 @@ route:
     matchers:
     - alertname="PrometheusMetaOperatorReconcileErrors"
     continue: false
+  - receiver: blackhole
+    matchers:
+    - alertname="ClusterUnhealthyPhase"
+    - name=~"t-.*"
```
marieroque, line 36: Are you sure about `name` ?
Isn't it rather `cluster_id` ?

AverageMarcus: `cluster_id` for this alert is the MC. `name` is the WC ID for this alert

marieroque (approved):

PR #1411, opened 2023-09-29
Fix newlines / whitespace trimming if opsrecipe is *not* specified or a dashboard *is* specified

AndiDog:
Check the `#opsgenie` channel and you'll see that `Dashboard:` does not render on its own line.

Similarly, if there was any alert without an opsrecipe (I don't think we have one, so this is minor), the whitespace would also not be trimmed as desired.

Fixing those by putting the `-` trimming indicator on the right spots. I didn't find tests covering the rendered output of templates, so I only regenerated the golden template files.

## Checklist

I have:


hervenicol:
We're probably lacking some tests, true.
And manually testing the alerts templates is a pain.
Especially given we have to make sure it works well with opsgenie AND with Slack, with the same rendered alert.

So kudos for trying to improve them! :grinning:

hervenicol (approved):
Looks fine.
Did you run it on a test installation, and check the generated alerts?

AndiDog:
> Looks fine. Did you run it on a test installation, and check the generated alerts?

@hervenicol How exactly would that work? Installing a newer version of prometheus-meta-operator is clear, but how would I trigger a test alert that ends up in `#opsgenie` chat where the changed templates matter?

hervenicol:
@AndiDog you could for instance deploy a custom prometheusrule, or edit a prometheusrule in-place to make it trigger. Like, by modifying the condition.

AndiDog:
@hervenicol Sorry, I had to give up the testing after spending this morning on it. First, my own `PrometheusRule` didn't get picked up at all. No idea why, since we don't seem to have selectors in place that could filter. Editing an existing rule worked after sometimes having to await its reconciliation, but then it didn't always end up in OpsGenie, making the testing a tiring procedure. And even though I had deployed my latest PMO image, the alerts were still rendered using the old template, it seems. Therefore I used https://gotemplate.io/ to validate my template changes, and hope we can go ahead without manual testing. If it doesn't work in production, I'll notice and follow up. It's just whitespace. An actual rendering unit test would be great to have in order to replace the cumbersome manual testing.

hervenicol:
All right, thanks for the effort and sorry that it's such a mess to test :disappointed: 
At least you've tested it does not break apps :+1: 
I had already approved anyway, feel free to merge and release.

PR #1502, opened 2024-02-05
add cilium netpol creation

QuantumEnigmaa:
Towards https://github.com/giantswarm/giantswarm/issues/29698

:warning:  This is still a work in progress.

This PR enables PMO to create a `ciliumNetworkPolicy` resource for each MC prometheus. 

## Checklist

I have:


QuantumEnigmaa:
Issue with the `prometheus` resource : 
```
cannot use corev1.ResourceRequirements{…} (value of type "k8s.io/api/core/v1".ResourceRequirements) as "k8s.io/api/core/v1".VolumeResourceRequirements value in struct literal
```

**Copilot explanation :** 

"The error message you're seeing is indicating that you're trying to use a value of type `corev1.ResourceRequirements` where a value of type `corev1.VolumeResourceRequirements` is expected.

In the Kubernetes API, `ResourceRequirements` and `VolumeResourceRequirements` are different types and are not interchangeable. `ResourceRequirements` is used to set the CPU and memory requests and limits for a container, while `VolumeResourceRequirements` is used to set the storage requests and limits for a volume.

In your code, you're trying to set the storage requests for a volume, so you should be using `VolumeResourceRequirements`, not `ResourceRequirements`.

However, as of Kubernetes API version v1, there is no `VolumeResourceRequirements` type. Instead, you should use `ResourceRequirements` to set the storage requests for a volume, like you're doing in your code.

If you're seeing this error, it's possible that you're using a different version of the Kubernetes API where `VolumeResourceRequirements` is a valid type. In that case, you should check the documentation for that version of the API to see how to set the storage requests for a volume.

If you're using Kubernetes API version v1, your code is correct and you should not be seeing this error. In that case, the error might be caused by something else in your code. Could you provide more information about your code and the context in which this error is occurring?"

QuantumEnigmaa:
When deploying it : 
```
"stack":{"annotation":"Object 'Kind' is missing in 'unstructured object has no kind'","kind":"unknown","stack: [{"file":"/root/project/service/controller/resource/ciliumnetpol/create.go","line":26},[...]]
```

QuantumEnigmaa:
PMO doesn't seem to like dynamic client : 
```
panic: runtime error: invalid memory address or nil pointer dereference
```

QuantumEnigmaa:
There is some progress  : 
```
"stack":{"annotation":"ciliumnetworkpolicies.cilium.io \"gaia-prometheus\" is forbidden: User \"system:serviceaccount:monitoring:prometheus-meta-operator\" cannot get resource \"ciliumnetworkpolicies\" in API group \"cilium.io\" at the cluster scope","kind":"unknown"
```

QuantumEnigmaa:
Adding the capabilities for cilium netpols in the clusterrole didn't change the error

helm/prometheus-meta-operator/templates/rbac.yaml
```diff
@@ -93,6 +93,16 @@ rules:
       - create
       - delete
       - update
+  - apiGroups:
+      - cilium.io
+    resources:
+      - ciliumnetworkolicies
```
pipo02mix, line 99:
```suggestion
      - ciliumnetworkpolicies
```
you have a typo ;)

QuantumEnigmaa: Oh wow  thx ! I would have definitely missed it !

QuantumEnigmaa:
PMO still failing but for yet another reason : 
```
"message":"retrying due to error","object":"default/kubernetes","resource":"ciliumnetpol","stack":{"annotation":"the server could not find the requested resource","kind":"unknown","stack: [{"file":"/root/project/service/controller/resource/ciliumnetpol/create.go","line":34
```

QuantumEnigmaa:
It finally works, but only for the MC prometheus (this was expected though)

QuantumEnigmaa:
When trying to create the ciliumnetpol for WCs, PMO is panicking with a segmentation error : 
```
2024-02-12T09:20:56Z    INFO    Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference        {"controller": "cluster", "controllerGroup": "cluster.x-k8s.io", "controllerKind": "Cluster", "Cluster": {"name":"gj83r","namespace":"org-zirko"}, "namespace": "org-zirko", "name": "gj83r", "reconcileID": "13c22583-216b-48e5-87c5-3d2a277eb2f1"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x19ffe0d]

goroutine 111 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:116 +0x1e5
panic({0x1ca3220?, 0x32b5d50?})
        /go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.5.linux-amd64/src/runtime/panic.go:914 +0x21f
github.com/giantswarm/prometheus-meta-operator/v2/service/controller/resource/ciliumnetpol.(*Resource).EnsureCreated(0xc000358600, {0x22a0cb8, 0xc00034def0}, {0x1f4a000, 0xc0006bab60})
        /root/project/service/controller/resource/ciliumnetpol/create.go:28 +0x12d
github.com/giantswarm/operatorkit/v7/pkg/resource/wrapper/retryresource.(*basicResource).EnsureCreated.func1()
        /go/pkg/mod/github.com/giantswarm/operatorkit/v7@v7.2.0/pkg/resource/wrapper/retryresource/basic_resource.go:50 +0x39
github.com/cenkalti/backoff/v4.RetryNotifyWithTimer.Operation.withEmptyData.func1()
```

QuantumEnigmaa:
Tested it on `gaia` with a manually created WC and it works : 
```
NAMESPACE          NAME                                                    AGE
gaia-prometheus    gaia-prometheus                                         22s
gj83r-prometheus   gj83r-prometheus                                        22s
```

go.mod
```diff
@@ -24,16 +26,22 @@ require (
 	github.com/spf13/viper v1.18.2
 	golang.org/x/crypto v0.19.0
 	golang.org/x/net v0.21.0
-	k8s.io/api v0.28.4
+	k8s.io/api v0.29.0-rc.1
```
QuentinBisson, line 29: I would not use a RC version here

service/controller/managementcluster/resource.go
```diff
@@ -152,6 +155,19 @@ func newResources(config resourcesConfig) ([]resource.Interface, error) {
 		}
 	}
 
+	var ciliumnetpolResource resource.Interface
+	{
+		c := ciliumnetpol.Config{
+			K8sClient: config.DynamicK8sClient,
```
QuentinBisson, line 161: Maybe call it dynamicclient or just client?

service/controller/resource/ciliumnetpol/create.go
```diff
@@ -0,0 +1,40 @@
...
+			Version:  "v2",
+			Resource: "ciliumnetworkpolicies",
+		}
+
+		desired, err := toCiliumNetworkPolicy(obj)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		//var current unstructured.Unstructured
```
QuentinBisson, line 26: Remove those comments :)

service/service.go
```diff
@@ -117,6 +118,14 @@ func New(config Config) (*Service, error) {
 		}
 	}
 
+	var DynamicK8sClient dynamic.Interface
```
QuentinBisson, line 121:
```suggestion
	var dynamicK8sClient dynamic.Interface
```

go.mod
QuantumEnigmaa: That's the `tidy` command work here, should I just move it to `v0.29.0` ?

service/controller/resource/ciliumnetpol/create.go
QuantumEnigmaa: Done

service/service.go
QuantumEnigmaa: Done

service/controller/managementcluster/resource.go
QuantumEnigmaa: I renamed it `DynamicK8sClient`

go.mod
QuantumEnigmaa: Done

go.mod
```diff
@@ -24,16 +26,22 @@ require (
...
 	golang.org/x/crypto v0.19.0
 	golang.org/x/net v0.21.0
-	k8s.io/api v0.28.4
+	k8s.io/api v0.29.0
 	k8s.io/apiextensions-apiserver v0.28.4
-	k8s.io/apimachinery v0.28.4
+	k8s.io/apimachinery v0.29.0
 	k8s.io/autoscaler/vertical-pod-autoscaler v1.0.0
-	k8s.io/client-go v0.28.4
+	k8s.io/client-go v0.29.0-rc.1
```
QuentinBisson, line 33:
```suggestion
	k8s.io/client-go v0.29.0
```

go.mod
```diff
@@ -114,7 +122,7 @@ require (
 	gopkg.in/resty.v1 v1.12.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	k8s.io/component-base v0.28.4 // indirect
+	k8s.io/component-base v0.29.0-rc.1 // indirect
```
QuentinBisson, line 125:
```suggestion
	k8s.io/component-base v0.29.0 // indirect
```

QuentinBisson (approved):
Apart from the go client then it's alright

QuantumEnigmaa:
So it still works :)

QuentinBisson:
Can you test it on gerbil as well ?

QuantumEnigmaa:
Seems to be working as well : 
```
NAMESPACE           NAME                                                                   AGE
gerbil-prometheus   gerbil-prometheus                                                      6m42s
monitoring          prometheus                                                             6m52s
```

QuentinBisson:
Can you check that alertmanager can communicate with all prometheus ? Also, I think that pormetheus cnp in monitoring is not needed but alright :)

QuantumEnigmaa:
> Can you check that alertmanager can communicate with all prometheus ? 

From what I see, it seems alright as the alerts redirect to the right prometheus instance. Don't know if there are more checks to do :shrug: 

> I think that pormetheus cnp in monitoring is not needed

Yeah but I don't see a quick way to avoid this in the code :/

QuentinBisson:
The cnp is coming from the helmchart 🙈 

Now for alertmanager you can open hubble using port-forwarding and check if there are any dropped connection in the monitoring ns and gerbil-prometheus :) That should be okay

QuantumEnigmaa:
> The cnp is coming from the helmchart 🙈

Oh yeah indeed ! I can remove it if you want :) 

Concerning hubble : 
image

QuentinBisson:
If you do not see any dropped below then let's go

QuantumEnigmaa:
Everything is "forwarded" :)

PR #1506, opened 2024-02-05
fix forgotten verticalpodautoscaler.Config

hervenicol:
@QuentinBisson 
Some fixes for your PMO branch.

### DIFF
--- a/service/controller/resource/resource.go
+++ b/service/controller/resource/resource.go
@@ -267,9 +267,11 @@ func New(config Config) ([]resource.Interface, error) {
 	var verticalPodAutoScalerResource resource.Interface
 	{
 		c := verticalpodautoscaler.Config{
-			Logger:    config.Logger,
-			K8sClient: config.K8sClient,
-			VpaClient: config.VpaClient,
+			Logger:       config.Logger,
+			K8sClient:    config.K8sClient,
+			VpaClient:    config.VpaClient,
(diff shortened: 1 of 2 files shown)


service/controller/resource/resource.go
```diff
@@ -267,9 +267,11 @@ func New(config Config) ([]resource.Interface, error) {
 	var verticalPodAutoScalerResource resource.Interface
 	{
 		c := verticalpodautoscaler.Config{
-			Logger:    config.Logger,
-			K8sClient: config.K8sClient,
-			VpaClient: config.VpaClient,
+			Logger:       config.Logger,
```
QuentinBisson, line 270: I think you need to add that to the other resource.go file under the managementcluster controller

QuentinBisson: https://github.com/giantswarm/prometheus-meta-operator/blob/6c67a80180f2833f972b78897028e56752bf3ffb/service/controller/managementcluster/resource.go#L200

hervenicol: You mean `service/controller/managementcluster/resource.go` ? That you already did! :grin: But you only did one on them, so I also added it there.

QuentinBisson: Ah awesome then

QuentinBisson (approved):

PR #1531, opened 2024-02-14
Add proxy port to CNP if needed

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/29873

## Checklist

I have:


service/controller/resource/ciliumnetpol/resource.go
```diff
@@ -46,6 +48,43 @@ func toCiliumNetworkPolicy(v interface{}) (*unstructured.Unstructured, error) {
...
+		}
+	}
+	if ok {
+		proxyURL, err := url.Parse(proxyString)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		proxyPort := proxyURL.Port()
+		if proxyPort == "" {
+			proxyPort = "80"
```
QuentinBisson, line 83: Well this is nasty but if it works

QuentinBisson (approved):

QuantumEnigmaa (approved):
I like nasty code :)

marieroque:
Need to be tested on 

- [ ] `goat`
- [ ] CAPI MC
- [ ] Vintage MC

service/controller/resource/ciliumnetpol/resource.go
marieroque: Do you have a suggestion to avoid that ?

QuentinBisson: Well, the think is if it is https, it should be 443 and not 80 but no the nasty part was more about the proxy things above btw

marieroque: Code refactored, tell me ;-)

QuentinBisson (approved):

service/controller/resource/ciliumnetpol/resource.go
QuentinBisson: Nice :)

PR #1477, opened 2023-12-20
Scrape some policy-reporter metrics from WCs

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/29351

Intends to scrape only metrics in the form `policy_result_summary{exported_namespace="...", "name="...", status="..."}` in order to measure PSS adoption impact and status.

## Checklist

I have:

- [x] Described why this change is being introduced
- [x] Separated out refactoring/reformatting in a dedicated PR
- [x] Updated changelog in `CHANGELOG.md`

PR #1570, opened 2024-03-14
Scrape limited policy-reporter metrics from WCs

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/29351
Replaces https://github.com/giantswarm/prometheus-meta-operator/pull/1477

## Checklist

I have:


QuentinBisson (approved):
The general idea is to move away from this scrape config but you don't have a choice now

stone-z:
Yeah I'll be happy to ditch it. giving this a quick test on gaia before merging.

side note -- when deploying a branch of PMO, it tries to pull from docker, but the dev branch CI doesn't push to docker. Is that known? I had to edit the deployment to pull from quay

QuentinBisson:
No I was not

stone-z:
With the change to Kyverno's report structure, we would need to scrape the `result` series in order to associate the policy state with the name of the policy. However, that series is _much_ noisier than the `summary` series, and we'd need to scrape and store a bunch of unnecessary information.
Until https://github.com/grafana/agent/issues/834 (or maybe https://github.com/prometheus/prometheus/issues/394 ?) is implemented, which I believe would allow us to drop the unnecessary data at scrape time instead, I don't think we can proceed here.

PR #1571, opened 2024-03-18
Fix labelling schemas

QuentinBisson:
This PR fixes the labelling schema in MC/WC proms to add the missing region and pipeline in existing rules to avoid issues with remote write to grafana cloud like:
image

With remote write and external labels, those actually are sent as 2 different metrics that have the same labels

## Checklist

I have:


QuantumEnigmaa (approved):
LGTM

PR #1580, opened 2024-03-28
Add alertmanager route and receiver for mimir heartbeat

QuentinBisson:
## Description

Towards https://github.com/giantswarm/giantswarm/issues/30293, this PR adds the `route` and `receiver` for mimir heartbeat. The heartbeat alert is added https://github.com/giantswarm/prometheus-rules/pull/1094 and the alert has a label `type: "mimir-heartbeat"` (for legacy we use type: heartbeat). This label will eventually not be needed but it is what it is right now.

This PR should be released before https://github.com/giantswarm/prometheus-rules/pull/1094.

## Checklist

I have:


files/templates/alertmanager/alertmanager.yaml
```diff
@@ -82,6 +93,7 @@ route:
     - team="atlas"
     - type!="heartbeat"
     - alertname!~"Inhibition.*"
+    - alertname!="Heartbeat"
```
QuentinBisson, line 96: I could have used type!=mimir-heartbeat but I feel like this is better

service/controller/resource/alerting/alertmanagerconfig/resource_test.go
```diff
@@ -110,7 +110,53 @@ func TestRenderingOfAlertmanagerConfig(t *testing.T) {
...
+		}
+
+		err = runner.Run()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+func TestRenderingOfAlertmanagerConfigWithMimirEnabled(t *testing.T) {
```
QuentinBisson, line 138: This add new tests to be able to render alertmanager config when mimir is enabled

QuantumEnigmaa:
Ok I think this answers my question asked here https://github.com/giantswarm/prometheus-rules/pull/1094

QuantumEnigmaa (approved):
LGTM

PR #1597, opened 2024-04-23
Add support for slack token

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/30612

Add support for the new slack token but keep backward compatibility with existing slack url.

This allow to use the Slack chat.postMessage url with Alertmanager (see https://prometheus.io/docs/alerting/latest/configuration/#slack_config).

This is required in order automated creation of Slack Token per installation.
As we can only create multiple Slack bot token programmatically but no Slack service Webhook can be created programmatically.


TheoBrigitte:
Here we go alerts are receive on slack via the new token
image

QuentinBisson (approved):

PR #1599, opened 2024-04-30
Remove alertmanager namespace injection in matchers

QuentinBisson:
Towards https://github.com/giantswarm/wepa/issues/62

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


QuentinBisson:
Tested and working on wallaby

hervenicol:
So, if I understand correctly, the idea here is:

Customers should be able to create their own alerts, and their own logic on how to manage these alerts.
- Alerts are `prometheusrules`, and we know how to do that
- logic on how to manage these alerts is defined by `alertmanagerconfigs`

But they currently can't easily create their `alertmanagerconfig` CRs because of a default namespace matching that prevents them to create these CRs in a namespace they can access.

Did I get it right?

QuentinBisson:
Yes you are correct :) By default, all receivers that are created via alertmanagerconfig CR have an matcher that contains namespace=monitoring and this is not a limitation that we need

marieroque (approved):

PR #1614, opened 2024-05-08
Replace apiserver down with controlplane unhealthy

weseven:
Towards: https://github.com/giantswarm/giantswarm/issues/26508

Removes "apiserver_down" dummy inhibition, substituted by "cluster_control_plane_unhealthy". 

## Checklist

I have:

### DIFF
--- a/files/templates/alertmanager/alertmanager.yaml
+++ b/files/templates/alertmanager/alertmanager.yaml
@@ -677,17 +677,6 @@ inhibit_rules:
   equal: [cluster_id]
 
 - source_matchers:
-  - apiserver_down=true
-  target_matchers:
-  - cancel_if_apiserver_down=true
-  equal: [cluster_id]
-
-- source_matchers:
-  - apiserver_down=true
-  target_matchers:
-  - cancel_if_any_apiserver_down=true
-
-- source_matchers:
   - outside_working_hours=true
   target_matchers:
   - cancel_if_outside_working_hours=true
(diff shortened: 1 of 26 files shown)


QuentinBisson (approved):

files/templates/alertmanager/alertmanager.yaml
```diff
@@ -617,6 +617,11 @@ inhibit_rules:
   - cancel_if_cluster_control_plane_unhealthy=true
   equal: [cluster_id]
 
+- source_matchers:
+  - cluster_control_plane_unhealthy=true
+  target_matchers:
+  - cancel_if_any_cluster_control_plane_unhealthy=true
```
QuentinBisson, line 623: Maybe you don't need the any section?

weseven: There were [a couple of sloth-based alerts](https://github.com/giantswarm/sloth-rules/pull/181/files) and [one prometheus-rules alert](https://github.com/giantswarm/prometheus-rules/pull/1167/files#diff-a30358e5d9e631784b9e7bdfbb678e0d12fd23d152089b531231ed66041872fc) using the "any" version of the apiserver_down inhibition so I've done one also for this to maintain parity

QuentinBisson: Right but if it's thé same inhibition in the end maybe we could have only 1 ?

QuentinBisson: Oh there's one for WC and MCs? Then that's fine. I m wondering if we need thé différence but having it will not break anything 😀

PR #1641, opened 2024-06-05
remove-generic-resource-from-alertmanager-config

QuentinBisson:
Towards making https://github.com/giantswarm/prometheus-meta-operator/pull/1638 more reviewable.

This PR gets rid of the generic resource

## Checklist

I have:


service/controller/resource/alerting/alertmanagerconfig/resource.go
```diff
@@ -40,6 +38,21 @@ type Config struct {
...
+
+	baseDomain     string
+	grafanaAddress string
+	installation   string
+	mimirEnabled   bool
+	opsgenieKey    string
+	pipeline       string
+	proxy          func(reqURL *url.URL) (*url.URL, error)
+	slackApiToken  string
+	slackApiURL    string
```
TheoBrigitte, line 53: Since all those fields are taken from `Config` I would rather suggest to embed the Config struct directly within `Resource` this would save code duplication and make future changes easier.

```suggestion
	config Config
```

With this all field access must be replace from
```
r.k8sClient
```
to
```
r.config.k8sClient
```

service/controller/resource/alerting/alertmanagerconfig/create.go
```diff
@@ -0,0 +1,59 @@
...
+	if len(d.GetLabels()) == 0 {
+		d.SetLabels(c.GetLabels())
+	}
+	// without this, it's impossible to change annotations on resources
+	if len(d.GetAnnotations()) == 0 {
+		d.SetAnnotations(c.GetAnnotations())
+	}
+	d.SetFinalizers(c.GetFinalizers())
+	d.SetOwnerReferences(c.GetOwnerReferences())
+	d.SetManagedFields(c.GetManagedFields())
```
TheoBrigitte, line 58: There might be a smarter way of doing this, let me check

service/controller/resource/alerting/alertmanagerconfig/create.go
```diff
@@ -0,0 +1,59 @@
...
+}
```
TheoBrigitte, line 59: This would be a simpler way of updating the object's metadata, but I also thing we should not include such refactoring in here as there's already a lot going on.

```suggestion
		desired.ObjectMeta = current.ObjectMetada
		_, err = r.k8sClient.K8sClient().CoreV1().Secrets(desired.GetNamespace()).Update(ctx, desired, metav1.UpdateOptions{})
		if err != nil {
			return microerror.Mask(err)
		}
	}
	r.logger.Debugf(ctx, "created")

	return nil
}
```

service/controller/resource/alerting/alertmanagerconfig/resource.go
QuentinBisson: Sure, let me redo my 10 PRs 😅

service/controller/resource/alerting/alertmanagerconfig/create.go
QuentinBisson: That's not simpler, that's breaking. Because this does not accept label or annotation changes

QuentinBisson:
@TheoBrigitte let's get this one merged first and then I can do the others :)

service/controller/resource/alerting/alertmanagerconfig/create.go
TheoBrigitte: Yeah a shared package would be nice.

QuentinBisson: It is added here https://github.com/giantswarm/prometheus-meta-operator/pull/1641/files#diff-f9121ba3abf019bf78f83931b583dd84ad4cf19d50fecca517ba6eed29cfde7b :)

TheoBrigitte (approved):
Nice :+1:

PR #1645, opened 2024-06-05
remove-generic-resource-from-prometheus-resource

QuentinBisson:
Towards making https://github.com/giantswarm/prometheus-meta-operator/pull/1638 more reviewable.

This PR gets rid of the generic resource for prometheus

## Checklist

I have:


files/templates/alertmanager/alertmanager.yaml
```diff
@@ -664,12 +664,6 @@ inhibit_rules:
   - cancel_if_nodes_down=true
   equal: [cluster_id]
 
-- source_matchers:
-  - scrape_timeout=true
-  target_matchers:
-  - cancel_if_scrape_timeout=true
-  equal: [cluster_id, instance]
-
```
TheoBrigitte, line 672: This change is not related to the refactoring here and has potential to break, please move this to a different PR

service/controller/resource/alerting/alertmanagerconfig/create.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	if r.hasChanged(current, desired) {
+		resourceutils.UpdateMeta(current, desired)
+		_, err = r.config.K8sClient.K8sClient().CoreV1().Secrets(desired.GetNamespace()).Update(ctx, desired, metav1.UpdateOptions{})
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+	r.config.Logger.Debugf(ctx, "created")
```
TheoBrigitte, line 36: I really dislike this kind of logic being duplicated everywhere.

files/templates/alertmanager/alertmanager.yaml
QuentinBisson: Yes it was done in another PR, but this branch was old :(

service/controller/resource/alerting/alertmanagerconfig/create.go
QuentinBisson: Well, I never really liked the generic resource so :D

Also, this is going away. The main idea is that we can start disabling things slowly when migrating to mimir (especially the clean up part) so that we can add  the pseudo algorigthm in the ensure create resources that we want ;)

```go
if mimirEnabled {
  return EnsureDelete()
}
```

TheoBrigitte: I get that, but it's a lot of code duplication and its very error prone.

QuentinBisson: I understand your concern

PR #1651, opened 2024-06-10
add-delete-if-mimir-enabled to generic resource

QuentinBisson:
This PR replaces all the generic resource removal PRs to make sure we can delete resources when moving to Mimir

## Checklist

I have:

### DIFF
--- a/service/controller/resource/generic/create.go
+++ b/service/controller/resource/generic/create.go
@@ -11,6 +11,9 @@ import (
 )
 
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	if r.deleteIfMimirEnabled {
+		return r.EnsureDeleted(ctx, obj)
+	}
 	desired, err := r.getDesiredObject(ctx, obj)
 	if err != nil {
 		return microerror.Mask(err)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Internal rework to remove the use generic resource to ease out the migration to Mimir.
+- Update generic resource so we can delete resources if mimir is enabled.
 - Remove unused scrape_timeout inhibition.
 
 ## [4.76.0] - 2024-06-03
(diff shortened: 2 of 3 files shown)


service/controller/resource/generic/create.go
```diff
@@ -11,6 +11,9 @@ import (
 )
 
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	if r.deleteIfMimirEnabled {
+		return r.EnsureDeleted(ctx, obj)
+	}
```
QuantumEnigmaa, line 16: I'm not sure to understand the logic behind this. Could you explain pls ? :pray:

QuentinBisson: Of course, so the idea is that if we enabled this property on the controller resource, then we will Always delete instead of create. The goal is that we ensure we delete all existing resources managed by pmo when Mimir is enabled on the installation. This is a conscious choices per resource déclaration

QuantumEnigmaa: Oh ok I think I get it :+1:

QuantumEnigmaa (approved):
LGTM according to Quentin's explanation

service/controller/resource/generic/create.go
QuentinBisson: So the goal is to mimick this behavior https://github.com/giantswarm/prometheus-meta-operator/blob/f26129377cacc9c48179e6efcbf895d10859f804/service/controller/resource/alerting/heartbeatwebhookconfig/create.go#L14 for ressources built using the generic resource. This then avoid the risk of moving away from it

marieroque: Once the resource is deleted, next reconciliation, we are going to enter again in that EnsureCreated and again in EnsureDeleted in case of  Mimir enabled, and we will encounter an error in the EnsureDeleted function as the object won't be there anymore, no ?

QuentinBisson: Nope https://github.com/giantswarm/prometheus-meta-operator/blob/435ec6573943b9440ea4d547d94003e029dadfbd/service/controller/resource/generic/delete.go#L20C2-L24C3

marieroque (approved):

PR #1652, opened 2024-06-11
Remove resources when mimir is enabled

QuentinBisson:
Towards https://github.com/giantswarm/roadmap/issues/3218

This PR always delete resources when mimir is enabled so we do not have leftovers

## Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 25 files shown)


marieroque:
Why are you using `DeleteIfMimirEnabled` and `MimirEnabled` ?
Isn't it the same idea ?

QuentinBisson:
Yes it is but it is clearer in the generic resource why we use it whereas for other resources it's not the case (for instance, for alertmanagerconfig we use the value for something different)

marieroque (approved):

PR #1654, opened 2024-06-11
send from alerts to grafana instead of prometheus

QuentinBisson:
This PR vendors Mimir grafana explore function to be able to send our oncallers to grafana instead of prometheus

## Checklist

I have:


hervenicol:
How did it go? Do you have some interesting findings to share?

QuentinBisson:
Yes, we need a custom alertmanager :)

PR #1764, opened 2024-12-02
Update Alertmanager notification template

TheoBrigitte:
This PR introduces the following changes into the Alertmanager notification template:

- Update alert and query URLs for Mimir to point to the Active notification rather than the rule page
- Move link section (runbook, dashboard, explors) before instance to avoid them being lost due to OpsGenie max description being reached
- Move the warnings for missing runbook and dashboard up into the link section

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Update Alertmanager notification template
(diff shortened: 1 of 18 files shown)


QuentinBisson (approved):

AverageMarcus (approved):

TheoBrigitte:
Here is what the new Opsgenie alert and Grafana alerting pages looks like

image
Note: links are now before instances and other informations. instances are empty, because that's just a dummy alert I created

image
Note: this page shows actual active notifications, but there might be duplicate due to alerts being sent to both opsgenie and slack for instance, are we fine with that ?

PR #1782, opened 2024-12-17
Use proxy_from_environment in Alertmanager

TheoBrigitte:
Use the `proxy_from_environment` option in the Alertmanager configuration to make it read the proxy settings from environment variables (HTTP_PROXY, HTTPS_PROXY, NO_PROXY, etc...) rather having the operator setting proxy via `proxy_url`.

See https://prometheus.io/docs/alerting/0.26/configuration/#http_config

2 things are happening here:
- Replace `proxy_url` in favor of `proxy_from_environment`
- Move the `SlackApiToken` condition so `proxy_from_environment` is not dependent on it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Use `proxy_from_environment: true` in Alertmanager configuration to read proxy settings from the environment.
+
 ## [4.83.0] - 2024-12-17
 
 ### Changed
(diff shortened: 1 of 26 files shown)


QuentinBisson (approved):
Do we need to do this as it's going away with vintage?

QuantumEnigmaa (approved):

TheoBrigitte:
> Do we need to do this as it's going away with vintage?

Not really, but it's good to align configs between the new and the old Alertmanager until the migration is done.

QuentinBisson:
Please dont release it before next year :)

PR #1868, opened 2025-05-08
Stop pushing to aws app collection

QuentinBisson:
Stop pushing to aws app collection

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,38 +21,7 @@ workflows:
         filters:
           tags:
             only: /^v.*/
-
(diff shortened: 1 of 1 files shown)


hervenicol:
Is it the end of PMO? :astonished: 
I don't know if :tada: or :cry: 
Farewell PMO!

Maybe we should directly archive then?

QuentinBisson:
I will archive next week, I just cleaned all our apps first

QuentinBisson:
We will do a little ceremony at the onsite :)

